PR #12, opened 2024-10-08
Upstream TS baseurl, outdir, declarationdir to base tsconfig

jmuzina:
Upstreams several ts config compiler options from the base storybook tsconfig to the base tsconfig

## QA
1. `bun install`
2. `cd packages/ds`
3. `bun run storybook`
4. See that storybook builds and runs successfully


jmuzina:
Closing as not needed, when you do this the types and tsc are emitted relative to the base config, not to the consuming package

PR #17, opened 2024-10-09
wip: Bun / TSC build structure & comparison

jmuzina:
Note: includes changes from #16 , review and merge that PR first.

Builds the `ds` package with both `tsc` & `bun build`, roughly following the structure shown [here](https://coda.io/d/_d5tlz-HCRxz#Notes-P1_tuGTgmNo/r60&view=center).

Adjusts module structure so that UI components can be imported in two ways, depending on user preference:

```ts
// Default import from a component implementation
import Button from "ui/button" ;
// Named import from the UI index
import { Button } from "ui";
```

Task list:

- [x] Build `ds` with `bun build`, with special attention paid to resolving non-TS assets & generating Typescript types
- [ ] Build `ds` with `tsc`, with special attention paid to resolving non-TS assets


jmuzina:
CI fails due to Typescript & Biome disagreeing on whether React imports should use `import type React` or `import React`. Biome by default prefers `import type React`, while Typescript prefers `import React`.

I am making a change to the biome configuration's `jsxRuntime` field to fix this. This follows guidance from Biome [here](https://biomejs.dev/reference/configuration/#javascriptjsxruntime). The `useImportType` rule [reads](https://biomejs.dev/linter/rules/use-import-type/#options) `jsxRuntime` to decide whether to flag these cases. 

We should use `"jsxRuntime": "reactClassic"` if we use `"jsx": "react"`, as indicated by the above documentation.

jmuzina:
>     * document the local override of `"allowSyntheticDefaultImports": true` in ds/tsconfig.json

@advl I was able to  remove this override without causing any errors - I think that was a holdover from our debugging stages. 

>     * let's try to use the modern `react-jsx` and make it work with biome ?

I was able to safely switch this over to modern `react-jsx` and undo the JSX runtime override - no obvious errors. I think it can go in this PR given how simple of a change it was.

Let me know if there's something I've missed that needs to be investigated as part of a separate PR.

PR #19, opened 2024-10-11
wip: Vite boilerplate

jmuzina:
Creates a boilerplate with Vite that can be used to generating React applications using our design system.


jmuzina:
Here's a "boilerplate of the boilerplate", so to speak :smile: 

Generated with `bun create vite` and adding in dependencies to our shared configs.

When setting this up and running checks on it, I wrestled with some `tsc` errors. Dependencies on React 19 types from `@types/react` and `@types/react-dom` that were introduced by this boilerplate bled into other packages. I fixed this by adding the `DOM` lib to the `ts_example` package.

jmuzina:
Next up: finish & test this boilerplate

Later: update this boilerplate to use our code generator, once that is implemented

jmuzina:
running into some issues with vite HMR - page is not reloading when I make a change, and I need to run build before running preview. trying to sort that out now

jmuzina:
@advl I'm getting an interesting error on `install` (which runs `build` via `prepare`) now.

```
 error during build:
      [commonjs--resolver] Failed to resolve entry for package "@canonical/ds". The package may have incorrect main/module/exports specified in its package.json.
          at packageEntryFailure (file:///Users/jmuzina/source/repos/vanilla-monorepo/node_modules/vite/dist/node/chunks/dep-Cyk9bIUq.js:46626:15)
```

the boilerplate uses `@canonical/ds` as a dependency and imports a button from it. 

Interestingly, the error doesn't happen if i run `bun install` a second time.

So, steps for recreation:

1. Clear build artifacts: `rm -rf node_modules packages/ds/dist apps/boilerplate-react-vite/dist`
2. Install: `bun install --frozen-lockfile`
3. Observe error
4. Run install again
5. Observe no error

Have you ever run into something like this before when working with a monorepo?

jmuzina:
@advl 

I have fixed the build error: it was a race condition due to using `--parallel` in the lerna build command. I took oiut `--parallel` for the build script, so now lerna uses `package.json` dependencies to determine build order. 

Regarding react types: these have been added as peer dependencies so that the CI passes. Otherwise, `tsc` fails (as seen [here](https://github.com/canonical/vanilla-monorepo/actions/runs/11408699344/job/31747444319?pr=19)).

I found information on the React 19 guide that seems to suggest installation of the types is still necessary; see [here](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#installing)

CI is now passing and there is an executable node script for initializing a boilerplate, so I'm happy with this. I'll write some QA instructions shortly and move this to ready for review.

apps/boilerplate-react-vite/vite.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [
+    react({
+      include: "src/**/*.tsx",
+    })
```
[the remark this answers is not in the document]
jmuzina: `react()` works, tweaked to use that instead

apps/boilerplate-react-vite/tsconfig.app.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "@canonical/typescript-config-react",
+  "compilerOptions": {
+    "composite": true,
+    "baseUrl": "src",
+    "rootDir": "src",
+    "outDir": "dist/esm",
+    "declaration": true,
+    "declarationDir": "dist/types"
```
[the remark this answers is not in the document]
jmuzina: Good point, tweaked accordingly (removed these compilation options)

jmuzina:
@advl Setup script has been removed

jmuzina:
need to fix some dependency resolution errors that cropped up since merging the CD, will do in the morning tomorrow

configs/typescript-react/package.json
```diff
@@ -27,9 +27,13 @@
...
 	},
 	"peerDependencies": {
+		"react": "^18.3.1 || ^19.0.0",
+		"react-dom": "^18.3.1 || ^19.0.0",
 		"typescript": "^5.6.2"
 	},
 	"dependencies": {
-		"@canonical/typescript-config-base": "^0.1.1-experimental.0"
+		"@types/react": "^18.3.12 || npm:types-react-dom@rc",
+		"@types/react-dom": "^18.3.1 || npm:types-react-dom@rc",
```
jmuzina, line 36: These were added explicitly.

Previously, they were being indirectly installed by one of our dependencies.

If these dependencies are not present, errors like this occur:

```
      src/ui/Page/Page.tsx(1,19): error TS7016: Could not find a declaration file for module 'react'. '/Users/jmuzina/source/work/canonical/repos/ds25/node_modules/react/index.js' implicitly has an 'any' type.
        Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts) file containing `declare module 'react';`
        Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts) file containing `declare module 'react/jsx-runtime';`
      src/ui/Page/Page.tsx(81,15): error TS7026: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
```

configs/typescript-react/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
-	"extends": "@canonical/typescript-config-base",
+	"extends": "@canonical/typescript-config-node",
 	"compilerOptions": {
-		"lib": ["ES2023", "dom"],
+		"module": "ESNext",
+		"moduleResolution": "Bundler",
```
jmuzina, line 5: The React config has been further modified to use `ESNext` and `Bundler` for `module` and `moduleResolution`, respectively, to address the following error that occurs when building the UI library:

```
 ../../node_modules/@joshwooding/vite-plugin-react-docgen-typescript/dist/index.d.ts(1,10): error TS2305: Module '"vite"' has no exported member 'Plugin'.
      ../../node_modules/@storybook/addon-themes/dist/index.d.ts(1,59): error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("storybook/internal/types")' call instead.
```

This lines up with the [Typescript docs](https://www.typescriptlang.org/tsconfig/#module) which state that `ESNext` should be used for code that is to be bundled.

jmuzina (commented):
@advl I've fixed the type-checking errors we were seeing earlier today, and it's also allowed us to remove our usage of `skipLibCheck` which is a nice bonus.

See [this comment](https://github.com/canonical/ds25/pull/19#issuecomment-2483634800) and my inline reviews for more.

Let me know what you think.

packages/ds-react-core/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './ui/index.js';
```
jmuzina, line 1: Fixed missing index export

packages/ds-react-core/src/ui/index.ts
```diff
@@ -1,3 +1,3 @@
-export { Button, type ButtonProps } from "./Button/index.js";
-export { Header, type HeaderProps } from "./Header/index.js";
-export { Page } from "./Page/index.js";
+export * from "./Button/index.js";
+export * from "./Header/index.js";
+export * from "./Page/index.js";
```
jmuzina, line 3: These index files already export named and type exports, we can use `export *` for simplicity

jmuzina:
@advl Updated per our discussions today:

## React Core UI - Storybook type checking fixes

The errors arose from me trying to remove `skipLibCheck` from the react core package for increased strictness. There are two ways to resolve this in the short term:

1. Type-check storybook files (`.stories` and `.stories.ts(x)`) with `"module": "ESNext", "moduleResolution": "Bundler"`
2. Enable `skipLibCheck` for storybook files (`.stories` and `.stories.ts(x)`)

I have used option # 2, `skipLibCheck` per our earlier discussion when you mentioned we should try and avoid changing module settings. This is already the behavior of the library on `main`, the change I made here is setting `skipLibCheck` to `false` in the library build (which excludes storybook files).

## Boilerplate Tsconfig

The boilerplate's typescript configuration has been brought more into line with the other packages - it now has a `tsconfig.json` that checks all typescript files (intended for type-checking), and a configuration for building the application (that excludes the vite config file). 

The building config file has been named `tsconfig.app.json`, as users are probably building their applications with `vite build` rather than `tsc` in this context. `vite.config.ts` has been configured to build with `tsconfig.app.json` using the [vite-tsconfig-paths](https://www.npmjs.com/package/vite-tsconfig-paths) plugin.

### Open question

It feels a bit strange to add a dependency just so that we can build with a tsconfig other than `tsconfig.json`. 

We either have to break with our current convention of `tsconfig.json` being used just for type-checking with `tsc --noEmit` (which would allow us to remove this dependency), or use this dependency to continue with that convention.

Have a look when you can, please.

apps/boilerplate-react-vite/vite.config.ts
```diff
@@ -0,0 +1,13 @@
+import react from "@vitejs/plugin-react";
+import { defineConfig } from "vite";
+import tsconfigPaths from "vite-tsconfig-paths";
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [
+    tsconfigPaths({
+      projects: ["tsconfig.app.json"],
+    }),
```
jmuzina, line 10: Let's try and remove this and just use one `tsconfig.json`

jmuzina:
Last task: cleanup the tsconfig.app.json in the boilerplate, try and use just one config @jmuzina

PR #21, opened 2024-11-06
High-level monorepo docs

jmuzina:
- Creates high-level documentation for the monorepo as a whole.
- Adds a PR template.
- Sets the monorepo root package to be private, as it will not be published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,54 @@
-# vanilla-monorepo
-Proof-of-concept / exploration of the Vanilla design system as a monorepo
+# DS-25 - Canonical's new design system
 
-## Installation
+Canonical's new design system.
+This is the successor to [Vanilla framework](https://vanillaframework.io).
 
-```sh   
-bun i
-git config diff.lockb.textconv bun
-git config diff.lockb.binary true
-```
+## Getting started
 
+1. Install [NodeJS](https://nodejs.org/en/download/package-manager) version 20 or later.
+2. Install the [bun package manager](https://bun.sh/).
+3. Configure your local git client to read Bun lockfiles as text: `git config diff.lockb.textconv bun && git config diff.lockb.binary true`.
+   1. This will no longer be necessary once Bun [switches to a text-based lockfile](https://github.com/oven-sh/bun/issues/11863)
+4. Install Node dependencies: `bun install`.
+
+The monorepo's dependencies should now be installed.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,11 +1,54 @@
...
+This is the successor to [Vanilla framework](https://vanillaframework.io).
 
-```sh   
-bun i
-git config diff.lockb.textconv bun
-git config diff.lockb.binary true
-```
+## Getting started
 
+1. Install [NodeJS](https://nodejs.org/en/download/package-manager) version 20 or later.
```
bartaz, line 8: Do you need node, if we use bun?

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "vanilla-monorepo",
```
bartaz, line 2: Not necessarily in scope of this PR, but could be worth renaming this (and repo) to not include "vanilla" in the name.

README.md
```diff
@@ -1,11 +1,54 @@
...
+2. Install the [bun package manager](https://bun.sh/).
+3. Configure your local git client to read Bun lockfiles as text: `git config diff.lockb.textconv bun && git config diff.lockb.binary true`.
```
bartaz, line 10: Is this needed to use it? Or is this some kind of "quality of life" developer option?

README.md
jmuzina: Technically no, but it is necessary for testing the boilerplate in #19 . I ran into an error running local package bin scripts in bun workspaces with `bunx` that I have been circumventing by using `npx`. 

Filed an [issue](https://github.com/oven-sh/bun/issues/15036) with Bun. 
There is a similar [issue](https://github.com/oven-sh/bun/issues/11729) for running package bin scripts with local filepaths, so there currently isn't a way for bun to run a package bin script unless it is published to a package registry first.

package.json
jmuzina: We can rename it to "DS-25" like we've been calling it recently, but I think it'll be less confusing to do that as a separate PR. I'll file one.

jmuzina: Filed as #23

README.md
jmuzina: It's not needed to use it but it does qualify as quality of life for sure. I'll move it to a separate section that is not in the main install notes.

PR #27, opened 2024-11-13
CI Enhancements

jmuzina:
## Done

- Builds and tests the project on pushes to `main`
- Adds [inclusive naming check](https://github.com/canonical/inclusive-naming) to pull requests

### PR readiness check

If this PR contains Vanilla SCSS code changes, it should contain the following changes to make sure it's ready for the release:

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -5,6 +5,7 @@ on:
       - main
 jobs:
   build:
+    name: Build & test
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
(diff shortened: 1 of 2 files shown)


jmuzina:
> Is there a way we can switch this to a cli test already with low effort ?

No, I don't think there's any low-effort ways to do this, see @bartaz comment [here](https://github.com/canonical/Inclusive-naming/issues/17#issuecomment-2478049065)

The first thing that comes to mind is using [act](https://nektosact.com/introduction.html) to run the existing action locally. However, Act itself is a Go binary, so we can't install it as a NPM package. 

To achieve this from the CLI with low-effort, we either need to install Woke or Act manually, which complicates the dev setup process more than I'm comfortable with.

jmuzina:
Per @advl : perhaps we can use or implement some Go wrapper that allows us to run this as a CLI tool

jmuzina:
Simplifying this to just run tests on push to `main`, moving inclusive naming to be a separate issue

PR #32, opened 2024-11-15
Configure NX Task execution analysis & caching

jmuzina:
## Done
- Configures task dependencies for "build", "check" tasks. This causes the "build" and "check" tasks to run in dependency order where needed, and otherwise in parallel. See [docs](https://lerna.js.org/docs/concepts/task-pipeline-configuration#define-task-dependencies-aka-task-pipelines).
- Configures NX caching. When a task is executed, the state of each package is cached. If a package task is re-run and the package has not been changed since it was cached, the task will not be re-run for that package. See [docs](https://lerna.js.org/docs/api-reference/configuration#cache) for more.

Fixes [comment](https://github.com/canonical/ds25/pull/30#discussion_r1844302462)

## QA

1. `bun i`
2. `bun run build`: see that no tasks are cached in the CLI output, so all builds are run
3. `bun run build`: see that all tasks are cached in the CLI output, so no builds are run
4. Change a package that has a build target (such as `ds-react-core`). Changing any file in it is sufficient to cause the cache to be invalidated.
5. Run `bun run build` again and see that the other packages use their caches but the changed package is re-built.

We can't directly test the sequential task execution as `main` does not have packages with `build` targets that depend on each-other yet.

### PR readiness check

## Screenshots
Screenshot 2024-11-15 at 14 08 58

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,6 @@ env[23]/
 # Project-specific ignores
 js/app.js.map
 static/js/modules
+
+# NX cache
+.nx/
--- a/nx.json
+++ b/nx.json
@@ -1,0 +1,17 @@
+{
+  "cacheableOperations": ["build", "test", "check", "check:fix"],
+  "targetDefaults": {
+    "build": {
+      "dependsOn": ["^build"],
+      "cache": true
+    },
+    "check": {
+      "dependsOn": ["^check"],
+      "cache": true
+    },
+    "check:fix": {
+      "dependsOn": ["^check:fix"],
+      "cache": true
+    }
+  }
+}
--- a/package.json
+++ b/package.json
@@ -22,11 +22,11 @@
     "apps/*"
   ],
   "scripts": {
-    "build": "lerna run build --parallel",
+    "build": "lerna run build",
     "lerna": "lerna",
     "test": "bun run check",
-    "check": "lerna run check --parallel",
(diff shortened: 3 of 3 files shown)


jmuzina:
>     1. Why is `--parallel` not necessary anymore ?
@advl 
`--parallel` was deprecated when combined with Nx target configuration in Lerna v6. See [docs](https://lerna.js.org/docs/lerna6-obsolete-options#--parallel).

If you use `lerna run <taskname> --parallel` when target configuration is enabled, this warning is thrown:
```
lerna WARN run "parallel", "sort", and "no-sort" are ignored when Nx targets are configured. See https://lerna.js.org/docs/lerna6-obsolete-options for details.
```


>  2. How in the build system are handled build dependencies ? (Package A build failing if package B build is not done before)

Nx handles build dependencies by constructing a dependency graph between packages (e.g., Package A depends on Package B). 

Nx uses this graph to sequence dependent tasks so that dependencies are built before their dependents (building B before A, for example). 

`dependsOn` is used to make this relationship between specific tasks more explicit; this could for example let you define `test` as depending on `build` (so running `lerna run test` automatically runs `lerna run build` first).

Some tasks, like custom checks or additional steps, aren't automatically handled by the dependency graph, so defining this relationship makes sure everything runs in the right sequence.

See the [Nx docs on `dependsOn`](https://nx.dev/reference/project-configuration#dependson) for more

PR #38, opened 2024-11-21
Fix : Solved first install issues by removing the prepare script in the root.

advl:
## Done

The installation of the monorepo was failing on a blank state (no mode_modules).

It seems there was a race condition where lerna and nx should not be called directly after the first monorepo install.

Removing the `prepare` script fixes the issue.

Fixed an additional typescript error on the boilerplate.

## QA

- Clone the monorepo
- `bun i` should pass

## Additional note
When having problems to install at the root of the monorepo, erasing node_modules seem to be a good solution.

`rm -rf ./node_modules`
`rm -rf ./**/node_modules`

### DIFF
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -1,7 +1,8 @@
+---
 name: Setup environment
 description: Installs build tools and dependencies
 inputs:
-  bun-version: 
+  bun-version:
     default: latest
     description: Version of Bun to use
 runs:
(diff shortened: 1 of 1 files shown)


bartaz:
I'm not sure what's the issue being fixed here. I cloned a clean repo from main branch and all seems working fine when running `bun i` for the first time.

How can I reproduce the issue this PR is fixing?

jmuzina:
I also cannot reproduce the issue mentioned here. Here's how I'm attempting to reproduce:
`rm -rf ./**/{dist,node_modules} && bun i`

I have done this at least 10 times and seen no errors - maybe I am just lucky? Let's touch base on this Monday @advl

PR #37, opened 2024-11-21
Add negative button, move variants to global intents

bartaz:
## Done

Moves positive variant to "intents".

## QA

- Go to ds-react-core package directory
- `bun run storybook`
- navigate to button page: `http://0.0.0.0:6006/?path=/docs/example-button--docs`

## QA

- [Add QA steps]

### PR readiness check

- [ ] PR should have one of the following labels:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [ ] All packages define the required scripts in `package.json`: `build`, `check`, and `check:fix`.

## Screenshots

image

PR #40, opened 2024-11-22
Implement button modifiers on component level

bartaz:
## Done

This is a alternative proposal of #37. Instead of abstracting component variants to global "intents" that currently cause CSS cascade/inheritance issues, I'm using exisiting component CSS variables to provide modifier "themes".

While this doesn't abstract the modifiers to global "intents", it does simplify definitions and managing of component variants as it avoids adding another abstraction and layer of variables. Modifiers defined on a level of component are easier to find/add/modify as they are defined together with component itself.

Abstraction and sharing of the styling is still possible on level of semantic tokens specific to given "intent" (positive, negative, etc).

## QA

- [Add QA steps]

### PR readiness check

### DIFF
--- a/packages/ds-react-core/src/ui/Button/Button.stories.ts
+++ b/packages/ds-react-core/src/ui/Button/Button.stories.ts
@@ -41,11 +41,31 @@ export const Positive: Story = {
 };
 
 /**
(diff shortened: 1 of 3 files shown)


bartaz:
Implemented as intents in #37

PR #46, opened 2024-11-26
Plop code generator

jmuzina:
## Done

Feature-parity version of #28 , implemented with [Plop](https://plopjs.com/). Use this to compare API surface of this generator with the others.

## QA

- TBD

### PR readiness check

### Remaining exploration
1. [ ] Plop uses [handlebars](https://handlebarsjs.com/) instead of [EJS](https://ejs.co/) for templating. HBS is a bit tougher than EJS to format & lint, and the generator's output also seems to have inconsistent formatting. This would need to be resolved before this could be used in production.
2. [ ] Plop's more declarative API makes it tougher to share global context (for example, generator package version) across the entire set of generators. Perhaps a "helper" can be defined to expose that context to templates.


jmuzina:
Closing in favor of #28

PR #48, opened 2024-12-03
Added basic heading styles

bartaz:
## Done

Adds basic typography styles for headings.

Note:
- styles are added to the styles package
- storybook docs are added to ds-react-core package (so that they can be presented with the rest of the docs)

## Open questions (can be addressed later):

1. How to address typography changes on different screen sizes? Current Vanilla uses media queries to change heading sizes between mobile/larger screens, and the basic unit size is increased on very large screens.
2. Base styles have their own package, but storybook is in `ds-react-core` package. Should docs/stories for other packages be kept in `ds-react-core`? Should we have separate storybooks in each package? Should we one storybook external to any packages, that would collect all examples from different packages (is that possible)?
3. In Vanilla we have different font-files for Ubuntu font based on character subsets. Do we still need this? How should we implement it in configurable way?

## QA

- Run the storybook in `ds-react-core`
- Check typography examples

## Screenshots

image

### DIFF
--- a/packages/ds-react-core/src/ui/Typography.stories.tsx
+++ b/packages/ds-react-core/src/ui/Typography.stories.tsx
@@ -1,0 +1,80 @@
+// FIXME
+// This doesn't really fit in the ds-react-core package,
+// but it's a simple way to show examples in existing Storybook.
+// TODO:
+// - do we need a separate Storybook/examples for styles package?
+// - should he have a separate Storybook package, that would group all stories from different sources?
+
+const meta = {
+  title: "Styles / Typography",
+};
+
+export default meta;
+
+export const Headings = {
+  decorators: [
+    () => (
+      <>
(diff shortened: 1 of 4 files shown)


bartaz:
> I have seen there is a [special storybook block for typography](https://storybook.js.org/docs/writing-docs/doc-blocks#typeset). Is is something we could use ?

Not exactly, as far as I can see. It can be useful to showcase a font with a list of available sizes (https://storybook.js.org/docs/api/doc-blocks/doc-block-typeset), but it doesn't seem to be designed to show specific styles assigned to specific text elements (headings or paragraphs). It seems to be accepting sizes/weights as React params, rather than taking them from CSS.

So I guess it wouldn't replace examples that we have, but we potentially could use something like that on Typography, or Ubuntu font documentation page.

Issue #53, opened 2024-12-04
Installing code generator globally with Bun does not make it visible to `yo`


jmuzina:
Installing the code generator package with `bun` globally does not list it as an available generator.

## Steps to reproduce

1. `bun add -g yo @canonical/generator-canonical-ds`
2. List installed generators: `yo --generators`
3. Observe our generator is not listed in the generators list
4. Try to run the generator directly: `yo @canonical/canonical-ds `
5. Observe that that the generator cannot be found.

### Installing with NPM
This problem does not occur with `npm`.

1. `npm i -g @canonical/generator-canonical-ds`
2. Repeat steps 2 and 4 from reproduction steps, see that the generator is runnable.

## Installing with Bun in package scope
This problem does not occur if the generator is installed in a package scope (rather than globally).

1. `bun add -d @canonical/generator-canonical-ds`
2. Repeat steps 2 and 4 from reproduction steps, see that the generator is runnable

PR #54, opened 2024-12-04
Update generator global install package manager documentation

jmuzina:
## Done

Mitigates #53 by suggesting users to use `npm` instead of `bun ` to globally install the code generator.

## QA
Perform these steps **outside of the monorepo** to verify that the generator package from the monorepo workspace is not visible to your node environment.

1. `npm i -g yo @canonical/generator-canonical-ds`
2. `yo --generators` - Verify the generator is listed in the options
3. `yo @canonical/generator-canonical-ds` - Verify the generator runs successfully.

### PR readiness check

### DIFF
--- a/apps/generator-canonical-ds/README.md
+++ b/apps/generator-canonical-ds/README.md
@@ -2,7 +2,7 @@
 
 ## Getting Started
 
-1. Install [Yeoman](https://yeoman.io/): `bun add -g yo`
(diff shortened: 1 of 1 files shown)


jmuzina:
> Couldn't we keep both ?

@advl We could, but I fear it might be a bit clunky to say something to the effect of "you can only use Bun to install this if you install it per-package. Use `npm` if you need to install globally".

PR #26, opened 2024-11-13
feat: Add a font dimension extractor and programmatic nudge reader.

advl:
## Done
Implements a typography package that :
- Provides a font dimension extractor `extractFontData fontfile.otf` or, `extractFontData fontfile.otf 1.5 --all`
- Provides an automated calculator of nudges based on context font-sizes and line-heights (`index.css`)
- Provides an example of usage `examples/index.html`

Additionally, this PR provides a basic implementation in our own styles.

## QA

- Open `packages/typography/example/index.html`. The font should be aligned on the baseline.
- Run the core storybook, look at the stories with the baseline addon toggled on, the font should be aligned.

### PR readiness check

- [x] PR should have one of the following labels:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [x] All packages define the required scripts in `package.json`: `build`, `check`, and `check:fix`.

## What this PR doesn't do

- It does not provide the full documentation of the feature.
- It does not implement all the required css variables on our own styles.
- It does not fully clean the example code and folder.

PR #55, opened 2024-12-04
add `@types/node` dependency to utils package

jmuzina:
## Done

Adds `@types/node` dev dependency to the utils package, which depends on Node types but does not explicitly list them in its package file. This is only working currently because other dependencies depend on `@types/node`.

## QA
CI check is sufficient to verify this works.

### PR readiness check


jmuzina:
This is being subsumed by #26 .

@advl is removing the `node` type dependency from the utils package.

PR #61, opened 2024-12-06
Incremental CI/CD workflows

jmuzina:
## Done

- The CI/CD pipeline now only builds & tests packages that need to be built and tested.
  - In the CI pipeline, the PR's merge commit is tested against its target branch. Packages that are changed (directly or indirectly) by a PR, as well as packages that have been changed on the target branch since the PR branch was created, will be built and tested. Other packages will be skipped.
  - In the CD pipeline, the packages and dependents that have changed since the previous tag will be be built and tested. 
- Drive-by: Consolidated the "Tag" workflow into two jobs - the "build" job was used to build and test, but the "version" job needs to build and test anyway in order to perform package versioning. This behavior has been consolidated in order to use less time and resources.

Fixes #60 

## QA

1. Open [test PR](https://github.com/jmuzina/ds25/pull/17) that makes a [change to the react tsconfig](https://github.com/jmuzina/ds25/pull/17/commits/43c9443859e066fca3f77e279633b27fcb1ff14e). 
2. Inspect the [PR CI workflow](https://github.com/jmuzina/ds25/actions/runs/12206931095/job/34057422032) and see that only the `ds-react-core` and `boilerplate-react-vite` packages are built (they both depend on the React tsconfig, and the React tsconfig has no build script). 

Screenshot 2024-12-06 at 17 28 13

3. See that the react tsconfig, boilerplate, and react core packges are all `check`'d

Screenshot 2024-12-06 at 17 28 42

4. Inspect the [CD workflow](https://github.com/jmuzina/ds25/actions/runs/12206947695/job/34057471269) and see that the same packages modified by QA step 2 are built.

Screenshot 2024-12-06 at 17 30 31

5. Verify that only the packages that are being versioned have format fixing applied to them
Screenshot 2024-12-06 at 17 31 22

### PR readiness check

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -3,16 +3,26 @@ on:
   pull_request:
     branches:
       - main
+env:
+  PR_TARGET_REF: "refs/remotes/origin/${{ github.event.pull_request.base.ref }}"
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
+        with:
+          # Lerna needs *all* the git history to determine which packages have changed
+          fetch-depth: 0
 
-        # "prepare" hook on `bun install` runs `bun run build`
       - name: Setup environment
         uses: ./.github/actions/setup-env
+        with:
+          # Ignore lifecycle scripts to avoid building packages that haven't changed
+          ignore_scripts: true
 
-      - name: Run tests
-        run: bun run test
+      - name: Build changed packages
+        run: bun run lerna run build --since ${{ env.PR_TARGET_REF }}
+
+      - name: Code quality checks
+        run: bun run lerna run check --since ${{ env.PR_TARGET_REF }}
--- a/nx.json
+++ b/nx.json
@@ -2,18 +2,15 @@
   "cli": {
     "packageManager": "bun"
   },
-  "cacheableOperations": ["build", "test", "check", "check:fix"],
   "targetDefaults": {
     "build": {
       "dependsOn": ["^build"],
       "cache": true
     },
     "check": {
-      "dependsOn": ["^check"],
       "cache": true
     },
     "check:fix": {
-      "dependsOn": ["^check:fix"],
       "cache": true
     }
   }
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "build": "lerna run build",
     "prepare": "bun run build",
     "lerna": "lerna",
-    "test": "bun run check",
     "check": "lerna run check",
     "check:fix": "lerna run check:fix"
   },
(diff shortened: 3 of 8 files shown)


jmuzina:
This PR is missing just one consideration: when versioning a jump from experimental phase to non-experimental or releasing a major version, Lerna expects to version and publish **all packages**, even if they weren't actually changed. ([ref](https://lerna.js.org/docs/features/version-and-publish#fixedlocked-mode-default)).

.github/workflows/pr.yml
```diff
@@ -3,16 +3,31 @@ on:
...
 
-        # "prepare" hook on `bun install` runs `bun run build`
       - name: Setup environment
         uses: ./.github/actions/setup-env
+        with:
+          # Ignore lifecycle scripts to avoid building packages that haven't changed
+          ignore_scripts: true
+
+      - name: Build changed packages
+        run: bun run lerna run build --since ${{ env.PR_TARGET_REF }}
```
jmuzina, line 30: Check if a package depends on another that has a build stage, and the dependency is not built because it is not changed - we need to make sure dependencies are built

jmuzina: This is handled by the `nx.json` "^build" - packages build their dependencies before themselves

nx.json
```diff
@@ -2,18 +2,15 @@
   "cli": {
     "packageManager": "bun"
   },
-  "cacheableOperations": ["build", "test", "check", "check:fix"],
```
jmuzina, line 5: `cacheableOperations` is a legacy configuration option, `cache: true` as defined below is now the [recommended configuration](https://lerna.js.org/docs/features/cache-tasks#setup)

nx.json
```diff
@@ -2,18 +2,15 @@
...
     "build": {
       "dependsOn": ["^build"],
       "cache": true
     },
     "check": {
-      "dependsOn": ["^check"],
       "cache": true
     },
     "check:fix": {
-      "dependsOn": ["^check:fix"],
```
jmuzina, line 16: If we `check` a package, we don't necessarily want to check its dependencies. This way, if a package is updated (and thus in-scope for `lerna run check --since <ref>`), only it is checked, not its dependencies.

package.json
```diff
@@ -25,7 +25,6 @@
     "build": "lerna run build",
     "prepare": "bun run build",
     "lerna": "lerna",
-    "test": "bun run check",
```
jmuzina, line 28: `test` running `check` is misleading API, we should add `test` once our packages define `test` scripts, then run `bun run lerna run check --since <ref> && bun run lerna run test --since <ref>` as a code quality check.

jmuzina: Also, in the case that tests/checks are run incrementally, we need to pass a `--since` argument to `lerna run` by invoking `bun run lerna run check` instead of `bun run check`. So, either all packages need to define a script that runs both tests and checks, or we run both test and check separately at the top level.

PR #66, opened 2024-12-09
Fix: declare types explicitly in packages to avoid overlapping implicit definitions

advl:
## Done
- Adds explicit types dependencies in `tsconfig.json`
- Explains the approach in the main README.md 
  
## QA

- Add a new package with conflicting `@types` in relation with another package, for instance add `@types/bun` while the existing package relies on `@types/node`. 
- In the new package, ensure that tsconfig.json types explicitly declare `"types":["bun"]`
- Run `bun run build` at the root level, it should pass.

### PR readiness check

  - [x] All packages: `check` and `check:fix`.


README.md
```diff
@@ -64,3 +64,15 @@ included in the monorepo's CI workflow, helping to avoid build errors later.
...
+
+The `@types` dependencies required by the monorepo packages and apps are by default hoisted in the root `node_modules`. However, this behaviour has two downsides : 
+- It can lead to conflicts between the types of different packages, for instance `bun` and `node` export overlapping types.
+- Consumption of types in packages and apps are not explicit, leading to hard to debug issues.
+
+For this reason, we have opted for explicitly declaring the `@types` dependencies both in the `package.json` under `@devDependencies` and in  `tsconfig.json` under `compilerOptions/types` for each package and app. 
+
+## Notes for developers
+
+- A hard-reset of this repo looks like this : `rm -rf node_modules .nx`. In some cases, you might want to reset the build output of packages with `rm -rf ./**/dist`.
```
jmuzina, line 78: Idea, not a blocker: maybe we should add a "clean" package script at the top-level for this. That way devs can just run `clean` instead of looking for these specific commands.

packages/utils/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "@canonical/typescript-config-base",
   "compilerOptions": {
-    "baseUrl": "src"
+    "baseUrl": "src",
+    "types": ["node"]
```
jmuzina, line 5: I believe we shouldn't need the `node` type here now, wince you changed the timer type in `throttle.ts` to `ReturnType<>`.

apps/ts_example/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "@canonical/typescript-config-base",
   "compilerOptions": {
-    "baseUrl": "src"
+    "baseUrl": "src",
+    "types": ["node"]
```
jmuzina, line 5: This shouldn't need `node` anymore, since we have removed the node-specific type depencies from the utils package.

packages/utils/tsconfig.json
jmuzina:
```suggestion
    "types": []
```

apps/boilerplate-react-vite/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "@canonical/typescript-config-react",
   "compilerOptions": {
     "baseUrl": "src",
+    "types": ["node", "react", "react-dom"],
```
jmuzina, line 5: Please add explicit type `devDependencies` for these to the boilerplate `package.json`, it is currently missing them

jmuzina (changes requested):

packages/ds-react-core/vite.config.js
jmuzina, line 1: Looks like a generated file made its way into the PR

packages/ds-react-core/vite.config.d.ts
jmuzina, line 1: I think an extra type declarations file made it into the PR here

README.md
jmuzina, line 78: This should consume the new clean command now:
```suggestion
- To clean dependencies and built artifacts, run `bun run special:clean`. In some cases, you might want to reset the build output of packages with `rm -rf ./**/dist`.
```

jmuzina (changes requested):

jmuzina (approved):
Works great once #68 is merged, let's merge them both. Thanks!

PR #67, opened 2024-12-09
WIP: Add baseline storybook addon

bartaz:
## Done

This is an attempt to implement baseline grid toggle as an addon.

Contents of `packages/storybook-addon-baseline-grid` are mostly a boilerplate based off https://github.com/storybookjs/addon-kit

Most notable files in this package:
- `src/components/Tool.tsx` - a component that displays a toolbox button (with lightning icon) that is supposed to trigger the addon (by setting a storybook global variable on click) - this mostly works as expected
- `src/manager.ts` - registers the addon - seems to work
- `src/preview.ts` - is supposed to add decorator and initial globals to storybook - works fine in local storybook in this package, but doesn't seem to work in external storybook
- `src/withGlobals.ts` - a decorator that does something based on global value, in our case toggles a class name on storybook body - doesn't seem to happen in external storybook

## QA

The part that works:

- Go to `packages/storybook-addon-baseline-grid`
- run `bun run build`
- run `bun run storybook`
- A storybook local to addon package will run, with boilerplate Storybook components
- Go to a story page (not Docs), for example http://localhost:6006/?path=/story/example-button--primary
- Button with ⚡ should be on the toolbar. Clicking it should trigger baseline grid.

storybook1

The part that doesn't work:

- In `packages/storybook-addon-baseline-grid`
- run `bun run build`
- Go to `packages/ds-react-core`
- run `bun run storybook`
- This would run our own storybook with DS25 components and styles
- Open a story (not Docs), for example http://localhost:6006/?path=/story/button--default
- Button with ⚡ should be in the toolbar
- Click on it - nothing will happed (baseline should become visible)
- Showing baseline via locally developed dropdown ("Baseline hidden") works fine

storybook2


jmuzina:
I was able to get this to work by rebasing it onto #66 . This should work fine after that PR is merged.

bartaz:
Rebased, so the build problems are gone.

Still, the addon doesn't work inside of the `ds-react-core` package. Button is added to toolbar, but it doesn't trigger the action.

bartaz:
As the addon doesn't seem to be working from within monorepo, we decided not to add it here. It would need to live as a separate repo.

PR #78, opened 2024-12-12
ft: WIP: Ft use svg

advl:
## Done

A work-in progress PR displaying a PoC to leverage `<svg><use>` as a way to consume SVG sprites.

The trick of the sprites is to include several groups in the svg, and display only the targetted one (`sprite.svg#my-icon-name`).

A second minor feature is included : a PoC of SVG theming using intent colors. The last story, `Illustration` displays this concept.
In case we would like to proceed with this approach, it would only be necessary to refine the API of the component.
A third minor feature is included : the ability to programatically animate strokes of SVGs.

## QA
Code Q/A :
- Look at `sprite.svg` using the source diff.
- Q/A the component api.

Visual Q/A : 
- Go to the design system package
- Run Storybook
- Look at the stories. Try to change the intents applied (negative <> positive). This should work on all stories.

### PR readiness check

- [ ] PR should have one of the following labels:
  - [ ] All packages: `check` and `check:fix`.

## Screenshots


jmuzina:
> In case we would like to proceed with this approach, it would only be necessary to refine the API of the component.

What do you see being refined about the component's API?

PR #121, opened 2025-02-03
chore(generator): rename style.css to styles.css, rename .test.tsx to .stories.tsx

jmuzina:
## Done

Renames files generated by the React Component generator for better naming consistency:
- style.css -> styles.css
- ComponentName.test.tsx -> ComponentName.tests.tsx

## QA

- Checkout this PR
- `bun i && cd packages/react/ds-core/src/ui && yo @canonical/ds:component TestComponent --withStyles --withStories`
- `bun run storybook` and make sure the component appears as expected
- `bun run test` and make sure the tests for the component run and pass as expected

### PR readiness check

- [x] PR should have one of the following labels:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [x] All packages define the required scripts in `package.json`:
  - [ ] All packages: `check`, `check:fix`, and `test`.
  - [ ] Packages with a build step: `build`.

PR #122, opened 2025-02-03
fix(generator): Fix generator creating index file in wrong directory

jmuzina:
## Done

Follow up to [#121](https://github.com/canonical/ds25/pull/121#issuecomment-2632393806) that fixes a bug where the component generator created component index file re-exports in the wrong directory when the cwd final path-part was not PascalCased.

Also, adds `CHANGELOG` entries for #121 

## QA
1. Checkout PR
2. `bun run special:clean && rm -rf packages/generator-ds/generators && bun i`
3. `cd packages/react/ds-react-core/src/ui`
4. `yo @canonical/ds:component MyComponent --withStyles --withStories`
5. See that the re-export is added to `ds-react-core/src/ui/index.ts`, not `ds-react-core/src/Ui/index.ts`

### PR readiness check


packages/generator-ds/src/component/index.ts
```diff
@@ -70,12 +70,14 @@ export default class ComponentGenerator extends Generator<ComponentGeneratorOpti
     const rawPath = super.destinationPath(...inPath);
     const dirName = path.dirname(rawPath);
 
+    // If the path is an ancestor of the cwd, we don't want to PascalCase the parent directory, as this would be an incorrect absolute filepath
+    // If the path is a descendant of the cwd, we want to PascalCase the parent directory to enforce React component naming conventions.
+    const componentParentDir = paths.isAncestor(this.env.cwd, rawPath)
+      ? path.basename(dirName)
+      : casing.toPascalCase(path.basename(dirName));
```
jmuzina, line 77: This can be simplified by only appending ../pascalcase() to the path if the path is an ancestor instead of having a ternary, I'll tweak that when I'm next in the office :)

jmuzina:
>  We should not account for the user giving an ancestor, but simply avoid it by documenting it and/or adding an invariant. 

This issue won't be solved by documentation or adding an invariant - the issue is that the existing code pascal-cases the last part of filepaths in order to enforce React component naming. If someone supplies `button`, their component will be called `Button`, for example, so their folder is `Button`, implementation file is `Button.tsx`, etc.

However, this pascal-casing should only apply when the path being written is a descendant of the CWD, and a new folder is being created (which should follow React convention). In other cases (like writing to `.index.ts` of the `ui` directory), the last directory path-part shouldn't be pascal-cased, as that would cause the path to be incorrect unless that path-part is actually pascal-cased in the user's filesystem.

I've implemented a much simpler fix for the bug that doesn't require changes to casing/paths utils.

PR #124, opened 2025-02-05
feat(Generator): added shorthands properties `-c` for styles and `-s`for stories.

advl:
## Done

Added shorthands for the two generator flags
`--withStyles` can now also be accessed with `-c` (as in css)
`--withStories` can now also be accessed with `-s`

## QA

- `bun i`
- `bun run yo @canonical/ds:component SomeComponent -cs`

### PR readiness check

### DIFF
--- a/packages/generator-ds/README.md
+++ b/packages/generator-ds/README.md
@@ -3,8 +3,7 @@
 ## Getting Started
 
 1. Install [Yeoman](https://yeoman.io/) and this generator: `npm i -g yo @canonical/generator-ds`
-2. Run the generator: `yo @canonical/ds:component`
-3. The required arguments and possible flags are listed by invoking the generator with the `--help` flag, for instance `yo @canonical/ds:component --help`
+2. Run the generator: `yo @canonical/ds`
 
 
 
--- a/packages/generator-ds/src/component/README.md
+++ b/packages/generator-ds/src/component/README.md
@@ -4,8 +4,5 @@ This generator creates a new React component with a basic structure.
 
 ### Getting Started
 1. Install [Yeoman](https://yeoman.io/): `npm i -g yo @canonical/generator-ds`
-2. Run the generator: `yo @canonical/ds:component path/to/component`
-
-The following optional flags can be used:
-- `--withStyles`: Include a CSS file for the component.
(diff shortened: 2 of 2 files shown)


jmuzina:
Is there a reason this PR is adding `.js` files for the generator (`instead of `.ts`), and adding them to the old generator directory `apps/generator-canonical-ds` instead of `packages/generator-ds`?

packages/generator-ds/README.md
```diff
@@ -3,9 +3,11 @@
 ## Getting Started
 
 1. Install [Yeoman](https://yeoman.io/) and this generator: `npm i -g yo @canonical/generator-ds`
-2. Run the generator: `yo @canonical/ds`
-3. Select a subgenerator from the list and follow its instructions
+2. Run the generator: `yo @canonical/ds:component`
```
jmuzina, line 6: This isn't technically correct for the root generator readme - someone should just invoke `yo @canonical/ds` which will prompt them to select a specific generator such as `:component`.

That prompt doesn't appear currently as we only have one generator. I guess it's a question of simplicity. Should we document the recommended use to be `:component`  for now as we only have one generator, and change it back to `yo @canonical/ds` when there are more generators?

packages/generator-ds/src/component/index.ts
```diff
@@ -58,40 +62,14 @@ export default class ComponentGenerator extends Generator<ComponentGeneratorOpti
...
     if (!this.answers) return;
 
-    const componentName = casing.toPascalCase(
-      path.basename(this.answers.componentPath),
+    const componentName = path.basename(this.answers.componentPath);
+
+    invariant(
+      casing.isPascalCase(componentName),
+      `The component name ${componentName} must be in PascalCase.`,
     );
```
jmuzina, line 73: This makes the README for the React component generator `(generator-ds/src/component/README.md)` inaccurate, as it recommends running `yo @canonical/ds:component path/to/component`, which would break this invariant.

Also, that component-specific README should be updated to remove the `--withStyles` and `--withStories` flags, as we are relying on `--help` to be authoritative parameter documentation.

jmuzina (commented):
Minor documentation-related comments

jmuzina (approved):

PR #128, opened 2025-02-11
feat(forms): Added a base package for the form components

advl:
## Done

- Added a package for the form components. This will serve as a primitive to implement [the form specification](https://docs.google.com/document/d/1-b7e2txdSvHay5IIlkaLebowhC_5acO5zSoH8uD__lU/edit?tab=t.0) after its approval.
- I took advantage to remove `.gitignore` and `bun.lockb` files _inside the packages_. I wonder how these files got there and why they weren't caught before (?)

## QA

- `bun install`
- `cd packages/react/ds-core-form && bun run storybook`
- `bun run build` from the root level

### PR readiness check

### DIFF
--- a/packages/react/ds-core-form/package.json
+++ b/packages/react/ds-core-form/package.json
@@ -29,7 +29,7 @@
     "check:biome": "biome check src *.json",
     "check:biome:fix": "biome check --write src *.json",
     "check:ts": "tsc --noEmit",
-    "storybook": "storybook dev -p 6006 --no-open --host 0.0.0.0",
+    "storybook": "storybook dev -p 6008 --no-open --host 0.0.0.0",
     "test": "bun run test:vitest",
     "test:watch": "bun run test:vitest:watch",
     "test:vitest": "vitest run",


packages/react/ds-core-form/src/ui/Button/Button.tsx
jmuzina, line 1: Why do we need to copy the `react-ds-core` Button component here? Couldn't we import it from the core lib if necessary?

packages/react/ds-core-form/src/ui/Chip/Chip.tsx
jmuzina, line 1: Similar comment as with Button; why is this component being redeclared instead of imported from the core?

packages/react/ds-core-form/src/ui/assets/accessibility.png
jmuzina, line 1: Similar to Button and Chip, looks like this folder has been copied from the React core and could be deleted from the form package?

packages/react/ds-core-form/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@canonical/react-ds-core-form",
```
jmuzina, line 2: In the spec, this package is referred to as `@canonical/react-ds-addon-forms`, should this be renamed?

At the least, I suspect "core" should not be in the name unless we anticipate multiple React form packages.

packages/react/ds-core-form/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "build:storybook": "storybook build",
+    "build:package": "bun run build:package:tsc && bun run build:package:copycss",
+    "build:package:copycss": "copyfiles -u 1 src/ui/**/*.css dist/esm",
+    "build:package:tsc": "tsc -p tsconfig.build.json",
+    "check": "bun run check:biome && bun run check:ts",
+    "check:fix": "bun run check:biome:fix && bun run check:ts",
+    "check:biome": "biome check src *.json",
+    "check:biome:fix": "biome check --write src *.json",
+    "check:ts": "tsc --noEmit",
+    "storybook": "storybook dev -p 6006 --no-open --host 0.0.0.0",
```
jmuzina, line 32: Can we use a different port? DS core and launchpad are currently using 6006 (even those two should probably be using different ports). 

The baseline grid addon is using 6007. This could be 6008.

```suggestion
    "storybook": "storybook dev -p 6008 --no-open --host 0.0.0.0",
```

packages/react/ds-core-form/src/ui/Configure.mdx
jmuzina, line 1: Duplicated from ds-core

packages/react/ds-core-form/src/ui/Typography.stories.tsx
jmuzina, line 1: Duplicated from ds-core

jmuzina (commented):
This seems to largely be a duplicate of the react-ds-core package, with a new name. I think if we're to implement the form package, we shouldn't add it to the repo until it has separate functionality from the react-ds-core package. Let me know if I am missing something, as I can't tell what this is solving in its current state.

jmuzina (approved):

PR #130, opened 2025-02-12
add base FileTree component

goulinkh:
## Done

- Add base skeleton for the `FileTree` component

## QA

- [Add QA steps]

### PR readiness check

- [ ] PR should have one of the following labels:
- [ ] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 

## Screenshots

### DIFF
--- a/packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
+++ b/packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
@@ -1,0 +1,96 @@
+import type React from "react";
+import { useCallback, useReducer } from "react";
+import Context from "./Context.js";
+import "./styles.css";
+import type {
+  FileSelectionOptions,
+  ProviderOptions,
+  SearchOptions,
+} from "./types.js";
+
+type State = {
+  internalSearchQuery: SearchOptions["searchQuery"];
+  internalSelectedFile: FileSelectionOptions["selectedFile"];
+};
+
+type Action =
+  | { type: "SEARCH"; payload: SearchOptions["searchQuery"] }
+  | { type: "SELECT_FILE"; payload: FileSelectionOptions["selectedFile"] };
+
+function reducer(state: State, action: Action): State {
+  switch (action.type) {
+    case "SEARCH":
+      return { ...state, internalSearchQuery: action.payload };
+    case "SELECT_FILE":
+      return { ...state, internalSelectedFile: action.payload };
+    default:
+      return state;
+  }
+}
+
+const componentCssClassName = "ds file-tree";
+
+const Provider = ({
+  id,
+  className,
+  style,
+  children,
+  searchQuery = "",
+  selectedFile = null,
+  // Controlled props
+  onSearch,
+  onSelectFile,
+}: ProviderOptions): React.ReactElement => {
+  const [state, dispatch] = useReducer(reducer, {
+    internalSearchQuery: searchQuery,
+    internalSelectedFile: selectedFile,
+  });
+
+  const currentSearchQuery = onSearch ? searchQuery : state.internalSearchQuery;
+  const currentSelectedFile = onSelectFile
+    ? selectedFile
+    : state.internalSelectedFile;
+
+  const onSearchHandler: SearchOptions["onSearch"] = useCallback(
+    (query) => {
+      if (onSearch) {
+        onSearch(query);
+      } else {
+        dispatch({ type: "SEARCH", payload: query });
+      }
+    },
+    [onSearch]
+  );
+
+  const onSelectFileHandler: FileSelectionOptions["onSelectFile"] = useCallback(
+    (node) => {
+      if (onSelectFile) {
+        onSelectFile(node);
+      } else {
+        dispatch({ type: "SELECT_FILE", payload: node });
+      }
+    },
+    [onSelectFile]
+  );
+
+  return (
+    <div
+      id={id}
+      style={style}
+      className={[componentCssClassName, className].filter(Boolean).join(" ")}
+    >
+      <Context.Provider
+        value={{
+          searchQuery: currentSearchQuery,
+          onSearch: onSearchHandler,
+          selectedFile: currentSelectedFile,
+          onSelectFile: onSelectFileHandler,
+        }}
+      >
+        {children}
+      </Context.Provider>
+    </div>
+  );
+};
+
+export default Provider;
--- a/packages/react/ds-app-launchpad/src/ui/FileTree/Context.tsx
+++ b/packages/react/ds-app-launchpad/src/ui/FileTree/Context.tsx
@@ -1,0 +1,6 @@
+import { createContext } from "react";
+import { ContextOptions } from "./types.js";
+
+const Context = createContext<ContextOptions | null>(null);
+
+export default Context;
--- a/packages/react/ds-app-launchpad/src/ui/FileTree/FileTree.stories.tsx
+++ b/packages/react/ds-app-launchpad/src/ui/FileTree/FileTree.stories.tsx
@@ -1,0 +1,58 @@
+/* @canonical/generator-ds 0.9.0-experimental.1 */
+
+import { FileTree } from "./index.js";
+// Needed for function-based story, safe to remove otherwise
+// import type { FileTreeProps } from './types.js'
+import type { Meta, StoryObj } from "@storybook/react";
+// Needed for template-based story, safe to remove otherwise
+// import type { StoryFn } from '@storybook/react'
+
+const meta = {
+  title: "",
+  component: FileTree,
+} satisfies Meta<typeof FileTree>;
+
+export default meta;
+
+/*
+  CSF3 story
+  Uses object-based story declarations with strong TS support (`Meta` and `StoryObj`).
+  Uses the latest storybook format.
+*/
+type Story = StoryObj<typeof meta>;
+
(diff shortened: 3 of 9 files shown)


packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
```diff
@@ -0,0 +1,96 @@
...
+      className={[componentCssClassName, className].filter(Boolean).join(" ")}
+    >
+      <Context.Provider
+        value={{
+          searchQuery: currentSearchQuery,
+          onSearch: onSearchHandler,
+          selectedFile: currentSelectedFile,
+          onSelectFile: onSelectFileHandler,
+        }}
+      >
```
jmuzina, line 89: Consider memoizing the provider context value. Many of these values are handled with useCallback already but the context value as a whole hasn't been memoized, this could help performance

```tsx
const contextValue = useMemo(() => ({
  searchQuery: currentSearchQuery,
  onSearch: onSearchHandler,
  selectedFile: currentSelectedFile,
  onSelectFile: onSelectFileHandler,
 }), [currentSearchQuery, onSearchHandler, currentSelectedFile, onSelectFileHandler]);
/// 
 return (
   // .....other jsx....
   <Context.Provider value={contextValue}>
    {children}
   </Context.Provider>
  // ..... other jsx.....
 );
```

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
```diff
@@ -0,0 +1,96 @@
...
+  | { type: "SELECT_FILE"; payload: FileSelectionOptions["selectedFile"] };
+
+function reducer(state: State, action: Action): State {
+  switch (action.type) {
+    case "SEARCH":
+      return { ...state, internalSearchQuery: action.payload };
+    case "SELECT_FILE":
+      return { ...state, internalSelectedFile: action.payload };
+    default:
+      return state;
```
jmuzina, line 27: Should this throw an error if an unexpected action type?

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
```diff
@@ -0,0 +1,96 @@
...
+  }
```
jmuzina, line 28: Perhaps the default case should create a `console.warn` in development environments if an unrecognized action type is dispatched?

goulinkh: removed the reducer in favor of external controlled state, this makes the component implementation and maintenance easier.
I will include usage examples in the stories.

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const onSearchHandler: SearchOptions["onSearch"] = useCallback(
+    (query) => {
+      if (onSearch) {
+        onSearch(query);
+      } else {
+        dispatch({ type: "SEARCH", payload: query });
+      }
+    },
+    [onSearch]
+  );
```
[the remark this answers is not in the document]
goulinkh: removed the reducer in favor of external controlled state, this makes the component implementation and maintenance easier.

This optimization should happen outside of this component.

goulinkh:
Hey @advl @jmuzina this PR is ready of review.

I followed this example as a reference regarding the keyboard accessibility: https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/

With the exception of the usage of `ul`, in my implementation all files are rendered flat on one level, for easier implementation as it is simpler to query sibling elements than nested elements.

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.stories.tsx
```diff
@@ -0,0 +1,137 @@
+/* @canonical/generator-ds 0.9.0-experimental.1 */
+
+import type { Meta, StoryFn } from "@storybook/react";
+import { useState } from "storybook/internal/preview-api";
+import { FileTree } from "./index.js";
+import type { FileTreeData, ProviderOptions } from "./types.js";
+
+const meta = {
+  title: "FileTree",
```
jmuzina, line 9: It feels a bit confusing to me that this component is named in the file-structure as `FileTree` but is called `TreeView` in the stories, is there a strong reason for that?

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
```diff
@@ -0,0 +1,127 @@
...
+  ...contextOptions
+}: ProviderOptions): React.ReactElement => {
+  // const focusedNodeRef = useRef<HTMLLIElement>(null);
+  const fileTreeRef = useRef<HTMLDivElement>(null);
+
+  const handleFocusNextNode: ManagedContextOptions["focusNextNode"] =
+    useCallback((direction) => {
+      if (!fileTreeRef.current) return;
+      const currentNode = fileTreeRef.current.querySelector<HTMLElement>(
+        "[data-path-hash]:focus",
```
jmuzina, line 23: `data-path-hash` literal is repeated several times, I'd consider extracting it to a constant

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
```diff
@@ -0,0 +1,127 @@
...
+      );
+      const currentNodeHash = currentNode?.getAttribute("data-path-hash");
+      if (!currentNodeHash) return;
+
+      const nextNode =
+        direction === "up"
+          ? fileTreeRef.current.querySelector<HTMLElement>(
+              `[data-path-hash]:has(+ [data-path-hash="${currentNodeHash}"])`,
```
jmuzina, line 31: `:has` is relatively new to browser baselines. It currently sits at ~93% support per [caniuse](https://caniuse.com/css-has). I wonder if there might be an implementation approach that has wider browser support that you could use, but ultimately this is a pretty high level of effort for this PR so I would suggest taking it up as a separate issue after this merges.

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
```diff
@@ -0,0 +1,127 @@
...
+        focusNextNode: handleFocusNextNode,
+        focusEndNode: handleFocusEndNode,
+        focusNextSiblingCharacter: handleFocusNextSiblingCharacter,
+      }}
+    >
+      <div
+        id={id}
+        style={style}
+        className={[componentCssClassName, className].filter(Boolean).join(" ")}
+        ref={fileTreeRef}
```
jmuzina, line 119: I believe we might need `role="tree"` here for a11y purposes

packages/react/ds-app-launchpad/src/ui/FileTree/common/Node/Provider.tsx
```diff
@@ -0,0 +1,232 @@
...
+        break;
+      case "Home":
+        focusEndNode("start");
+        event.preventDefault();
+        break;
+      case "End":
+        focusEndNode("end");
+        event.preventDefault();
+        break;
+      case "Enter":
```
jmuzina, line 140: I think something may be off with the "ENTER" key handling here. When I use the keyboard to navigate your file tree, and press "ENTER" on a node, it is not selected. I'd expect that a file would be selected and a folder would be expanded/collapsed when I press enter.

jmuzina (commented):
A few minor comments here, otherwise I'm happy with this. I believe @advl also has some final comments that he'll submit after his leave.

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.stories.tsx
goulinkh: `TreeView` is also a thing but is a part of `FileTree`, so `FileTree.TreeView` which is a render helper for when consuming a file tree from an API.

packages/react/ds-app-launchpad/src/ui/FileTree/common/Node/Provider.tsx
goulinkh: The key `Enter` is responsible for selecting a node.
In my case I only need selectable files, I have no use cases for selecting folders so I limited selection to files only.
Which results in Hitting Enter on folder has no effect. For expansion/collapse the keys `ArrowLeft` and `ArrowRight` are  used.

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.stories.tsx
jmuzina: Makes sense, thanks

packages/react/ds-app-launchpad/src/ui/FileTree/Provider.stories.tsx
```diff
@@ -0,0 +1,137 @@
+/* @canonical/generator-ds 0.9.0-experimental.1 */
+
+import type { Meta, StoryFn } from "@storybook/react";
+import { useState } from "storybook/internal/preview-api";
+import { FileTree } from "./index.js";
```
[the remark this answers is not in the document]
goulinkh: `FileTree` is different than `Provider` in this case, `FileTree` has sub components attached to the `Provider`.

PR #139, opened 2025-02-27
feat: add RelativeTime component

goulinkh:
## Done

- add RelativeTime component

Fixes LP-2285

## QA

- [Add QA steps]

### PR readiness check

## Screenshots

image


packages/react/ds-app-launchpad/src/ui/RelativeTime/styles.css
jmuzina, line 1: This file can be safely deleted if you don't plan on adding any custom CSS styles.

packages/react/ds-app-launchpad/src/ui/RelativeTime/RelativeTime.tsx
```diff
@@ -0,0 +1,143 @@
...
+    return () => {
+      if (intervalRef.current) window.clearInterval(intervalRef.current);
+    };
+  }, [updateLive, getOptimalUpdateInterval, updateTime]);
+
+  return (
+    <time
+      id={id}
+      style={style}
+      className={[componentCssClassName, className].filter(Boolean).join(" ")}
```
jmuzina, line 135: The `time` element should include a [`datetime` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#datetime) that contains a valid time string, to make this element more machine-readable for SEO & A11y purposes.

packages/react/ds-app-launchpad/src/ui/RelativeTime/RelativeTime.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+const meta = {
+  title: "",
+  component: Component,
+} satisfies Meta<typeof Component>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Default: StoryFn<typeof meta> = (args) => {
```
jmuzina, line 16: It would be nice to either add a story for an invalid date or just add it as an additional date in this story, that way we can see what happens when the date is invalid.

goulinkh:
Thanks for the review @advl I've applied the suggested changes, let me know if this is ready to go.

packages/react/ds-app-launchpad/src/ui/RelativeTime/RelativeTime.tsx
```diff
@@ -0,0 +1,106 @@
...
+const RelativeTime = ({
+  id,
+  className,
+  style,
+  time,
+  relativeTimeFormat,
+  nowThreshold = 10,
+  disableLiveUpdate = false,
+  nowKeyword = "now",
+  invalidDateKeyword = "invalid date",
```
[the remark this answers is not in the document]
goulinkh: I think this is fine, since the developer can always control the props to test different behaviors.

The props don't a difference in terms of return value and UI, so I don't think it is worth having different stories for different props in a `loop for each prop: create story` mindset.

packages/react/ds-app-launchpad/src/ui/RelativeTime/RelativeTime.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Temporal } from "@js-temporal/polyfill";
+import type { Meta, StoryFn } from "@storybook/react";
+import Component from "./RelativeTime.js";
+
+const meta = {
+  title: "RelativeTime",
+  tags: ["autodocs"],
+  component: Component,
+  parameters: {
+    // this component has no visual representation, so we disable snapshot testing
```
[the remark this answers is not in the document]
goulinkh: it is true that the time changing will make it challenging to test visually, however the reason for disabling the visual test is the fact that this has no styling overhead.

PR #143, opened 2025-02-28
feat(form): Ft form boilerplate pt2

advl:
## Done

- Adds state registration
- Adds shared types for inputs
- Improves types in general
- Added boilerplate (not Q/Aed) common components
- Added a form decorator for stories
- Added name in stories for inputs

This PR does not need to be Q/Aed for boilerplate code.

## QA

- N/A

### PR readiness check


packages/react/ds-core-form/src/storybook/decorators.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const FormWrapper: React.ElementType = () => {
+      const methods = useForm({
+        mode: "onChange",
+        defaultValues: initialData,
+      });
+
+      const onSubmit = (data: Record<string, unknown>) => {
+        console.log("[FORM SUBMIT]", data);
+      };
+
```
jmuzina, line 19: I imagine this is temporary, and later a user will pass their own submit handler, correct?

packages/react/ds-core-form/src/ui/Field/types.ts
```diff
@@ -19,19 +24,16 @@ export enum InputType {
...
-export interface FieldProps {
-  /* A unique identifier for the Field */
-  id?: string;
-  /* Additional CSS classes */
-  className?: string;
-  /* Child elements */
-  children?: React.ReactNode;
-  /* Inline styles */
-  style?: React.CSSProperties;
+export type UnionInputProps = CheckboxProps | TextProps | TextareaProps;
```
jmuzina, line 27: Generally, I'd try to avoid describing the type itself in the type name. "Union" is a set theory descriptor of the type itself, which seems a bit semantically clunky. 

Might we be able to use something simpler, like `InputProps`?

packages/react/ds-core-form/src/ui/Field/inputs/Checkbox/Checkbox.t-est.tsx
jmuzina, line 1: Are these t-est files named like so to ignore tests temporarily?

packages/react/ds-core-form/src/ui/Field/inputs/Textarea/Textarea.t-ests.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { beforeEach, describe, expect, it } from "vitest";
+import Component from "./Textarea.js";
+
+describe("Textarea component", () => {
+	// beforeEach((context) => {
+	//   context.test = decorators.form()(context.test);
+	// });
+	it("renders", () => {
+		render(<Component name="content">Textarea</Component>);
+		expect(screen.getByText("Textarea")).toBeInTheDocument();
```
jmuzina, line 14: This is currently causing a failing build. I think this is because you have renamed your test files to `t-ests.tsx`, which is not excluded by `tsconfig.build.json`. If you intend to push these, I'd suggest either getting the tests to a place where they pass by functionality or bypassing the tests in a more standardized/declarative way, like this:

```tsx
describe("Textarea component", () => {
  it("test bypass", () => {
    expect(true).toBe(true);
  });
});

You will also need to rename the files back to `*.tests.tsx` so they are excluded by the build tsconfig.

This should allow your tests to "pass" while also avoiding a build failure.

jmuzina (commented):
Some initial comments

jmuzina (approved):

PR #155, opened 2025-03-18
feat(ci): Chromatic CI Stage 1 : Reusable Workflows and base deployments

advl:
## Done

- Added [chromatic projects](https://www.chromatic.com/apps?accountId=67bdd0b0d5ea0caee8ffc953) to chromatic.com (preparation)
- Added project variables in settings
- Added a reusable workflow in the common folder of `.github/workflows`
- Added three consuming workflows for `react-ds-core`, `react-ds-core-form`, `react-ds-app-launchpad`

## QA

- The workflows are untested

### PR readiness check

### DIFF
--- a/.github/workflows/chromatic._template.yml
+++ b/.github/workflows/chromatic._template.yml
@@ -32,6 +32,7 @@ jobs:
         run: bun run test
   publish:
     runs-on: ubuntu-latest
+    needs: build
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4


.github/workflows/_chromatic.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: ./.github/actions/setup-env
+
+      - name: Publish to Chromatic
+        uses: chromaui/action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          exitZeroOnChanges: true
+          workingDir: ${{ inputs.working-directory }}
+          buildScriptName: "build:storybook"
```
jmuzina, line 50: Chromatic [recommends](https://www.chromatic.com/docs/github-actions/#maintain-a-clean-main-branch) we consider the main branch to be "auto-accepted" - whatever is merged to main should be an approved visual diff.

With Percy, this would mean running a separate test on pushes to `main`, but Chromatic makes this a bit easier: all we need to do is use `autoAcceptChanges: "main".

```suggestion
          buildScriptName: "build:storybook"
          autoAcceptChanges: "main"
```

.github/workflows/chromatic.ds-react-core-form.yml
```diff
@@ -0,0 +1,16 @@
+name: "ds-react-core/Publish to chromatic"
+on:
+  push:
+    branches:
+      - '**'  # This excludes tags
```
jmuzina, line 5: This is interesting, it will run tests on *any* push to these packages, even if there's not a pull request.

This has the upside of maintaining always-up-to-date versions of each branch, even feature branches.

However, it also means we will be running a lot more tests (such as people's work-in-progress feature branches that aren't yet PR ready).

I think it will be OK to change this to a pull request trigger.

.github/workflows/chromatic.ds-react-core-form.yml
```diff
@@ -0,0 +1,16 @@
...
+defaults:
+  run:
+    working-directory: packages/react/ds-core
+jobs:
+  test-compile:
+    uses: ./.github/workflows/_chromatic.yaml
+    secrets:
+      PROJECT_TOKEN: ${{ secrets.CHROMATIC_TOKEN_CANONICAL_REACT_DS_CORE_FORM }}
+    with:
+      working-directory: packages/react/ds-core
```
jmuzina, line 16: Might we be able to simplify the repetition of package dirs a bit by using a workflow-level env var?

```
# define this at the top of the workflow, near the workflow triggers
env:
  WORKING_DIRECTORY: packages/react/ds-core-form
# other yaml...
# 
# Inside an invocation of `_chromatic.yaml`
with: 
  working-directory: ${{ env.WORKING_DIRECTORY }}
```

jmuzina:
@advl I've left some code comments, and answered your open questions below:

>     * The reusable workflow needs to be at the root level, otherwise they will [error](https://github.com/canonical/ds25/actions/runs/13933818964) as here with `Invalid workflow file: .github/workflows/chromatic.ds-react-core.yml#L12 invalid value workflow reference: workflows must be defined at the top level of the .github/workflows/ directory`

I don't think this is a problem, as long as the calling workflows' names are primarily scoped to the `chromatic` namespace, like you've done here.

>     * I am unsure about the workflow naming. I am thinking to use a "." as domain separator, but I am unsure whether the main domain should be the workflow name, or the package/app name.

The "." is a good domain separator. As mentioned above, I think the workflow name should be the main domain.

>     * It seems that, tentatively, it is not possible to "dry" further the calls, where the folder name needs to be repeated three times.

I've left a [code comment ](https://github.com/canonical/ds25/pull/155#discussion_r2002110073) that might be able to help with this. Ultimately though, I'm not concerned if this remains like this, as these files shouldn't be modified often once they're in place.

>     * Atm, the workflows are triggered on `push` but I wonder if we wouldn't want instead to trigger on PR or push to main ?

Both `push` and `pull_request` are valid. We will definitely want `pull_request` as this will allow outside contributors to have jobs run - `push` only considers upstream branches. Also, `push` will likely run more tests than we need (we'll be deploying all upstream branches, even if they have no proposed changes to `main`).

.github/workflows/chromatic.ds-react-core-form.yml
```diff
@@ -0,0 +1,11 @@
+name: "react-ds-core/chromatic"
+on:
+  pull_request:
+    paths: [packages/react/ds-core/**]
```
jmuzina, line 4: Core and form have been mixed up
```suggestion
    paths: [packages/react/ds-core-form/**]
```

.github/workflows/chromatic.ds-react-core-form.yml
```diff
@@ -0,0 +1,11 @@
...
+jobs:
+  test-compile:
+    uses: ./.github/workflows/chromatic._template.yaml
+    secrets:
+      PROJECT_TOKEN: ${{ secrets.CHROMATIC_TOKEN_CANONICAL_REACT_DS_CORE_FORM }}
+    with:
+      working-directory: packages/react/ds-core
```
jmuzina, line 11: Core and form have been mixed up
```suggestion
      working-directory: packages/react/ds-core-form
```

.github/workflows/chromatic.ds-react-core-form.yml
```diff
@@ -0,0 +1,11 @@
+name: "react-ds-core/chromatic"
```
jmuzina, line 1:
```suggestion
name: "react-ds-core-form/chromatic"
```

.github/workflows/chromatic.ds-react-core.yml
```diff
@@ -0,0 +1,11 @@
+name: "react-ds-core-form/chromatic"
```
jmuzina, line 1:
```suggestion
name: "react-ds-core/chromatic"
```

.github/workflows/chromatic.ds-react-core.yml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  pull_request:
+    paths: [packages/react/ds-core-form/**]
```
jmuzina, line 4: Core and form have been mixed up
```suggestion
    paths: [packages/react/ds-core/**]
```

.github/workflows/chromatic.ds-react-core.yml
```diff
@@ -0,0 +1,11 @@
...
+jobs:
+  test-compile:
+    uses: ./.github/workflows/chromatic._template.yaml
+    secrets:
+      PROJECT_TOKEN: ${{ secrets.CHROMATIC_TOKEN_CANONICAL_REACT_DS_CORE }}
+    with:
+      working-directory: packages/react/ds-core-form
```
jmuzina, line 11: Core and form have been mixed up
```suggestion
      working-directory: packages/react/ds-core
```

.github/workflows/chromatic.ds-react-app-launchpad.yml
```diff
@@ -0,0 +1,11 @@
+name: "react-ds-app-launchpad/chromatic"
+on:
+  pull_request:
+    paths: [packages/react/ds-app-launchpad/**]
+jobs:
+  test-compile:
+    uses: ./.github/workflows/chromatic._template.yaml
```
jmuzina, line 7: Tested in #159 , the [job fails](https://github.com/canonical/ds25/actions/runs/13949753860). I think it's because there is a mismatch between name of the template file in the call and the file name itself - the extension is yml but you've used yaml here.

```suggestion
    uses: ./.github/workflows/chromatic._template.yml
```

.github/workflows/chromatic.ds-react-core-form.yml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  pull_request:
+    paths: [packages/react/ds-core/**]
+jobs:
+  test-compile:
+    uses: ./.github/workflows/chromatic._template.yaml
```
jmuzina, line 7:
```suggestion
    uses: ./.github/workflows/chromatic._template.yml
```

.github/workflows/chromatic.ds-react-core.yml
```diff
@@ -0,0 +1,11 @@
+name: "react-ds-core-form/chromatic"
+on:
+  pull_request:
+    paths: [packages/react/ds-core-form/**]
+jobs:
+  test-compile:
+    uses: ./.github/workflows/chromatic._template.yaml
```
jmuzina, line 7:
```suggestion
    uses: ./.github/workflows/chromatic._template.yml
```

.github/workflows/chromatic._template.yml
```diff
@@ -0,0 +1,50 @@
...
+
+        # "prepare" hook on `bun install` runs `bun run build`
+      - name: Setup environment
+        uses: ./.github/actions/setup-env
+
+      - name: Publish to Chromatic
+        uses: chromaui/action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
```
jmuzina, line 47: Something is going wrong with passing this token from the react core workflow to the template, see this [error](https://github.com/canonical/ds25/actions/runs/13950596087)

```

Invalid workflow file: .github/workflows/chromatic.ds-react-core.yml#L7
The workflow is not valid. .github/workflows/chromatic.ds-react-core.yml (Line: 7, Col: 11): Secret CHROMATIC_PROJECT_TOKEN is required, but not provided while calling. .github/workflows/chromatic.ds-react-core.yml (Line: 9, Col: 22): Invalid secret, PROJECT_TOKEN is not defined in the referenced workflow.
```

I haven't tested the other two packages, but they may have the same problem

.github/workflows/chromatic._template.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Setup environment
+        uses: ./.github/actions/setup-env
+
+      - name: Code quality checks
+        run: bun run check
+
+      - name: Test
+        run: bun run test
+  publish:
+    runs-on: ubuntu-latest
```
jmuzina, line 34: Currently the publish and build run in parallel, is that your intention?

Otherwise we can make them run in sequence with `needs`:

```suggestion
    runs-on: ubuntu-latest
    needs: build
```

.github/workflows/chromatic._template.yml
jmuzina: Fixed by @advl

jmuzina (approved):

PR #162, opened 2025-03-20
feat(boilerplate): Add storybook

advl:
## Done

Adds storybook to the boilerplate
- add dependencies to package.json
- add the base storybook config
- adapter the tsconfig for storybook related files. This also solves, indirectly, the error where `index.css` cannot be resolved.
- Added an example component - this is necessary for the storybook to run

Note : This PR can likely be used as an example of the modifications necessary to add storybook to an application 

## QA

- cd `apps/react/boilerplate-vite`
- `bun run storybook`,
- `bun run build:storybook`

### PR readiness check


apps/react/boilerplate-vite/storybook-static/canonical.svg
jmuzina, line 1: I think you will want to gitignore the `storybook-static` folder, there are many many files here that we probably don't need version controlled as they are build artifacts

jmuzina (commented):
Nice work! Could you also setup the boilerplate for Chromatic visual testing since we are adding a storybook?

PR #167, opened 2025-03-20
feat(ds-core-form): boilerplate pt 4

advl:
## Done

- Adds Range Input
- Adds native Select Input
- Adds SimpleChoices input (fieldset of radios/comboboxes)
- Adds `useOptionAriaProperties` hook in an initial stage
- Fixes `useFieldAriaProperties` call of `useId`
- Enhances props of components with a union to native HTML properties (e.g. `export type TextareaProps = BaseInputProps & React.HTMLProps<HTMLTextAreaElement>;  `)
- _drive_by:_ Added a note about the autodocs on the storybook package

## QA

- ! Atm, there are a bunch of type errors. Not ready for final Q/A yet.

### PR readiness check

### DIFF
--- a/.github/workflows/chromatic._template.yml
+++ b/.github/workflows/chromatic._template.yml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Test
         run: bun run test
-  publish:
+  publish-to-chromatic:
     runs-on: ubuntu-latest
     needs: build
     steps:


packages/react/ds-core-form/src/ui/Field/inputs/Range/Range.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <input
+        id={id}
+        type="range"
+        className={[componentCssClassName, className].filter(Boolean).join(" ")}
+        style={style}
+        {...otherProps}
+        {...register(name, registerProps)}
+      />
+      <output htmlFor={id} className="ds range-output">
+        {value}
```
jmuzina, line 35: I noticed that the output is currently empty until the user changes the slider value (no initial value is loaded).

Even when I modified the range story to use a pre-filled value, the output was not populated until after I interacted with the form.

packages/react/ds-core-form/src/ui/Field/inputs/Range/Range.tsx
```diff
@@ -0,0 +1,41 @@
...
+  name,
+  registerProps,
+  ...otherProps
+}: RangeProps): React.ReactElement => {
+  const { register, watch } = useFormContext();
+  const value = watch(name);
+  return (
+    <>
+      <input
+        id={id}
```
jmuzina, line 27: I think something is going wrong with the aria labelling - 

```html
<div class="ds form-wrapper">
  <div id=":r0:-age-description" class="ds description"></div>
  <label id=":r0:-age-label" for="form-age" class="ds label">age</label>
  <input id=":r0:-age" class="ds range" aria-labelledby=":r0:-age-label" aria-describedby=" :r0:-age-label" aria-errormessage=":r0:-age-error" aria-invalid="false" min="0" max="100" step="1" type="range" value="3" name="age">
  <output for=":r0:-age" class="ds range-output">5</output>
</div>
```

1. The `for=` in your label element is not mapping to the input element. This presents an a11y problem.
2. the aria-describedby for the input element itself also has an extra space at the front of it: `aria-describedby=" :r0:-age-label"`
3. `aria-errormessage` prop should not be present when the field is in a valid state.

jmuzina (approved):

PR #169, opened 2025-03-20
fix(React core): Popup placement accounts for target margin, target m…

### LINKED ISSUE #168: `useWindowFitment` includes target margins in popup placement calculation
The current `useWIndowFitment` hook assumes that trigger elements had no margin in order to calculate popup placement.

Consider this screenshot of a button and its margin:
Screenshot 2025-03-20 at 16 40 07

Using the bounding client rect of the target as a whole would consider the button margin as part of the target size. This causes, for example, top-positioned tooltips to be off-center due to button right margin, and right-positioned tooltips to be positioned at the outer edge of the target margin, as shown in the following two screenshots:
Screenshot 2025-03-20 at 16 39 57
Screenshot 2025-03-20 at 16 46 29

Currently, I've gotten around this by collapsing the margin on tooltip target elements. This correctly positions the tooltip, but disrupts the flow of the tooltip target:
Screenshot 2025-03-20 at 16 47 27
Screenshot 2025-03-20 at 16 49 38

To both correctly position the tooltip and avoid disrupting content flow, we should not consider target margin in positioning calculation.

jmuzina:
To correctly position the tooltip and avoid disrupting the flow of content by ignoring target margins, I have adjusted the `useWindowFitment` hook to consider the **internal** sizing (the bounding box of the target's children, rather than the target itself) for positioning calculations.

## Done

Fixes #168 

## QA

- Checkout PR
- Open React Core storybook
- See that the buttons in the tooltip examples have the expected button margin, and the tooltips are still correctly positioned.

### PR readiness check

## Screenshots

Screenshot 2025-03-20 at 16 39 13

Screenshot 2025-03-20 at 16 52 02

### DIFF
--- a/packages/react/ds-core/src/ui/Tooltip/common/TooltipArea/styles.css
+++ b/packages/react/ds-core/src/ui/Tooltip/common/TooltipArea/styles.css
@@ -4,10 +4,5 @@
     cursor: pointer;
     position: relative;
     display: inline-block;
-
-    & > * {
-      /* The tooltip targets must have no margin in order for tooltip placements to work correctly */
-      margin: 0 !important;
-    }
   }
 }
--- a/packages/react/ds-core/src/ui/hooks/useWindowFitment/useWindowFitment.tsx
+++ b/packages/react/ds-core/src/ui/hooks/useWindowFitment/useWindowFitment.tsx
@@ -61,6 +61,42 @@ const useWindowFitment = ({
   }, [gutter, windowDimensions, isServer]);
 
   /**
+   * Gets the internal sizing of an element.
+   * The internal sizing of an element is the maximum bounding client rect of its children.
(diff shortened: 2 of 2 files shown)


jmuzina:
I'm not sure how we could simplify this markup without losing functionality, and we certainly need to support the target having margin: it would be quite a common use case to apply a tooltip to a component that has some spacing applied (like buttons).

A similar approach (used by MUI's dependency Popper) is to [calculate target margin](https://github.com/vusion/popper.js/blob/d6f744bf9219e09892d1f76e6c9002fd2995ace5/packages/popper/src/utils/getPopperOffsets.js#L18) and apply that as a positioning offset. This would likely be a bit simpler than calculating child element bounding box. 

We could see further simplification by making and documenting the assumption that there is only one child element. This is done by [Shoelace](https://shoelace.style/components/tooltip).

PR #195, opened 2025-04-09
feat(tokens): Base token build pipeline with style dictionary

advl:
## Done

- Adds a `tokens` package
- Adds a basic folder structure for both primitive and semantic tokens
- Adds a build system for both primitives and semantic tokens. 

## QA

- `bun run i`
- `bun run build`
- Review files in created dist folder. The semantic tokens should be class-based mappings. (Leave `motion` on the side as it's meant to be replaced with mappings)

### PR readiness check

## Screenshots

### DIFF
--- a/packages/tokens/config.semantic.js
+++ b/packages/tokens/config.semantic.js
@@ -2,52 +2,38 @@ import StyleDictionary from "style-dictionary";
 
 StyleDictionary.registerFormat({
   name: "css/semantic-selectors",
-  format: async ({ dictionary, options, file }) => {
-    let output = "";
-    let index = 0;
-    // Iterate over top-level groups in the token tree
-    for (const groupKey in dictionary.tokens) {
-      const group = dictionary.tokens[groupKey];
-      const selector = group.$extensions?.selector || `.${groupKey}`;
-      const tokens = dictionary.allTokens.filter(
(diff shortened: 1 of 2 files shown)


packages/tokens/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@canonical/tokens",
+  "description": "Canonical design tokens: primitives and mode mappings for use with Style Dictionary",
+  "version": "0.9.0-experimental.1",
+  "main": "src/index.css",
+  "files": ["src"],
```
jmuzina, line 6: Should this be `dist`? 
```suggestion
  "files": ["dist"]
```

packages/tokens/config.semantic.js
jmuzina, line 1: The generated `semantics.css` seems to have some mismapped variables. For example:

```css
--background: var(--background-default); /* --background-default is not defined anywhere */
```

`primitives-canonical` does provide `--color-background-default`, not `--background-default`.


Screenshot 2025-04-11 at 10 46 35

Only the motion classes are bound to valid variables. Is this because those tokens will be provided by a followup PR?

jmuzina: I don't think you've pushed your changes - I don't see any new commits

jmuzina (approved):

PR #219, opened 2025-04-30
feat(form): tokens

advl:
## Done

- Provides `state` tokens both as primitives and semantic mappings (`tokens` package)
- Adds prefix/suffix to the Text Input, providing a div wrapper to accommodate those
- Consumes gap tokens in SimpleChoices and Text (with prefix/suffix)
- Consumes state tokens : error, disabled, valid, idle
- Parametrize the border

## Missing (for part 2)
- Adding padding and size of inputs
- Adding colors on brand

## QA

- [Add QA steps]

### PR readiness check

### DIFF
--- a/packages/react/ds-core-form/src/ui/Field/common/Wrapper/Wrapper.tsx
+++ b/packages/react/ds-core-form/src/ui/Field/common/Wrapper/Wrapper.tsx
@@ -3,7 +3,7 @@ import type React from "react";
 import { useMemo } from "react";
 import type { BaseInputProps, WrapperProps } from "../../types.js";
 import "./styles.css";
-import { states } from "ui/constants.js";
+import { states } from "../../../constants.js";
 import { Description, Error as FieldError, Label } from "../index.js";
 import { useFieldWrapper } from "./hooks/index.js";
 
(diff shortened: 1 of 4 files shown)


packages/react/ds-core-form/src/index.css
```diff
@@ -47,3 +54,9 @@
   --form-label-columns: 1 / -1;
   --form-payload-columns: 1 / -1;
 }
+
+/* Utility */
+
+.rtl {
+  direction: rtl;
+}
```
jmuzina, line 62: In which situations would we want to override the text direction with a utility class? Is it just for demonstrative purposes in storybooks?

jmuzina: If you define a `.rtl`, you should probably also define a `.ltr`

packages/react/ds-core-form/src/ui/constants.ts
jmuzina, line 1: Does this still create build errors? If so, we may need to switch to string constants or adjust ts build settings

packages/styles/modes/intents/src/index.css
```diff
@@ -1,3 +1,5 @@
+@import url("@canonical/tokens/dist/css/semantic.css");
```
jmuzina, line 1: What are your thoughts on export aliases to make this less dependent on file structure?

jmuzina (approved):

PR #221, opened 2025-04-30
fix chromatic cancelling itself

### LINKED ISSUE #220: Chromatic: CI job concurrency issues causes some jobs to cancel prematurely
We are intermittently seeing some Chromatic CI jobs cancel themselves before they have been completed. This seems to be more likely to occur the more packages are affected by a PR. Re-running the cancelled job always results in a pass.

Here is an example of a workflow that was cancelled: https://github.com/canonical/ds25/actions/runs/14764402236/attempts/1

The following message is shown by GitHub on the canceled workflow:
```
Canceling since a higher priority waiting request for 'react-ds-core/chromatic-refs/pull/202/merge' exists
```

This may have something to do with the [concurrency settings](https://github.com/canonical/ds25/blob/d8348986ba1e6feea6d3380ad128b13bf9611ccc/.github/workflows/chromatic._template.yml#L18-L21) for the chromatic template.

jmuzina:
## Done

Fixes #220 
Fixes https://warthogs.atlassian.net/browse/WD-21828

## QA

- See that multiple chromatic tests can run on this PR at once without cancelling themselves

### PR readiness check

### DIFF
--- a/.github/workflows/chromatic._template.yml
+++ b/.github/workflows/chromatic._template.yml
@@ -16,8 +16,7 @@ on:
         required: true
         description: Chromatic project token for publishing
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref
-    }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-${{ inputs.working-directory }}
   cancel-in-progress: true
 jobs:
   build:


jmuzina:
need to battle-test this by making a file change that causes many chromatic tests to run. test-editing a file in the react core should suffice. I'll try that out tomorrow.

jmuzina:
Looking good, every chromatic-enabled package was tested and none of them were cancelled. 

I'm going to push another similar commit to make sure nothing gets cancelled, then push two commits in quick succession to ensure that the concurrency groupings still work and tests get cancelled in the right situation.

PR #223, opened 2025-05-04
feat(DiffChangeMarker): add DiffChangeMarker component and include it in GitDiffViewer.FileHeader

goulinkh:
### Unrelated change

Add a custom color to the git diff viewer line numbers.

### PR readiness check

### DIFF
--- a/packages/react/ds-app-launchpad/src/ui/DiffChangeMarker/DiffChangeMarker.tsx
+++ b/packages/react/ds-app-launchpad/src/ui/DiffChangeMarker/DiffChangeMarker.tsx
@@ -1,9 +1,9 @@
 /* @canonical/generator-ds 0.9.0-experimental.12 */
 import type React from "react";
-import { useMemo } from "react";
-import * as icons from "./icons.js";
+import SimpleChangeMarker from "./common/SimpleChangeMarker/SimpleChangeMarker.js";
+import { DetailedChangeMarker } from "./common/index.js";
 import "./styles.css";
-import type { DiffChangeMarkerProps } from "./types.js";
+import type { DiffChangeMarkerProps, DiffChangeType } from "./types.js";
 
 const componentCssClassName = "ds diff-change-marker";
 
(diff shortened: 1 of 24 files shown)


packages/react/ds-app-launchpad/src/ui/DiffChangeMarker/DiffChangeMarker.tsx
```diff
@@ -0,0 +1,83 @@
...
+      return "modified";
+    }
+    if (autoOption?.additions) {
+      return "added";
+    }
+    if (autoOption?.deletions) {
+      return "deleted";
+    }
+    return "modified";
+  }, [autoOption?.additions, autoOption?.deletions, manualOption?.type]);
```
steciuk, line 41: I would suggest getting rid of this `useMemo`, as it:

- returns a string which is referentially stable if the value stays the same,
- doesn't contain any heavy computation and just a couple of ifs.

packages/react/ds-app-launchpad/src/ui/DiffChangeMarker/DiffChangeMarker.tsx
```diff
@@ -0,0 +1,83 @@
...
+  id,
+  className,
+  style,
+  markerStyle,
+  ...markerOptions
+}: DiffChangeMarkerProps): React.ReactElement => {
+  const displayStyle = markerStyle;
+  const isManual = "type" in markerOptions;
+  const manualOption = isManual ? markerOptions : undefined;
+  const autoOption = isManual ? undefined : markerOptions;
```
steciuk, line 25: The `manualOption`, `autoOption` referencing the same object, but conditionally and mutually exclusive being set to `undefined`, complex prop union type, the conditional in `useMemo` and in the template make this simple component unnecessarily hard to reason about.

In my opinion, all those things feel as if it supposes to be two separate components (`Manual...` and `Auto...`). They can (and possibly should) both use a common styling utility component.

packages/react/ds-app-launchpad/src/ui/DiffChangeMarker/styles.css
```diff
@@ -0,0 +1,79 @@
...
+      background-color: var(--diff-change-marker-modified-alt);
+    }
+  }
+
+  &.style-detailed {
+    font-size: var(--diff-change-marker-font-size);
+    font-weight: var(--diff-change-marker-font-weight);
+    line-height: var(--diff-change-marker-line-height);
+
+    & > span {
```
steciuk, line 60: Do we use the nesting selector before combinators or not? We have one here, but there is none in the line 34: `> svg {`.

packages/react/ds-app-launchpad/src/ui/DiffChangeMarker/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type DiffChangeMarkerProps = {
+  /* A unique identifier for the DiffChangeMarker */
+  id?: string;
+  /* Additional CSS classes */
+  className?: string;
+  /* Inline styles */
+  style?: React.CSSProperties;
+} & MarkerOptions &
+  (DiffChangeAutoOptions | DiffChangeManualOptions);
```
steciuk, line 27: As outlined in the comment in `DiffChangeMarker.tsx` – maybe we should split this component into two separate ones?

PR #224, opened 2025-05-04
refactor(FileTree): Enhance structure and styling of FileTree component

goulinkh:
### PR readiness check


packages/react/ds-app-launchpad/src/ui/FileTree/Provider.tsx
```diff
@@ -102,6 +109,18 @@ const Provider = ({
...
+  const searchBoxes: React.ReactNode[] = [];
+  const treeItems: React.ReactNode[] = [];
+
+  Children.forEach(children, (child) => {
+    if (isValidElement(child) && child.type === SearchBox) {
+      searchBoxes.push(child);
+    } else {
+      treeItems.push(child);
+    }
+  });
```
steciuk, line 122: Is this a common practice in case of composed library components? Seems a bit brittle. Let's say someone wants to wrap a `SearchBox` with a `div` before passing it to the `FileTree` (for styling or any other reasons). Wouldn't it break?

I suggest defining a separate prop `searchbox(es)` or even `header` and skip on using `Children` and manually filtering the children. That, in my opinion, would give the library consumer a proper degree of freedom while keeping the component code more robust.

packages/react/ds-app-launchpad/src/ui/FileTree/common/SearchBox/styles.css
```diff
@@ -6,6 +6,12 @@
   border-bottom: 1px solid var(--file-tree-search-box-border-color);
   padding: var(--file-tree-search-box-vertical-padding)
     var(--file-tree-search-box-horizontal-padding);
+  background-color: var(--file-tree-node-background-color);
+
+  position: sticky;
+  top: 0;
+  z-index: 5;
```
steciuk, line 13: Is this `z-index` needed? I believe that the less `z-index` shenanigans the better for the code maintainability, and in this case it seems like it isn't necessary.

steciuk, line 13: It feels like that kind of styles should be set by the parent rather than on the component itself.

packages/react/ds-app-launchpad/src/ui/FileTree/common/SearchBox/styles.css
```diff
@@ -6,6 +6,12 @@
...
+
   &:has(.input:focus) {
```
steciuk, line 15: `:focus-within` instead? It's cleaner and a bit more performant, I think

steciuk (approved):

PR #225, opened 2025-05-05
feat(ds-core-form): Middleware examples, MSW, Stories

advl:
## Done

- Added MSW to `ds-core-form` to mock back-end requests, both globally and at a story level, tested in `msw.stories.tsx`. Please note I decided against using "msw-storybook-addon" as it seems to be falling out of maintenance, creating a dependnecy issue to upgrade to the upcoming Storybook v9.
- Added stories to test `addRESTOptions` and `addRESTValidation` middlewares, together with MSW mocks.
- Refactored `addConditionalDisplay` logic, splitting it in a HOC and a middleware. The reason to be for the middleware, is to not encourage the growth of the base `Field` api, providing a more declarative, opt-in api for conditional display. In the background, the middleware only spans the HOC `withConditionalDisplay` - which itself executes the logic.

## QA

- `bun run storybook`
- Review the `msw` stories in the `Test` folder. Both stories should be functional.
- Review the `middleware` stories. They should all be functional.

### PR readiness check

### DIFF
--- a/packages/react/ds-core-form/src/ui/Field/common/Wrapper/withWrapper.ts
+++ b/packages/react/ds-core-form/src/ui/Field/common/Wrapper/withWrapper.ts
@@ -1,69 +1,69 @@
 import * as React from "react";
 import { useMemo } from "react";
 import type {
-	BaseInputProps,
-	BaseWrapperProps,
-	Condition,
-	WrappedComponentProps,
-	WrapperProps,
+  BaseInputProps,
+  BaseWrapperProps,
+  Condition,
+  WrappedComponentProps,
+  WrapperProps,
 } from "../../types.js";
 import DefaultWrapper from "./Wrapper.js";
 import withConditionalDisplay from "./withConditionalDisplay.js";
 
 type WrappedComponentPropsInternal<ComponentProps extends BaseInputProps> =
-	Omit<WrappedComponentProps<ComponentProps>, "Component"> & {
-		condition?: Condition;
-	};
+  Omit<WrappedComponentProps<ComponentProps>, "Component"> & {
+    condition?: Condition;
+  };
 
 const withWrapper = <
-	ComponentProps extends BaseInputProps,
-	ComponentWrapperProps extends
-		BaseWrapperProps<ComponentProps> = WrapperProps<ComponentProps>,
+  ComponentProps extends BaseInputProps,
+  ComponentWrapperProps extends
+    BaseWrapperProps<ComponentProps> = WrapperProps<ComponentProps>,
 >(
-	Component: React.ComponentType<ComponentProps>,
-	options?: Partial<ComponentWrapperProps>,
-	Wrapper: React.ComponentType<WrapperProps<ComponentProps>> = DefaultWrapper,
-	// Ideally, we should use something like this, but this creates a type error for the WrappedComponentProps
-	// Wrapper: React.ComponentType<ComponentWrapperProps> = DefaultWrapper as React.ComponentType<ComponentWrapperProps>,
+  Component: React.ComponentType<ComponentProps>,
+  options?: Partial<ComponentWrapperProps>,
+  Wrapper: React.ComponentType<WrapperProps<ComponentProps>> = DefaultWrapper,
+  // Ideally, we should use something like this, but this creates a type error for the WrappedComponentProps
+  // Wrapper: React.ComponentType<ComponentWrapperProps> = DefaultWrapper as React.ComponentType<ComponentWrapperProps>,
 ) => {
-	const MemoizedComponent = React.memo(Component);
+  const MemoizedComponent = React.memo(Component);
 
-	function WrappedComponent({
-		middleware = [],
-		WrapperComponent = Wrapper,
-		...props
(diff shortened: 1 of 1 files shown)


packages/react/ds-core-form/.storybook/preview.ts
```diff
@@ -1,19 +1,47 @@
...
+
+import { worker } from "mocks/browser.js";
+
+import { type ComponentType, createElement, useEffect } from "react";
 
 import "index.css";
 
+const withMSW = (Story: StoryFn, context: StoryContext) => {
+	useEffect(() => {
+		console.log("Story decorator called with ctx", context);
```
jmuzina, line 13:
```suggestion
```
remove console log if not needed

packages/react/ds-core-form/public/mockServiceWorker.js
jmuzina, line 1: this is quite a bit of setup that seems generally abstract from this package. I wonder if it would be worth moving it to its own package so that other packages and teams can benefit from it more easily.

packages/react/ds-core-form/src/ui/Field/common/Wrapper/withWrapper.ts
```diff
@@ -1,16 +1,18 @@
...
   WrappedComponentProps,
   WrapperProps,
 } from "../../types.js";
 import DefaultWrapper from "./Wrapper.js";
+import withConditionalDisplay from "./withConditionalDisplay.js";
 
 type WrappedComponentPropsInternal<ComponentProps extends BaseInputProps> =
-  Omit<WrappedComponentProps<ComponentProps>, "Component">;
+  Omit<WrappedComponentProps<ComponentProps>, "Component"> & {
+    condition?: [string[], (depsValues: unknown[]) => boolean];
```
jmuzina, line 14: can you extract this condition type into its own named type, so that this is more readable?

jmuzina:
MSW stories are currently showing 404 in Chromatic
image

requests to `/api/test` are returning 404.

packages/react/ds-core-form/src/storybook/msw.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+    useEffect(() => {
+      fetch("/api/test")
+        .then((res) => res.json())
+        .then((data) => setData(data));
+    }, []);
+
+    return (
+      <div>
+        <p>
+          Should return "{decoratorResponsePayload}", incidating MSW
```
jmuzina, line 48:
```suggestion
          Should return "{decoratorResponsePayload}", indicating MSW
```
typo

README.md
```diff
@@ -37,8 +37,11 @@ To learn more, check out the following resources:
 | [Contribution Documentation](./guides/CONTRIBUTING.md) | Learn how to contribute to DS-25. Find detailed instructions for installing, running, and working with the project locally.             |
 | [Architecture Documentation](./guides/ARCHITECTURE.md) | Explore the architectural decisions and principles behind DS-25. This includes explanations of project structure, benchmarks, and more. |
 
+## 4. Caveats
+For the time being, node 23 seems to provoke [an error](./issues/226). Use node v22 for the time being, for instance with `nvm use 22`.
```
jmuzina, line 41: This issue link is a dead link when viewed directly in github
https://github.com/canonical/ds25/blob/c716db422ecf2fe1a2f3c5d5ebfd892c539b2ed3/README.md

It may be worth using an absolute URL.

packages/react/ds-core-form/package.json
```diff
@@ -33,7 +33,8 @@
     "test": "bun run test:vitest",
     "test:watch": "bun run test:vitest:watch",
     "test:vitest": "vitest run",
-    "test:vitest:watch": "vitest"
+    "test:vitest:watch": "vitest",
+    "init:msw": "npx msw init public/"
```
jmuzina, line 37: I don't see this command executed anywhere, is it "magic" invoked by the worker script?

packages/react/ds-core-form/public/mockServiceWorker.js
jmuzina, line 1: Is this a dist file from MSW? Why do we need to include it directly in our version control like this?

jmuzina: You're right - quite a strange setup but I see the benefits.

packages/react/ds-core-form/package.json
jmuzina: It would be nice to include a note in the README about this command and what scenarios we might need to re-run it in. For instance, if upgrading versions of the MSW script. That way, the relationship between this command, the worker file, and the package are a bit more clear to somebody with less exposure to MSW.

packages/react/ds-core-form/src/ui/Field/types.ts
```diff
@@ -108,6 +108,11 @@ export type WrapperHOCAdditionalProps<
    * An optional wrapper component to render around the input.
    */
   WrapperComponent?: React.ComponentType<ComponentWrapperProps>;
+
+  /**
+   * A tuple determining whether to render the input
+   */
+  condition?: [string[], (depsValues: unknown[]) => boolean];
```
jmuzina, line 115: Use your `Condition` type here

PR #239, opened 2025-05-23
feat(shi): Semantic DS Schema and base tooling for automation (WIP)

advl:
## Done

- Implements a base version of the Design System Schema (Ontology) using RDF Turtle
- Provides an example fixture for component specifications
- Refactors the class-based initial cli command approach into a functional approach, offering both a command api and a programmatic api (in js), with four commands
  - `specification create`
  - `specification improve`
  - `implementation create`
  - `implementation verify`

To do : 
- Document
- Add tests
- Change name to a better one. I was thinking "blue" as "blueprint" ? Also as this tool is not DS-specific but could leverage any schema ?

## QA

- [Add QA steps]

### PR readiness check

### DIFF
--- a/packages/shi/src/cli.ts
+++ b/packages/shi/src/cli.ts
@@ -1,70 +1,146 @@
-import fs from "node:fs";
+import { parseArgs } from "util";
 import { operations } from "./operations.js";
 import { shi } from "./shi.js";
 
-const parseArgs = (args: string[]) => {
-  const options: Record<string, string> = {};
-  for (let i = 0; i < args.length; i += 2) {
-    const key = args[i].replace(/^-+/, ""); // Remove leading dashes (- or --)
-    const value = args[i + 1];
-    if (!value) {
-      throw new Error(`Missing value for option: ${key}`);
+// Simple help generation from operations
+const showHelp = () => {
+  console.log("Usage: shi <entity> <operation> [options]\n");
+  console.log("Commands:");
+
+  for (const [entity, ops] of Object.entries(operations)) {
+    for (const [opName, op] of Object.entries(ops)) {
+      console.log(`  ${entity} ${opName} - ${op.description}`);
+      for (const opt of op.options) {
+        console.log(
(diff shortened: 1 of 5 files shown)


packages/shi/src/cli.ts
```diff
@@ -0,0 +1,146 @@
+import { parseArgs } from "util";
```
jmuzina, line 1: `util` is built into NodeJS so it doesn't generally require an entry in `package.json` dependencies, so I don't think there's really anything to change. But it makes me think about our approach to documenting these sorts of dependencies, as it could effect which versions of Node our project is compatible with. 

I imagine this could at some point become a table of dependencies and node versions, mapping out which features are supported by which node versions, to give more visibility/clarity to our Node. version support.

packages/shi/src/cli.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+  }
+
+  console.log("Global options:");
+  console.log("  --provider <n>  LLM provider (openai, anthropic)");
+  console.log("  --model <n>     Model to use");
+  console.log("  -o, --output <file> Write output to file");
+  console.log("  -q, --quiet        Suppress progress messages");
+  console.log("  -h, --help         Show this help");
+  console.log("  -v, --version      Show version");
```
jmuzina, line 28: It may be more scalable to log these with some mapping function over a predictable data structure of cli options, rather than hard-coding them. But this can be addressed later.

packages/shi/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "type": "git",
+    "url": "https://github.com/canonical/ds25"
+  },
+  "license": "LGPL-3.0",
+  "bugs": {
+    "url": "https://github.com/canonical/ds25/issues"
+  },
+  "homepage": "https://github.com/canonical/ds25#readme",
+  "scripts": {
+    "build": "tsc -p tsconfig.build.json",
```
jmuzina, line 30: I'm not 100% familiar with ttl, but is there some way of validating that the code is valid? Is there some build step to increase confidence in code in a given PR? How do we validate its output programmatically?

PR #243, opened 2025-05-30
chore: Enable TICS

jmuzina:
Improve compliance with SDLC by enabling TICS for DS25

**Still needed**: I'm not sure if there is any other setup required to enable TICS, such as configuring it on the TICS backend ourselves or requesting someone else to do it. This is just an adapation of the [react components TICS workflow](https://github.com/canonical/react-components/blob/811db9f701194ebd145969a50bf0e5ebe315ff98/.github/workflows/tics-coverage.yml) for DS25.

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -1,0 +1,26 @@
+name: Upload TICS report
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  tics-report:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


jmuzina:
Asking the TICS group [here](https://chat.canonical.com/canonical/pl/3wm88eafoif8updtg7pucgz5zr) if there are any further steps needed to configure TICS.

jmuzina:
There might be a better way to encapsulate it, but as far as I'm aware, this is the standard way across WE that teams invoke TICS. Growing away from it might make us an outlier

PR #249, opened 2025-06-06
feat(svelte): Add linting and code formatting

steciuk:
This PR addresses the investigation into Biome support for Svelte formatting and linting. Since [Biome currently only support formatting `<script>` tags and does not lint Svelte files](https://biomejs.dev/internals/language-support/#html-super-languages-support), Prettier and ESLint have been set up instead. The Prettier configuration is aligned as closely as possible with project's Biome config.

When Biome rolls out full support for Svelte, we should switch to using it instead. Biome + Svelte discussions to follow:
- https://github.com/biomejs/biome/discussions/6276
- https://github.com/biomejs/biome/issues/4726

## Done

- Added Prettier and ESLint for formatting `@canonical/svelte-ds-app-launchpad` package.
- Adjusted Prettier config to align formatting output with the project's conventions.
- Added `check` and `check:fix` scripts compatible with the global project's linting/formatting scripts.

## QA

- Make some random changes in `.ts`, `.css` and `.svelte` files in `@canonical/svelte-ds-app-launchpad` package.
- Check if the following commands give the expected results:
  - `bun run check` in the package directory
  - `bun run check:fix` in the package directory
  - `bun run check` in the project's root.
  - `bun run check:fix` in the project's root.

### PR readiness check


steciuk:
Sorry for not responding, I was on a leave yesterday.

I agree that extracting the configs to `configs` seems to be the best way to move forward.  Next week I'll follow with separate PRs to do just that. I have some doubts about whether (and how) to include Svelte specific options in the configs, but this will be a topic for the discussion at said PRs.

steciuk:
After some discussion, we've decided that we think a better approach would be to first experiment some more with the ESLint and Prettier in our Svelte project. Only then, when we are happy with the setup (probably next week, or week after that) we will upstream the configs (in the way you recommended) and switch to consuming them.

I am closing this PR for now. Let me know if that's something you are ok with.

PR #260, opened 2025-06-13
docs: add standards for code

jmuzina:
Updates and refines standards for code to follow recently agreed structure for standards files.


jmuzina:
How to capture cases like `cli.ts` that generally is executed as a process and not as a provider of code for other modules?

PR #267, opened 2025-06-18
feat(generator): Generate SSR output test file

steciuk:
## Done
Following [the recommendation from the previous PR](https://github.com/canonical/pragma/pull/248#pullrequestreview-2922002322), the addition of the SSR tests file generation for Svelte was extracted into a separate PR.

## QA

- Run `yo @canonical/ds:sv-component`.
- Confirm that the SSR test file got generated for the new Svelte component and has the expected contents.

### PR readiness check


steciuk:
> What is the recommended approach you would propose for React ? Would it be simple enough to be included in this PR ?

I added a similarly minimal SSR tests file template to the React generator and to give an example of how it could work I:
- added SSR tests to `react-ds-core`,
- generated and adjusted an example SSR test for the `Button` component.

Running `bun run test` for the `ds-core` package now gives the following results:

image

@advl, what do you think of this approach?

steciuk:
I do believe that pointless tests are worse than lack of them, so having a flag makes sense :)

However, I also see the value in encouraging developers to at least having in mind SSR when creating new components. For that reason, I lean more towards `--noSSRTests`, which makes skipping them an explicit choice. On the other hand, this wouldn't be consistent with the current generator interface (`--withStories`, `--withStyles`).

@advl, what's your take on that?

PR #268, opened 2025-06-19
feat(webarchitect): Minimal implementation of the webarchitect tool

advl:
## Done

Webarchitect is a tool that compares a package structure against an established schema. It is meant to be used in the context of a (our) monorepo, to verify that our packages implement the correct apis. Different uses cases involve : 
- verifying that a configuration file is present (for instance biome.json or package.json)
- verifying that the configuration files implement the right fields (for instance package.json has a "version" or "scripts")
- verifying that the configuration files fields have the right values (for instance biome.json extends the base canonical config)
At the moment, webarchitect only tests .json files against the ruleset/schema - providing a base to be extended to test plain text files as well 

## Note

- What is in the `schemas` directory is a pure wget from the original schemas - they are used to contextualize/initialize ajv prior to validating our schemas.

## QA

- After running bun install, webarchitect should be run using `bun run webarchitect [ruleset]` where [ruleset] is one of `base`, `base-package` and `package-react` or a ruleset in `.json` present in the local folder. Base is at the moment empty (meant to include a license, etc.. but plain text checking is not implemented for now). 
- Try `bun run webarchitect base-package` in different packages of the monorepo
- Try `bun run webarchitect package-react` in the react design system packages `packages/react/ds-core` 
- Try the same commands with `--verbose` or `--json`
- Try `bun run webarchitect --list` to list available rulesets. Rulesets now must have the extension `.ruleset.json` to be recognized.

### PR readiness check

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1184,7 +1184,7 @@
 
     "@types/body-parser": ["@types/body-parser@1.19.6", "", { "dependencies": { "@types/connect": "*", "@types/node": "*" } }, "sha512-HLFeCYgz89uk22N5Qg3dvGvsv46B8GLvKKo1zKG4NybA8U2DiEO3w9lqGg29t/tfLRJpJ6iQxnVw4OnB7MoM9g=="],
 
-    "@types/bun": ["@types/bun@1.2.18", "", { "dependencies": { "bun-types": "1.2.18" } }, "sha512-Xf6RaWVheyemaThV0kUfaAUvCNokFr+bH8Jxp+tTZfx7dAPA8z9ePnP9S9+Vspzuxxx9JRAXhnyccRj3GyCMdQ=="],
+    "@types/bun": ["@types/bun@1.2.19", "", { "dependencies": { "bun-types": "1.2.19" } }, "sha512-d9ZCmrH3CJ2uYKXQIUuZ/pUnTqIvLDS0SK7pFmbx8ma+ziH/FRMoAq5bYpRG7y+w1gl+HgyNZbtqgMq4W4e2Lg=="],
 
     "@types/chai": ["@types/chai@5.2.2", "", { "dependencies": { "@types/deep-eql": "*" } }, "sha512-8kB30R7Hwqf40JPiKhVzodJs2Qc1ZJ5zuT3uzw5Hq/dhNCl3G3l83jfpdI1e20BP348+fV7VIL/+FxaXkqBmWg=="],
 
@@ -1390,7 +1390,7 @@
 
     "buffer-from": ["buffer-from@1.1.2", "", {}, "sha512-E+XQCRwSbaaiChtv6k6Dwgc+bx+Bs6vuKJHHl5kox/BaKbhiXzqQOwK4cO22yElGp2OCmjwVhT3HmxgyPGnJfQ=="],
 
-    "bun-types": ["bun-types@1.2.18", "", { "dependencies": { "@types/node": "*" }, "peerDependencies": { "@types/react": "^19" } }, "sha512-04+Eha5NP7Z0A9YgDAzMk5PHR16ZuLVa83b26kH5+cp1qZW4F6FmAURngE7INf4tKOvCE69vYvDEwoNl1tGiWw=="],
+    "bun-types": ["bun-types@1.2.19", "", { "dependencies": { "@types/node": "*" }, "peerDependencies": { "@types/react": "^19" } }, "sha512-uAOTaZSPuYsWIXRpj7o56Let0g/wjihKCkeRqUBhlLVM/Bt+Fj9xTo+LhC1OV1XDaGkz4hNC80et5xgy+9KTHQ=="],
 
     "byte-size": ["byte-size@8.1.1", "", {}, "sha512-tUkzZWK0M/qdoLEqikxBWe4kumyuwjl3HO6zHTr4yEI23EojPtLYXdG1+AQY7MN0cGyNDvEaJ8wiYQm6P2bPxg=="],
 


packages/webarchitect/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+### Ruleset Resolution Mechanism
+
+The `<ruleset>` parameter follows a specific resolution order that determines where webarchitect looks for ruleset definitions:
+
+1. **Local Files First** - Webarchitect first checks the current directory for a file matching your ruleset name. If you specify `my-rules`, it looks for `my-rules.json` in the current directory. This allows you to override built-in rulesets or use project-specific validation rules.
+
+2. **Bundled Rulesets Second** - If no local file is found, webarchitect searches its internal collection of built-in rulesets. These include `base`, `base-package`, and `package-react`. Built-in rulesets are distributed with the tool and provide standard validation patterns for common project types.
+
+3. **Remote URLs** - You can also specify a complete URL (starting with `http://` or `https://`) to fetch rulesets from remote locations. Remote rulesets must end with `.json` and be accessible via standard HTTP requests. This enables sharing rulesets across organizations or using centrally managed validation rules.
```
jmuzina, line 121: Do you plan to account for remote rulesets that are behind an auth requirement?

packages/webarchitect/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+```typescript
+import { validate } from '@canonical/webarchitect';
+import type { ValidationResult } from '@canonical/webarchitect';
+
+// Validate the current directory against a ruleset
+async function checkProjectCompliance(): Promise<void> {
+  try {
+    // The validate function takes a project path and ruleset identifier
+    const results: ValidationResult[] = await validate(process.cwd(), 'base-package');
```
jmuzina, line 143: Is there any sort of validation metadata that might live at the process level, rather than the rule level? If so, that might change the return of validate() to some object that has multiple properties, like `rules` (the current array of results, for rule-scoped info) and additional properties, for instance `passed` (could save the user from having to manually check all of the rules with `Array.some()`, which is a very common use case)

packages/webarchitect/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "files": ["dist", "rulesets"],
+  "author": {
+    "email": "webteam@canonical.com",
+    "name": "Canonical Webteam"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/canonical/ds25"
+  },
+  "license": "LGPL-3.0",
```
jmuzina, line 20: Double check our licensing page, I think this might be intended as a standalone tool so it needs GPL3

https://library.canonical.com/legal/licensing-policy

packages/webarchitect/src/executeValidationRules.ts
```diff
@@ -0,0 +1,39 @@
+import type { Schema, ValidationResult } from "./types.js";
+import validateDirectoryRule from "./validateDirectoryRule.js";
+import validateFileRule from "./validateFileRule.js";
+
+export default async function executeValidationRules(
```
jmuzina, line 5: Please add tsdoc

packages/webarchitect/src/listDirectory.ts
```diff
@@ -0,0 +1,14 @@
+import { readdir } from "node:fs/promises";
+
+export default async function listDirectory(
```
jmuzina, line 3: Tsdoc

packages/webarchitect/src/loadFullSchema.ts
```diff
@@ -0,0 +1,18 @@
+import resolveSchema from "./resolveSchema.js";
+import type { Schema } from "./types.js";
+
+export default async function loadFullSchema(
```
jmuzina, line 4: Tsdoc

packages/webarchitect/src/resolveSchema.ts
```diff
@@ -0,0 +1,82 @@
...
+import ajv from "./ajv.js";
+import motherSchema from "./schema.json" with { type: "json" };
+import type { Schema } from "./types.js";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const validateSchema = ajv.compile(motherSchema);
+
+function formatValidationErrors(errors: ErrorObject[]): string {
```
jmuzina, line 14: Tsdoc

packages/webarchitect/src/resolveSchema.ts
```diff
@@ -0,0 +1,82 @@
...
+      const uniqueMessages = [...new Set(messages)]; // Remove duplicates
+      const property = path === "/" ? "root" : path.replace("/", "");
+      return `  ${property}: ${uniqueMessages.join(", ")}`;
+    })
+    .join("\n");
+
+  return `Schema validation failed:\n${formattedErrors}`;
+}
+
+export default async function resolveSchema(
```
jmuzina, line 41: Tsdoc on all functions, especially exported functions please

packages/webarchitect/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  rule: string;
+  passed: boolean;
+  message?: string;
+  // Verbose context information
+  context?: {
+    type: "file" | "directory";
+    target: string; // The file or directory path being validated
+    description?: string; // Human-readable description of what was checked
+    schema?: unknown; // The schema that was applied (for verbose output)
+    actualValue?: unknown; // What was found (for verbose output)
```
jmuzina, line 41: `actual` in a property adds a bit of clutter, is there a reason we can't just use value?

packages/webarchitect/src/validateDirectoryRule.ts
```diff
@@ -0,0 +1,137 @@
+import { stat } from "node:fs/promises";
+import { join } from "node:path";
+import listDirectory from "./listDirectory.js";
+import type { DirectoryRule, ValidationResult } from "./types.js";
+import validateFileRule from "./validateFileRule.js";
+
+export default async function validateDirectoryRule(
```
jmuzina, line 7: Tsdoc

packages/webarchitect/README.md
jmuzina: It's 100% out of scope right now - just picking your brain. My mind went there because of the wording "standard HTTP requests" - i wondered what "standard" meant.

packages/webarchitect/rulesets/base-package.json
```diff
@@ -0,0 +1,97 @@
...
+          "extends": {
+            "type": "array",
+            "items": { "const": "@canonical/biome-config" },
+            "minItems": 1,
+            "maxItems": 1
+          },
+          "files": {
+            "type": "object",
+            "properties": {
+              "include": {
```
jmuzina, line 23:
```suggestion
              "includes": {
```

Now that we're on Biome v2, the package checks are failing:

```
✗ FAILED VALIDATIONS
──────────────────────────────────────────────────
✗ biome
  Validation failed:
  • files: must NOT have additional properties

✓ PASSED VALIDATIONS
──────────────────────────────────────────────────
✓ package
✓ package-react

Summary: 1/3 validations failed
```

packages/webarchitect/src/validateFileRule.ts
```diff
@@ -0,0 +1,176 @@
...
+ *   }
+ * };
+ * const results = await validateFileRule("/project", rule, "package-config");
+ * ```
+ */
+export default async function validateFileRule(
+  projectPath: string,
+  fileRule: FileRule,
+  ruleName: string, // Add rule name parameter
+): Promise<ValidationResult[]> {
```
jmuzina, line 95: This returns an array of validation results, but you only ever return an array of one result, and each error case short-circuit returns an error result. 

Maybe you should construct an array `results` and push each result onto it, so that all errors are returned, rather than just whichever one came first in your validation

packages/webarchitect/src/validateDirectoryRule.ts
```diff
@@ -0,0 +1,160 @@
...
+          rule: ruleName, // Use the rule name from schema
+          passed: false,
+          message: `Expected directory but found file: ${dirPath}`,
+          context: {
+            type: "directory" as const,
+            target: dirPath,
+            description: `Validates directory structure for ${dirRule.name}`,
+            actualValue: "[File found instead of directory]",
+          },
+        },
```
jmuzina, line 53: same comment as in validateFileRule - these are constructed repetitively with the same values, would recommend DRYing

packages/webarchitect/src/validateDirectoryRule.ts
```diff
@@ -0,0 +1,160 @@
...
+ * const results = await validateDirectoryRule("/project", rule, "source-structure");
+ * ```
+ */
+export default async function validateDirectoryRule(
+  projectPath: string,
+  dirRule: DirectoryRule,
+  ruleName: string, // Add rule name parameter
+): Promise<ValidationResult[]> {
+  const dirPath = join(projectPath, dirRule.name);
+  const results: ValidationResult[] = [];
```
jmuzina, line 36: Your error cases all short-circuit return literal `ValidationResult` objects rather than append to this array, which means errors will be hidden if multiple are thrown. I'd recommend avoiding short-circuiting these errors and instead appending them to your results array.

packages/webarchitect/src/executeValidationRules.ts
```diff
@@ -0,0 +1,59 @@
...
+      } else if ("directory" in rule) {
+        // Pass the rule name context to directory validation
+        const dirResults = await validateDirectoryRule(
+          projectPath,
+          rule.directory,
+          ruleName,
+        );
+        results.push(...dirResults);
+      }
+    }
```
jmuzina, line 56: At the risk of abstracting too early, this looks like a decent candidate for abstraction. validateFileRule and validateDirectoryRule are very similar in structure and function. They are also called very similarly here. I could see this branching becoming a bit cumbersome and verbose in this file if more validation rule types are added in the future. 

Maybe we could add a `validateRule` function that, given a rule, calls the appropriate validator (rule, directory, any other we may come across). This would clean up this validation logic at this level and may also consolidate some of the logic in the file and directory validators.

jmuzina:
Currently, in `ds-core`, I get this error when running `bun run webarchitect base-package` and other rulesets that inherit

```
jmuzina@jmuzina-ubuntu-desktop:~/source/work/canonical/repos/ds25/packages/react/ds-core$ bun run webarchitect base-package
Loaded ruleset from: /home/jmuzina/source/work/canonical/repos/ds25/packages/webarchitect/rulesets/base-package.json
Loaded ruleset from: /home/jmuzina/source/work/canonical/repos/ds25/packages/webarchitect/rulesets/base.json

✗ FAILED VALIDATIONS
──────────────────────────────────────────────────
✗ package
  Validation failed:
  • license: must be equal to constant

✓ PASSED VALIDATIONS
──────────────────────────────────────────────────
✓ biome

Summary: 1/2 validations failed (1 passed)
error: "webarchitect" exited with code 1
```

PR #305, opened 2025-08-20
feat: Add Svelte SSR testing library renderer

goulinkh:
## Done
- SSR render function: wraps Svelte SSR renderer and exposes assert helpers
- Include unit tests
- Vite config that uses `@sveltejs/vite-plugin-svelte` instead of SvelteKit

Fixes [LP-2917](https://warthogs.atlassian.net/browse/LP-2917)

### PR readiness check

[LP-2917]: https://warthogs.atlassian.net/browse/LP-2917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -11,7 +11,6 @@
       "name": "@canonical/react-boilerplate-vite",
       "version": "0.10.0-experimental.0",
       "dependencies": {
-        "@canonical/react-ds-core": "^0.10.0-experimental.0",
         "@canonical/react-ssr": "^0.10.0-experimental.0",
         "@canonical/storybook-config": "^0.10.0-experimental.0",
         "@canonical/styles": "^0.10.0-experimental.0",
@@ -581,7 +580,6 @@
       "name": "@canonical/styles-primitives-canonical",
       "version": "0.10.0-experimental.0",
       "dependencies": {
-        "@canonical/tokens": "^0.10.0-experimental.0",
         "@canonical/typography": "^0.10.0-experimental.0",
         "normalize.css": "^8.0.1",
       },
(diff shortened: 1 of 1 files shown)


packages/svelte/ssr-test/src/render.ts
```diff
@@ -0,0 +1,24 @@
+import { getQueriesForElement, prettyDOM } from "@testing-library/dom";
+import { JSDOM } from "jsdom";
+import { render as renderSvelte } from "svelte/server";
+import type { RenderFunction } from "./types.js";
+
+export const render: RenderFunction = (...args: Parameters<RenderFunction>) => {
```
steciuk, line 6: I believe that this typing of the `render` function bakes in the _no-props_ conditional type branch, due to `Props` generic in `RenderFunction` being undefined, essentially disallowing providing component props. Please, correct me if that's not true.

Anyway, please add a fixture component that accepts props and a test that uses this `render` with props being passed in. Maybe something similar to what we have in the README:

```ts
    const { getByText, window, container } = render(Greeting, {
      props: { name: "World" },
    });
```

packages/svelte/ssr-test/src/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  body: string;
+  pretty: (maxLength?: number) => string;
+};
+
+type SvelteRenderOptions = {
+  context?: Map<unknown, unknown>;
+  idPrefix?: string;
+};
+
+export type RenderFunction = <
```
steciuk, line 21: Is there an option not to type our render function parameters ourselves but derive the typing from Svelte's `render`? I believe that we want to keep the input typing the same and only amend the return type, so this should be possible, right?

packages/svelte/ssr-test/vite.config.ts
```diff
@@ -0,0 +1,14 @@
+/// <reference types="vitest/config" />
+import { svelte } from "@sveltejs/vite-plugin-svelte";
+import { defineConfig, type PluginOption } from "vite";
+
+export default defineConfig({
+  plugins: [svelte() as PluginOption],
```
steciuk, line 6: Why is this type assertion necessary?

steciuk (changes requested):

packages/react/ds-app-launchpad/src/ui/GitDiffViewer/fixtures.tsx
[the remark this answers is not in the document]
goulinkh: Old changes committed by accident, will remove them.

packages/svelte/ssr-test/src/render.ts
goulinkh: The conditional type allows for making options argument optional if there are no required props in a component.
Without this conditional typing, you always have to provide options.props: `render(Div, {props: {}})`.

I also didn't want to implement this typing inside this package, but it's not possible to reuse Svelte internal type for render.

packages/svelte/ssr-test/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "check": "bun run check:biome",
+    "check:fix": "bun run check:biome:fix && bun run check:ts",
+    "check:biome": "biome check",
+    "check:biome:fix": "biome check --write",
+    "check:ts": "tsc --noEmit",
+    "test": "vitest run"
+  },
+  "peerDependencies": {
+    "@testing-library/dom": "^10.4.1",
+    "jsdom": "^26.1.0",
```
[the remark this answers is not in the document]
steciuk: _could be_, but not **have to be**. Wouldn't following this logic make all the dependencies `peerDependencies`? Nobody would use this package without having `svelte`, but there may be cases where someone doesn't have JSDom used, no? What package would you decide to put in `dependencies` then?

packages/svelte/ssr-test/vite.config.ts
goulinkh: There is a slight typing conflict where svelte is missing `warn, info, error` I think this is due to the vite version that I am using.

But casting the type explicitly somehow is helping with the error.

packages/svelte/ssr-test/package.json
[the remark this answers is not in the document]
goulinkh: Not sure about when would be a good time to use dependencies, maybe when a package depends on specific dependency version. I am not an NPM package maintainer so not sure about all cases.

But in this case, there a reason why all three deps should be peer deps.

packages/svelte/ssr-test/src/render.ts
steciuk: I've checked out the branch locally to confirm my suspicions. This demonstrates what I meant:

image
 
Do I misunderstand how this should be used?

packages/svelte/ssr-test/vite.config.ts
steciuk: Would it be possible to align the versions of the plugin and Vite instead?

packages/svelte/ssr-test/package.json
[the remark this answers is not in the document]
steciuk: > there a reason why all three deps should be peer deps

I don't think I understand what is that reason. Imagine someone has a project with vitest-browser based tests, doesn't have `jsdom`, doesn't have `@testing-library/dom`, and he thinks _oh! This `svelte-ssr-test` will make my SSR DX much better!_. He installs our package and... it doesn't work, requiring manual installation of the additional packages. On the other hand, a scenario in which the consumer doesn't have `svelte` installed is not possible, as he would not need **SVELTE**-ssr-test then. Does that make sense?

Take a look on `vite-plugin-react` for example: https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react/package.json

See that the `peerDependencies` contain only `vite` as this won't ever be used by someone who doesn't use `vite` in the first place? `react-refresh` or `@babel/core` on the other hand, may or may not be used in the project of the consumer.

Does this line of thinking make sense? If you disagree, I'll leave the decision whether to keep them in peer or move to dependencies for you to take :)

packages/svelte/ssr-test/src/render.ts
goulinkh: should be fixed now.

packages/svelte/ssr-test/vite.config.ts
goulinkh: should be fixed now

packages/svelte/ssr-test/src/types.ts
steciuk: What do you think about:

```ts
export const render = ((
  ...args
): ReturnType<typeof renderSvelte> & RenderResult => {
  const output = renderSvelte(...(args as [any]));
  const { window } = new JSDOM(
    `<!doctype html><html><body>${output.body}</body></html>`,
  );
  const { document } = window;
  const { body: container } = document;
  const queries = getQueriesForElement(container);

  return {
    ...queries,
    ...output,
    window,
    document,
    container,
    pretty: (maxLength = 10000) => prettyDOM(container, maxLength) || "",
  };
}) satisfies typeof renderSvelte;
```
This seems to work, allows us to not have to type `RenderFunction` ourselves and only type those things in `RenderResult` that we add.
```ts
export type RenderResult = Queries & {
  window: DOMWindow;
  document: DOMWindow["document"];
  container: HTMLElement;
  pretty: (maxLength?: number) => string;
};
```

packages/svelte/ssr-test/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "@canonical/typescript-config-base",
+  "compilerOptions": {
+    "baseUrl": "src",
+    "skipLibCheck": true,
+    "types": ["vitest/globals"],
+    "paths": {
+      "@fixtures/*": ["fixtures/*"]
+    }
```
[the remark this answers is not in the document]
goulinkh: yes left by mistake, will remove

PR #308, opened 2025-08-22
feat(tokens): add primitive and semantic color tokens

jmuzina:
## Done

Adds the rest of the [primitive and semantic color tokens](https://docs.google.com/spreadsheets/d/1AY6BLFU-Cnj48ggHUnNVlWRxwWZ6Z-66s7jxJC1HLSU/edit?gid=1738212977#gid=1738212977) to the tokens package.

A style dictionary integration is forthcoming pending some finalization of the token work - these tokens being defined in CSS directly is temporary and will allow me to build new components in the meantime. 

## QA

- See that there are no Chromatic diffs

### PR readiness check

### DIFF
--- a/packages/styles/primitives/canonical/src/tokens.css
+++ b/packages/styles/primitives/canonical/src/tokens.css
@@ -4,51 +4,242 @@
 :root {
   /* color tokens */
   /* based on: https://docs.google.com/spreadsheets/d/1AY6BLFU-Cnj48ggHUnNVlWRxwWZ6Z-66s7jxJC1HLSU/edit?gid=1738212977#gid=1738212977 */
-  --color-palette-white: #fff;
-  --color-palette-black: #000;
-  --color-palette-blue: #24598f;
-  --color-palette-yellow: #cc7900;
 
-  --color-background-default: var(--color-palette-white);
+  --color-palette-black: rgba(0, 0, 0, 1);
+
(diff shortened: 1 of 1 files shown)


jmuzina:
> Splitting primitives, semantic and component in three files

How would you recommend architecting this on a package basis? Currently, we have a `tokens.css` file inside a `@canonical/styles-primitives-canonical` - so in the long term this package should only hold primitives. The component and semantic tokens are only here temporarily. Is it worth adding new `component.css` and `semantic.css` files or even separate packages, if they're going to be replaced soon by a dedicated token pipeline?

> prefixing temporary tokens with --tmp, so they are easier to identify and grep

My understanding here would be that this would be all of the semantic and primitive tokens, do you share that understanding?

jmuzina:
The visual regressions actually come from before the `tmp-` prefixing, due to a difference in intents between how we've 
been using them and how they are defined in the color spreadsheet. We have been using button colors as global intent colors for some time, but the color tokens draft differentiates here.

`color-background-positive-default`, for example, is bound to the primitive `color-palette-green-1`, while `color-background-button-positive-default` is bound to a higher contrast `color-palette-green-8`. This means the buttons are now consuming global intents colors which are truly global, decreasing their contrast. 

Before:
Screenshot 2025-09-04 at 13 03 37

After:
Screenshot 2025-09-04 at 13 03 03

This means some components with specific intents tokens, like Button, would need to override their intent color variables.

PR #357, opened 2025-10-14
docs(Storybook): storybook standards

jmuzina:
## Done

Adds storybook standards based on our work so far.

## QA

- Read the standards.


advl:
LGTM

tough the comment above wasn't answered : 
>  I wonder if we'd like one for the 'as' prop consumption - as it requires an explicit helper function definition

advl (approved):

PR #423, opened 2026-02-17
feat(svelte-ds-app-launchpad): upstream Chip component and update styles

goulinkh:
## Done

- Added a new Chip component with various states and modifiers.
- Introduced new styles for density and severity modifiers.
- Removed the Example component and its related files.
- Updated README with usage instructions for the new Chip component.

Fixes [LP-3723](https://warthogs.atlassian.net/browse/LP-3723)

### PR readiness check

### Light

image

### Dark

image

[LP-3723]: https://warthogs.atlassian.net/browse/LP-3723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -100,10 +100,10 @@
       "name": "@canonical/storybook-config",
       "version": "0.15.1",
       "dependencies": {
-        "@canonical/ds-assets": "^0.14.0",
-        "@canonical/storybook-addon-baseline-grid": "^0.14.0",
-        "@canonical/storybook-addon-shell-theme": "^0.14.0",
-        "@canonical/styles-debug": "^0.14.0",
+        "@canonical/ds-assets": "^0.15.1",
+        "@canonical/storybook-addon-baseline-grid": "^0.15.1",
+        "@canonical/storybook-addon-shell-theme": "^0.15.1",
+        "@canonical/styles-debug": "^0.15.0",
         "@chromatic-com/storybook": "^5.0.1",
         "@storybook/addon-a11y": "^10.2.8",
         "@storybook/addon-docs": "^10.2.8",
@@ -115,8 +115,8 @@
       },
       "devDependencies": {
         "@biomejs/biome": "2.3.14",
-        "@canonical/biome-config": "^0.14.0",
-        "@canonical/typescript-config-react": "^0.14.0",
+        "@canonical/biome-config": "^0.15.0",
+        "@canonical/typescript-config-react": "^0.15.1",
         "@types/node": "^24.10.13",
         "@types/react": "^19.2.13",
         "@types/react-dom": "^19.2.3",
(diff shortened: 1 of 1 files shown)


packages/svelte/ds-app-launchpad/src/lib/components/Chip/styles.css
jmuzina: I noticed you haven't implemented the [inline variant](https://vanillaframework.io/docs/patterns/chip#inline-chips) - was this determined to be no longer necessary or just something that we don't need yet in the library?

packages/svelte/ds-app-launchpad/src/lib/components/Chip/styles.css
```diff
@@ -0,0 +1,87 @@
...
+    --lp-dimension-padding-block-context,
+    var(--lp-dimension-spacing-block-xxxs)
+  );
+  padding-inline: var(
+    --lp-dimension-padding-inline-context,
+    var(--lp-dimension-spacing-inline-s)
+  );
+  transition-duration: var(--lp-transition-duration-snap);
+  transition-property: background-color, border-color;
+  transition-timing-function: var(--lp-transition-timing-ease-in);
```
jmuzina: I'd recommend to remove these transition properties in a `prefers-reduced-motion` query. The animations aren't especially noticeable right now because of the very fast transition duration you currently have defined - but if this were increased this would be an accessibility problem for users who prefer reduced motion.

advl: Could we document this as [a code-standard](https://github.com/canonical/code-standards) to follow-up ?

packages/svelte/ds-app-launchpad/src/lib/components/Chip/types.ts
```diff
@@ -0,0 +1,39 @@
...
+   */
+  lead?: string;
+
+  /** The icon of the Chip
+   * @default no icon
+   */
+  icon?: Snippet | undefined | null;
+
+  /** The badge to be displayed on the Chip */
+  badge?: Snippet | undefined | null;
```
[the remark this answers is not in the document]
goulinkh: Regarding your first question, `Snippet` is a very powerful pattern that is unique to Svelte 5. It essentially replaces the older slot mechanism and gives us much better inversion of control. By accepting a snippet for the icon and badge, we avoid having to bundle or map the entire `@canonical/ds-assets` icon set inside the `Chip` component itself, which keeps our component types and imports lean.

This approach is highly beneficial for tree-shaking, as the consumer only imports the exact icon component they need at the implementation level and passes it in.

It also provides the flexibility to render a custom SVG or a different element if an edge case comes up, without us needing to update the component's internal logic. For instance, [`@canonical/svelte-icons`](https://github.com/canonical/svelte-icons?tab=readme-ov-file#custom-icons) allows for defining custom temporary icons in your project without having to wait for them to be upstreamed and published to `ds-assets`. The snippet approach makes it very easy to drop those custom icons right into this component's API.

PR #424, opened 2026-02-18
fix(tty): acknowledge case when Ink can't read from STDIN [AC-4269]

Stefan3002:
## Done
**CURRENT STATE OF PR:**
- Acknowledge the problem in the docs

**PREVIOUS ACTIONS DONE IN PR:**

- Improved the way we handle the ```isTTY``` check
- On top of checking if it is a tty, I also tried setting the stdin mode to raw, in order to not get cooked input, but raw one
- If it throws an error, we can't read from tty

Fixes [AC-4269](https://warthogs.atlassian.net/browse/AC-4269)

## QA

- Run ```summon component react ./path
- If you are in an env where you can't read from stdin but still in tty (actually do not kn ow how this happens, will have to keep investigating), you should see a handled error
- If you pass in the ```--yes``` flag, it should work as expected

### PR readiness check

## Screenshots

image

[AC-4269]: https://warthogs.atlassian.net/browse/AC-4269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ The server starts at http://localhost:6006. Each component package has its own S
 
 - **React 19** or later for component packages (Svelte 5 for Svelte packages)
 - **Bun 1.3.9** or later for package management and script execution
+  - Installing Bun via the snap package is not recommended as it may cause permission issues. This is a known [issue](https://github.com/shakeelansari63/snap-packages/issues/79). Use the official installation script instead.
 - **Node.js 22.12 or 24** because Storybook and Lerna depend on Node internals that Bun does not yet fully support. Node 22.12 specifically is required as earlier 22.x versions have module resolution issues.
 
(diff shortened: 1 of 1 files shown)


Stefan3002:
@advl Hmm, let me investigate a bit to see if that is actually the problem

Stefan3002:
@advl yes, the snap-installed bun was the problem, apparently we can't use rawMode which is required by Ink with the snap-installed bun. Now, with this fix in this PR, it will successfully detect this and prompt the user to use the --yes flag. We can specify this somewhere in the docs. "You can use both, but know that in the snap-installed bun, you MUST use --yes flag".

PR #433, opened 2026-02-24
feat(svelte-ds-app-launchpad): upstream Button, ButtonPrimitive, and Spinner components

goulinkh:
Fixes [LP-3723](https://warthogs.atlassian.net/browse/LP-3723)

### PR readiness check

## Screenshots

image

[LP-3723]: https://warthogs.atlassian.net/browse/LP-3723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/svelte/ds-app-launchpad/src/lib/components/Switch/Switch.svelte
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/Switch/Switch.svelte
@@ -44,6 +44,7 @@
   {disabled}
   bind:checked={getChecked, setChecked}
   aria-checked={ariaChecked}
+  aria-readonly={disabled}
   {...rest}
 />
 
--- a/packages/svelte/ds-app-launchpad/vite.config.ts
+++ b/packages/svelte/ds-app-launchpad/vite.config.ts
@@ -24,7 +24,7 @@ export default defineConfig({
 					},
 					include: ["src/**/*.svelte.test.{js,ts}"],
 					exclude: ["src/lib/server/**"],
-					setupFiles: ["vitest-browser-svelte"],
+					setupFiles: ["./vitest-setup-client.ts"],
 				},
 			},
 			{
--- a/packages/svelte/ds-app-launchpad/vitest-setup-client.ts
+++ b/packages/svelte/ds-app-launchpad/vitest-setup-client.ts
@@ -1,0 +1,9 @@
+/// <reference types="@vitest/browser/matchers" />
+
+import "@canonical/launchpad-design-tokens/dist/css/dimension/responsive.css";
+import "@canonical/launchpad-design-tokens/dist/css/typography/responsive.css";
+import "@canonical/launchpad-design-tokens/dist/css/opacity/opacity.css";
+import "@canonical/launchpad-design-tokens/dist/css/transition/preferred.css";
+import "@canonical/launchpad-design-tokens/dist/css/color/light.css";
+import "./src/lib/index.css";
+


packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -0,0 +1,135 @@
...
+	{#snippet template(args)}
+		<Button {...args} onclick={fn()}>
+			{#snippet iconLeft()}
+				<ArchiveIcon />
+			{/snippet}
+			With left icon
+		</Button>
+		&nbsp;
+		<Button {...args} onclick={fn()}>
+			With right icon
```
jmuzina: A little nitpick, these buttons are not vertically aligned

Image

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -0,0 +1,135 @@
...
+			{#snippet iconRight()}
+				<ArchiveIcon />
+			{/snippet}
+		</Button>
+		&nbsp;
```
jmuzina: Just curious, is there a reason you went for spacing the buttons with `nbsp` instead of applying `margin-inline`?  I think creating a set of buttons next to each other might be a pretty common use-case and needing to manually create nbsp's to space them out might become cumbersome

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.svelte
```diff
@@ -0,0 +1,63 @@
...
+<ButtonPrimitive
+	bind:ref
+	class={[
+		componentCssClassName,
+		className,
+		severity,
+		density,
+		{ loading, "explicit-disabled": disabled },
+	]}
+	disabled={isDisabled}
```
jmuzina: Heads up, using `disabled` like this has negative accessibility implications in many situations. It completely removes the button from the tabbing order so it becomes effectively invisible to assistive technologies.

In https://github.com/canonical/vanilla-framework/issues/4299 it was decided that this problem was too high-effort to solve in Vanilla, but we have a chance here to build this the right way going forward. Maybe we could bind `disabled` to `aria-disabled` and `explicit-disabled` (to disable it for assistive tech and visually), and disable the `onClick` event for disabled buttons, what do you think?

packages/svelte/ds-app-launchpad/src/lib/components/common/ButtonPrimitive/ButtonPrimitive.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
```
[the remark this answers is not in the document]
goulinkh: AFAIK Pragma exports convention is the way to achieve this, when explicitly not setting a component's module (`index.ts`) as exported from the main components' index file it makes it impossible to import from default.
Where the consumer can still import it manually from the complete path if the package.json exports allows such rule.

In the case of svelte pragma the setup only allows import from root path, so if a component is not exported, it is impossible to import by consumers.

I don't think Svelte has any notion of private components though, if that's what you are wondering about.

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.svelte
jmuzina: Yes, resolved in thread https://github.com/canonical/pragma/pull/435#discussion_r2850293391

steciuk: #437 - Created an issue to investigate this :)

PR #606, opened 2026-04-08
feat(svelte-ds-app-launchpad): Add UserAvatar component

steciuk:
## Done
- Add `UserAvatar` component that displays an avatar image when available, falls back to user initials if `userName` is provided, and a generic icon placeholder otherwise.
- No-JS fallback: when the image fails to load without JS, initials are shown via CSS `::after` on the `` if `userName` is provided and a `?` otherwise.

## QA
- verify `Components/UserAvatar` in Storybook
  - check image renders when `userAvatarUrl` is provided
  - check initials fallback when no image URL is given or image fails to load
  - check icon placeholder renders when neither image nor name is provided

### PR readiness check

## Screenshots

image

image

### DIFF
--- a/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
@@ -12,6 +12,7 @@
     userName: userNameProp,
     userAvatarUrl,
     size = "medium",
+    alt,
     ...rest
   }: UserAvatarProps = $props();
 
@@ -31,7 +32,7 @@
   <img
     class={[componentCssClassName, size, className]}
     src={userAvatarUrl}
-    alt={userName ? `${userName}'s avatar` : "User avatar"}
+    {alt}
     title={userName || undefined}
     data-initials={userInitials}
     onerror={() => (imageError = true)}
@@ -42,7 +43,7 @@
     {#if userName}
       <abbr title={userName}>{userInitials}</abbr>
     {:else}
-      <UserIcon aria-label="User avatar icon" />
+      <UserIcon />
     {/if}
   </div>
 {/if}
--- a/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
@@ -1,0 +1,61 @@
+<script lang="ts">
+  /* @canonical/generator-ds 0.9.0-experimental.22 */
+  import { UserIcon } from "@canonical/svelte-icons";
+  import type { UserAvatarProps } from "./types.js";
+  import "./styles.css";
+  import { getInitials } from "./utils/index.js";
+
+  const componentCssClassName = "ds user-avatar";
+
+  const {
+    class: className,
+    userName: userNameProp,
+    userAvatarUrl,
+    size = "medium",
+    ...rest
+  }: UserAvatarProps = $props();
+
+  const userName = $derived(userNameProp?.trim() || null);
+  const userInitials = $derived(userName ? getInitials(userName) : null);
+
+  let imageError = $state(false);
+
+  // Reset image error state when the avatar URL changes
+  $effect.pre(() => {
+    userAvatarUrl;
+    imageError = false;
+  });
+</script>
+
+{#if userAvatarUrl && !imageError}
+  <img
+    class={[componentCssClassName, size, className]}
+    src={userAvatarUrl}
+    alt={userName ? `${userName}'s avatar` : "User avatar"}
+    title={userName || undefined}
+    data-initials={userInitials}
+    onerror={() => (imageError = true)}
+    {...rest}
+  />
+{:else}
+  <div class={[componentCssClassName, "no-image", size, className]} {...rest}>
+    {#if userName}
+      <abbr title={userName}>{userInitials}</abbr>
+    {:else}
+      <UserIcon aria-label="User avatar icon" />
+    {/if}
+  </div>
+{/if}
+
+<!-- @component
+`UserAvatar` A component that displays a user's avatar.
+
+The avatar will display the user's image if available and able to be loaded, otherwise it will display the first two initials. If neither is available, it will display a default icon placeholder.
+
+In case JavaScript is disabled, and the image at `userAvatarUrl` fails to load, the component will provide a fallback: displaying the user's initials when `userName` is provided, or `?` when it is not.
+
+## Example Usage
+```svelte
+<UserAvatar userName="John Doe" userAvatarUrl="https://example.com/avatar.png" />
+```
+-->
--- a/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
@@ -8,7 +8,7 @@
   const componentCssClassName = "ds user-avatar";
 
   const {
-    class: className,
+    class: classProp,
     userName: userNameProp,
     userAvatarUrl,
     size = "medium",
@@ -16,6 +16,7 @@
     ...rest
   }: UserAvatarProps = $props();
 
+  const className = $derived([componentCssClassName, size, classProp]);
   const userName = $derived(userNameProp?.trim() || null);
   const userInitials = $derived(userName ? getInitials(userName) : null);
 
@@ -30,7 +31,7 @@
 
 {#if userAvatarUrl && !imageError}
   <img
-    class={[componentCssClassName, size, className]}
+    class={className}
     src={userAvatarUrl}
     {alt}
     title={userName || undefined}
@@ -38,13 +39,13 @@
     onerror={() => (imageError = true)}
     {...rest}
   />
+{:else if userName}
+  <abbr class={className} title={userName} {...rest}>
+    {userInitials}
+  </abbr>
 {:else}
-  <div class={[componentCssClassName, "no-image", size, className]} {...rest}>
-    {#if userName}
-      <abbr title={userName}>{userInitials}</abbr>
-    {:else}
-      <UserIcon />
-    {/if}
+  <div class={className} {...rest}>
+    <UserIcon />
   </div>
 {/if}
 
--- a/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
@@ -13,4 +13,6 @@ export interface UserAvatarProps
   extends Omit<HTMLAttributes<HTMLElement>, "children">,
     UserOptions {
   size?: "small" | "medium" | "large";
+  /** The alt text for the avatar image, if available */
+  alt?: string;
 }
--- a/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
@@ -1,0 +1,16 @@
+/* @canonical/generator-ds 0.9.0-experimental.22 */
+
+import type { HTMLAttributes } from "svelte/elements";
+
+export type UserOptions = {
+  /** The user's name */
+  userName?: string;
+  /** The URL of the user's avatar image */
+  userAvatarUrl?: string;
+};
+
+export interface UserAvatarProps
+  extends Omit<HTMLAttributes<HTMLElement>, "children">,
+    UserOptions {
+  size?: "small" | "medium" | "large";
+}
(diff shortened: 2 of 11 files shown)


packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
```diff
@@ -0,0 +1,61 @@
...
+    userAvatarUrl;
+    imageError = false;
+  });
+</script>
+
+{#if userAvatarUrl && !imageError}
+  <img
+    class={[componentCssClassName, size, className]}
+    src={userAvatarUrl}
+    alt={userName ? `${userName}'s avatar` : "User avatar"}
```
jmuzina: Avatars should generally be considered decorative (no alt text) unless they are interactive or add additional information not presented elsewhere. See [comment](https://warthogs.atlassian.net/browse/WD-32072?focusedCommentId=968683) for an explainer. 

Rather than setting the `alt` here based on username, maybe we should let the caller decide what the alt should be and just spread it onto here?

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type UserOptions = {
+  /** The user's name */
+  userName?: string;
+  /** The URL of the user's avatar image */
+  userAvatarUrl?: string;
+};
+
+export interface UserAvatarProps
+  extends Omit<HTMLAttributes<HTMLElement>, "children">,
```
jmuzina, line 13: Would we like callers to be able to pass image props here? In that case, it would be nice to extend `HTMLAttributes<HTMLImageElement>`, but this is complicated because the root element of this component is a `div` in some cases. 

Maybe something like this would work (haven't tried this):

```typescript
interface BaseAvatarProps {
  size?: "small" | "medium" | "large";
  class?: string;
}

interface AvatarWithImage extends BaseAvatarProps, Omit<HTMLImgAttributes, "size" | "src"> {
  userAvatarUrl: string;
  userName?: string;
}

interface AvatarWithoutImage extends BaseAvatarProps, Omit<HTMLAttributes<HTMLDivElement>, "size"> {
  userAvatarUrl?: never; // This is the magic "discriminated" part
  userName?: string;
}

export type UserAvatarProps = AvatarWithImage | AvatarWithoutImage;
```

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
```diff
@@ -0,0 +1,16 @@
+/* @canonical/generator-ds 0.9.0-experimental.22 */
+
+import type { HTMLAttributes } from "svelte/elements";
+
+export type UserOptions = {
+  /** The user's name */
+  userName?: string;
+  /** The URL of the user's avatar image */
+  userAvatarUrl?: string;
```
jmuzina, line 9: Why can't we just use the native `src` for this?

steciuk: We decided to use `userAvatarUrl` to keep the `UserOptions` type reusable and decoupled from the specifics of the `img` element (with all its properties relating to a user `"user..."`). You can see an example of its use in the [Timeline.Event component](https://github.com/canonical/pragma/pull/607/changes#diff-fe9cb39605e6b4240f1c40e10607171be626cf55b631d6af429441dd7af50065R13).

We could stick to `src` here and e.g., map `userAvatarUrl` to it within the `Timeline`, but that would create an API inconsistency.

Please let me know which approach you prefer :)

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
steciuk: This approach would be problematic here, because the root element type not only depends on the presence of `userAvatarUrl` (where this would be ok to use), but also on the possibility of an image loading error when the `img` is replaced with the `div`.

What we could do is to manually pick a subset of the most commonly used `img` attributes, destructure them from `rest` in `$props()` definition and only forward to the `img` element. If you think that's the move, what attributes should we choose? `alt`, `decoding`, `fetchpriority`, `loading`? Do you have any other ideas on how to handle that?

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
steciuk: Thanks for pointing that out! I removed the default image's `alt` and icon's `aria-label`.

Additionally, I flattened the component's structure (`abbr` now is rendered as root) to allow for `title`/`aria-label` overrides if necessary.

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
```diff
@@ -0,0 +1,63 @@
...
+  const userName = $derived(userNameProp?.trim() || null);
+  const userInitials = $derived(userName ? getInitials(userName) : null);
+
+  let imageError = $state(false);
+
+  // Reset image error state when the avatar URL changes
+  $effect.pre(() => {
+    userAvatarUrl;
+    imageError = false;
+  });
```
jmuzina, line 29: When QA'ing this time, the ubuntu assets manager was experiencing some nasty performance issues. On the bright side, it revealed to me a question / possible improvement:

While the image was loading (it took quite a long time!) the success state of the avatar loading is unknown - so an empty area is rendered.

Image

It only switches to the real content after the loading completes or fails. 

I wonder how (or if) we should handle this? Should we display a loading spinner while the network request is ongoing, and possibly make that configurable? Or a "skeleton"?

Feel free to address this later, I know it's an optimization that probably hasn't been discussed yet. But I'm curious for your thoughts

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
jmuzina: > We could stick to src here and e.g., map userAvatarUrl to it within the Timeline, but that would create an API inconsistency.

Is that really an "inconsistency", or is it a more honest modeling of how the information's meaning is different from component to component? At the level of the Timeline, it makes sense to scope these props to "user" because it's a different information space than the component in question. In the user avatar, though, it feels a bit redundant to repeat `user...` in these props (the component name already tells us this is related to a user).

We should be sure this is a sound approach though, I don't want to make the API or implementation cumbersome due to this. Do you have any opinions here @nathanclairmonte @jademathre-canonical @goulinkh @alvaromateo ?

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
jmuzina: My gut is to restructure things somewhat so that we actually have different components based on what the root element is to simplify the types, so we don't need to own the maintenance of picking "commonly used" attributes.

- `UserAvatar`  that controls the logic for determining which specific component to use based on loading state, etc
  - `UserAvatarImage` that is always an image
  - `UserAvatarPlaceholder` or other name that is the div used in error/loading

Then at the `UserAvatar` level you could expose two sets of props, `imageProps`, and `placeholderProps`, and spread those onto the respective elements when they are rendered. Do you think this is any nicer?

steciuk: My main concern is that we'd be trading our
> maintenance of picking "commonly used" attributes

for a more complex consumer API - one that doesn't occur in any of our other components so far.

For example, if a consumer wants to apply a `class` and an `aria-label`, they would be forced to specify those props twice:

```svelte
<UserAvatar
  userName="John"
  imageAttributes={{
    src: "https://assets.ubuntu.com/v1/fca94c45-snap+icon.png",
    alt: "User Avatar",
    
    class: "my-class",
    "aria-label": "User Avatar",
  }}
  placeholderAttributes={{
    class: "my-class",
    "aria-label": "User Avatar",
  }}
/>
```

If we were to eventually separate things further into `imageProps`, `initialsProps`, and `iconProps`, they might even have to define them three times! And to be frank, I am not a fan of this direction.

It also forces us into other tricky decisions. For example - regarding attachments. Should each root element have its own way of passing an attachment to keep things consistent but force the user into questionable ergonomics of relying on `createAttachmentKey`?

```svelte
<UserAvatar
  imageAttributes={{
    [createAttachmentKey()]: (el) => {
      // ...
    },
  }}
  placeholderAttributes={{
    [createAttachmentKey()]: (el) => {
      // ...
    },
  }}
/>
```

Or, do we expose one shared attachment at the root level, breaking the pattern of grouping things by element?

```svelte
<UserAvatar
  imageAttributes={{
    //...
  }}
  placeholderAttributes={{
    //...
  }}
  {@attach (el) => {
    // ...
  }}
/>
```

Personally, I value keeping the consumer API simple, ergonomic, and consistent with the rest of our components, even if it means we have to take on a bit of extra maintenance under the hood.

If our main goal is to avoid making a decision on "which img-specific attributes to choose", I would prefer to just manually expose all that the standard currently defines:

- `alt`
- `crossorigin`
- `decoding`
- `fetchpriority`
- `height`
- `ismap`
- `loading`
- `referrerpolicy`
- `sizes`
- `src`
- `srcset`
- `usemap`
- `width`

There really aren't _that many_ of them, and I wouldn't expect the list of standard `img` attributes to change very often anyway.

Let me know what you think.

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
steciuk: I like this framing, and I'm happy with either option. Let's see what other folks think!

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/UserAvatar.svelte
steciuk: That's a valid point and an area for further improvement. I'll speak with @EnzoYD about this. Thanks!

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
steciuk: I came up with another, somewhat "hybrid" option that I think is the best middle ground between the two approaches. The idea is to keep **all** attributes that are shared between the root elements as flat component props and allow for `imageAttributes` that'd contain **only** the ones that are `img`-specific.

```ts
import type { HTMLAttributes, HTMLImgAttributes } from "svelte/elements";

type ImageOnlyAttributes = Omit<
  HTMLImgAttributes,
  | keyof HTMLAttributes<HTMLElement>
  | "bind:naturalWidth"
  | "bind:naturalHeight"
  | "children"
>;

export interface UserAvatarProps extends Omit<
  HTMLAttributes<HTMLElement>,
  "children"
> {
  userName?: string;
  size?: "small" | "medium" | "large";
  imageAttributes?: ImageOnlyAttributes;
}
```

And then:

```svelte
<script lang="ts">
  const {
    class: classProp,
    userName: userNameProp,
    size = "medium",
    imageAttributes,
    ...rest
  }: UserAvatarProps = $props();
  
  // ...
</script>

{#if imageAttributes?.src && !imageError}
  <img
    class={className}
    title={userName || undefined}
    data-initials={userInitials}
    onerror={() => (imageError = true)}
    {...imageAttributes}
    {...rest}
  />
{:else if userName}
  <abbr class={className} title={userName} {...rest}>
    {userInitials}
  </abbr>
{:else}
  <div class={className} {...rest}>
    <UserIcon />
  </div>
{/if}
```

This raises a different question, related to your [other comment](https://github.com/canonical/pragma/pull/606#discussion_r3083303352). If we move forward with this approach, do we:
1. Keep `src` within `imageAttributes`. This splits user-related data across different levels (the `userName` root property vs. `imageAttributes.src`).
2. Remove `src` from `imageAttributes` and move it to the root as `userImageUrl`, placing it directly alongside `userName`.
3. Remove `src` from `imageAttributes` and create a root-level `user` object containing both `name` and `url`.

I personally lean toward _Option 2_, as it surfaces the most important properties for better visibility.

Please let me know what your thoughts are.

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
jmuzina: Moved to its own ticket to unblock: https://github.com/canonical/pragma/issues/627

packages/svelte/ds-app-launchpad/src/lib/components/UserAvatar/types.ts
jmuzina: Moved to separate thread to unblock https://github.com/canonical/pragma/issues/628

jmuzina (approved):

PR #607, opened 2026-04-09
feat(svelte-ds-app-launchpad): Add Timeline

steciuk:
## Done

- Add a new `Timeline` component for rendering chronological activity entries.
- Add `Timeline.Event` with support for avatar/icon markers, marker sizing, title-row content, and optional event body content.
- Add `Timeline.Event.TitleRow` to format leading text, event summary text, and date metadata.
- Add `Timeline.HiddenEvents` and `Timeline.HiddenEvents.Link` to represent collapsed event groups with one or more reveal actions.

## QA

- verify `Components/Timeline` in Storybook
- verify `Components/Timeline/Event` in Storybook
- verify `Components/Timeline/Event/TitleRow` in Storybook
- verify `Components/Timeline/HiddenEvents` in Storybook
- verify `Components/Timeline/HiddenEvents/Link` in Storybook

### PR readiness check

## Screenshots

image

image


packages/svelte/ds-app-launchpad/src/lib/components/Timeline/common/Event/styles.css
```diff
@@ -0,0 +1,113 @@
...
+      var(--dimension-gap-row-timeline) -
+      var(--marker-alignment-adjustment)
+    );
+    margin-block-start: var(--marker-alignment-adjustment);
+  }
+
+  &:last-child::before {
+    /* End the line at the marker */
+    height: var(--marker-alignment-adjustment);
+  }
```
jmuzina: If there is only one event, it is both the first and last child. This seems to work fine in your current implementation because of source order (last-child is after first-child so  the last-child height is overridding the first-child height).

However, this is *slightly* brittle to modifications if somebody shifts the order of selectors around. 

I'd suggest to maybe make your first-child height contingent on not also being the last child. Something like this:

```scss
&:first-child {
  // Height for first events that are not the only event
  &:not(:last-child)::before {
    height: // calc....
  }
  // Shared by all first events
  &::before {
    margin-block-start: var(--marker-alignment-adjustment);
  }
}

&:last-child::before {
  height: // var......
}

File: packages/svelte/ds-app-launchpad/src/lib/components/Timeline/common/Event/Event.svelte [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+    marker === undefined
+      ? "marker-empty"
+      : markerSize === "small"
+        ? "marker-small"
+        : "marker-large",
+    { "with-title-row": titleRow },
+  ]}
+  {...rest}
+>
+  <div class="marker" aria-hidden="true">
```
jmuzina: Are we sure the marker will never contain interactive elements? 

If not, we might want to consider allowing specific subcomponents inside it to handle setting their aria appropriately

packages/svelte/ds-app-launchpad/src/lib/components/Timeline/common/Event/Event.svelte
```diff
@@ -0,0 +1,54 @@
...
+
+<li
+  class={[
+    componentCssClassName,
+    className,
+    marker === undefined
+      ? "marker-empty"
+      : markerSize === "small"
+        ? "marker-small"
+        : "marker-large",
```
jmuzina: If `marker` is null we will end up not using `marker-empty` because you're only checking for `undefined`. It's not easily possible to get this to happen because the prop types don't accept null, but it could be a source of confusion if that changes. Could we just check for `!marker`?

packages/svelte/ds-app-launchpad/src/lib/components/Timeline/common/Event/styles.css
steciuk: Good point! I actually went a step further and conditioned all the connector-related CSS on `&:not(:only-child)` as if an event were the only child we don't need to render the connector at all!

packages/svelte/ds-app-launchpad/src/lib/components/Timeline/common/Event/Event.svelte
steciuk: Very good point!
I removed `aria-hidden`. By applying your recommended changes to [`UserAvatar`](https://github.com/canonical/pragma/pull/606#discussion_r3083280045) we’ve reduced the accessibility "noise," so hiding the marker by default is no longer necessary.

The change ensures that any ARIA attributes applied to custom content within the marker Snippet are no longer silenced.

Thank you!

packages/svelte/ds-app-launchpad/src/lib/components/Timeline/common/Event/common/TitleRow/TitleRow.svelte
```diff
@@ -0,0 +1,32 @@
...
+>
+  <span class="content">
+    {#if leadingText}
+      <span class="leading-text">{leadingText}</span>
+    {/if}
+    <span class="description">
+      {@render children()}
+    </span>
+  </span>
+  <span class="date">{@render date()}</span>
```
jmuzina, line 31: I'm curious, why make `date` a snippet that can be fully replaced instead of accepting a `Date` or maybe also a `string` and then taking care of how to render that ourselves instead of expecting the user to? We could abstract the use of `RelativeDateTime` away... but maybe that would be too restrictive

jmuzina (approved):
Approving with one question I'm curious about

packages/svelte/ds-app-launchpad/src/lib/components/Timeline/common/Event/common/TitleRow/TitleRow.svelte
steciuk: > but maybe that would be too restrictive

That was our intuition. We didn't want to make assumptions about the date formatting and wanted to avoid forcing date-related configuration props onto the `TitleRow` component. Expecting the consumer to decide whether they want to go with the [`DateTime`](https://main--698d9cc0a0decf9afb7a90cb.chromatic.com/?path=/docs/components-datetime--docs), [`RelativeDateTime`](https://main--698d9cc0a0decf9afb7a90cb.chromatic.com/?path=/docs/components-relativedatetime--docs) or any other means of reflecting the time seemed like a safer move.

PR #619, opened 2026-04-21
feat(svelte-ds-app-launchpad): Add Table

steciuk:
## Done

- Add `Table` component supporting sortable columns via a `TH` sub-component with `SortButton`
- Temporarily add `SortIcon` icon used by the sort indicator (to be removed when the icon is approved for general use and upstreamed to `@canonical/ds-assets`).
- Add `DistributiveOmit` type utility.
- Add `SeededRandom` helper for deterministic test fixtures/stories data generation.

## QA
- verify `Components/Table` in Storybook
  - Check default rendering with multiple columns and rows
  - Make sure the table is styled properly when the native table-related elements are used
  - Interact with sortable column headers and confirm that sort direction is correctly indicated both visually and to assistive technologies 

### PR readiness check

## Screenshots

image

### DIFF
--- a/packages/svelte/ds-app-launchpad/src/lib/components/Table/Table.svelte
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/Table/Table.svelte
@@ -1,0 +1,52 @@
+<!-- @canonical/generator-ds 0.10.0-experimental.5 -->
+
+<script lang="ts">
+  import type { TableProps } from "./types.js";
+  import "./styles.css";
+
+  const componentCssClassName = "ds table";
+
+  let { class: className, children, ...rest }: TableProps = $props();
+</script>
+
+<table class={[componentCssClassName, className]} {...rest}>
+  {@render children?.()}
+</table>
+
+<!-- @component
+`Table` is a styled table element.
+
+## Example Usage
+```svelte
+<Table>
+  <caption>Sample Data</caption>
+  <thead>
+    <tr>
+      <Table.TH aria-sort="ascending" scope="col">
+        Sortable Column
+        {#snippet action()}
+          <Table.TH.SortButton onclick={handleSortClick} aria-label="Sort descending by Sortable Column" />
+        {/snippet}
+      </Table.TH>
+      <th scope="col">Non-Sortable Column</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>Data 1</td>
+      <td>Data 2</td>
+    </tr>
+    <tr>
+      <td>Data 3</td>
+      <td>Data 4</td>
+    </tr>
+  </tbody>
+  <tfoot>
+    <tr>
+      <td>Footer 1</td>
+      <td>Footer 2</td>
+    </tr>
+  </tfoot>
+</Table>
+```
+-->
--- a/packages/svelte/ds-app-launchpad/src/lib/components/Table/styles.css
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/Table/styles.css
@@ -1,0 +1,46 @@
+.ds.table {
+  border-collapse: collapse;
+
+  --border-table-row: var(--lp-dimension-stroke-thickness-default) solid
+    var(--lp-color-border-low-contrast);
+  --dimension-padding-block-table-cell: var(--lp-dimension-spacing-block-xs);
+  --dimension-padding-inline-table-cell: var(--lp-dimension-spacing-inline-xs);
+
+  th,
+  td {
+    padding-inline: var(--dimension-padding-inline-table-cell);
+    padding-block: var(--dimension-padding-block-table-cell);
+  }
+
+  th {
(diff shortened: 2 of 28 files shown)


steciuk:
Sorry, @jmuzina, I must have assigned you as a reviewer on autopilot before :sweat_smile:

Unassigning now, so you can pick it up if and when you feel like it :)

packages/svelte/ds-app-launchpad/src/lib/components/Table/styles.css
```diff
@@ -0,0 +1,46 @@
...
+  --border-table-row: var(--lp-dimension-stroke-thickness-default) solid
+    var(--lp-color-border-low-contrast);
+  --dimension-padding-block-table-cell: var(--lp-dimension-spacing-block-xs);
+  --dimension-padding-inline-table-cell: var(--lp-dimension-spacing-inline-xs);
+
+  th,
+  td {
+    padding-inline: var(--dimension-padding-inline-table-cell);
+    padding-block: var(--dimension-padding-block-table-cell);
+  }
```
jmuzina, line 13: Should we be targeting bare "tag" selectors inside class-qualified `ds.table`? I guess the only risk would be if somebody decides to nest a non-DS table inside this table - it could break the layout. Alternative would be we need to target these by class selectors and build Svelte components for *everything* that's currently selected by tag selectors, then style those within their component-specific stylesheets.

packages/svelte/ds-app-launchpad/src/lib/type-utils/DistributiveOmit.ts
jmuzina, line 1: nice, i just [read about this](https://stackoverflow.com/a/57103940), maybe something to later add upstream

packages/svelte/ds-app-launchpad/src/lib/components/icons/index.ts
```diff
@@ -0,0 +1,2 @@
+// TODO: Remove icons when they are approved, upstreamed to `@canonical/ds-assets` included in `@canonical/svelte-icons`.
```
jmuzina: how do you feel about setting a convention that our TODOs always qualify who owns them (ex: `TODO(steciuk)`)?

jmuzina (approved):

packages/svelte/ds-app-launchpad/src/lib/components/icons/index.ts
steciuk: I like that idea! It adds a bit of extra pressure so they don't just sit around indefinitely (as TODOs tend to do) :)

packages/svelte/ds-app-launchpad/src/lib/components/Table/Table.svelte
```diff
@@ -0,0 +1,52 @@
...
+        Sortable Column
+        {#snippet action()}
+          <Table.TH.SortButton onclick={handleSortClick} aria-label="Sort descending by Sortable Column" />
+        {/snippet}
+      </Table.TH>
+      <th scope="col">Non-Sortable Column</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
```
jmuzina, line 35: I'm curious for your thought process on which subcomponents of Table merit their own Svelte components (you built a custom `Table.TH` and `Table.TH.SortButton`, but other components like `tr`, `thead`, `tbody`, `tfoot` remain native elements)

packages/svelte/ds-app-launchpad/src/lib/components/Table/styles.css
steciuk: The thinking was that this pattern makes sense for components represented by elements with strict requirements for their child types (e.g., table-related elements within a `<table>`, or `<option>` within a `<select>`). The alternative of componentizing every single element felt like it would just add extra boilerplate without much real benefit.

packages/svelte/ds-app-launchpad/src/lib/components/Table/Table.svelte
steciuk: We built `Table.TH` to encapsulate the extra sorting-related behavior, so it's there to use whenever sorting is required. That said, the `Table` still [fully supports native `th` elements](https://698d9cc0a0decf9afb7a90cb-vzjmdiggfb.chromatic.com/?path=/story/components-table--default) if you just need a static table or prefer to implement sorting indicators yourself.

I could definitely see us adding components like `Table.TR` or `Table.TD` down the line to encapsulate new features as they come up.

packages/svelte/ds-app-launchpad/src/lib/components/Table/styles.css
jmuzina: Yeah, makes sense - I agree - no need to componentize something unless we are adding custom behavior / markup that would otherwise be cumbersome to add by the caller

goulinkh (approved):

PR #620, opened 2026-04-21
feat(svelte-ds-app-launchpad): WD-35462 Upstream Tooltip

jademathre-canonical:
## Done

Upstreamed Tooltip component from Launchpad UI
Minor refactor in Tooltip.stories.svelte to remove the type coercions around the Snippets 
Note: there are couple preexisting TODO items in the styles.css

### PR readiness check


packages/svelte/ds-app-launchpad/src/lib/components/Tooltip/Tooltip.svelte
```diff
@@ -0,0 +1,163 @@
...
+    element.addEventListener("mouseleave", onMouseLeave);
+    element.addEventListener("focus", onFocus);
+    element.addEventListener("blur", onBlur);
+
+    return () => {
+      triggerRef = undefined;
+      element.removeEventListener("mouseenter", onMouseEnter);
+      element.removeEventListener("mouseleave", onMouseLeave);
+      element.removeEventListener("focus", onFocus);
+      element.removeEventListener("blur", onBlur);
```
steciuk: Nit: While we are at it, let's maybe change the listeners' teardown to a bit more modern `signal`-based? WDYT?


```suggestion

    const abortController = new AbortController();
    element.addEventListener("mouseenter", () => (isTriggerHovered = true), {
      signal: abortController.signal,
    });
    element.addEventListener(
      "mouseleave",
      (e) => {
        // Check if hover moved to the tooltip, to avoid any flicker before `mouseenter` on the tooltip is fired and handled
        if (isEventTargetInElement(e.relatedTarget, tooltipRef)) {
          isTooltipHovered = true;
        }
        isTriggerHovered = false;
      },
      { signal: abortController.signal },
    );
    element.addEventListener("focus", () => (isTriggerFocused = true), {
      signal: abortController.signal,
    });
    element.addEventListener("blur", () => (isTriggerFocused = false), {
      signal: abortController.signal,
    });

    return () => {
      triggerRef = undefined;
      abortController.abort();
```

packages/svelte/ds-app-launchpad/src/lib/components/Tooltip/styles.css
```diff
@@ -0,0 +1,118 @@
...
+  --dimension-width-tooltip: max-content;
+
+  position: absolute;
+  width: var(--dimension-width-tooltip);
+  z-index: var(--z-index-tooltip);
+  background-color: var(--color-background-tooltip);
+  font: var(--typography-tooltip);
+  color: var(--color-text-tooltip);
+  padding-inline: var(--dimension-padding-inline-tooltip);
+  padding-block: var(--dimension-padding-block-tooltip);
```
steciuk: Initially, when we were building these components, we tended to introduce component-specific token aliases for almost every CSS property, even when they were just passing through an existing semantic token. Over time, we realized that this added a fair amount of boilerplate without giving much in return, especially when the alias was only used in one place. As part of the upstreaming work, we’re cleaning that up and removing the extra layer where it doesn’t really help, so the CSS can rely on the semantic tokens directly.


```suggestion
    --color-background-tooltip: var(--lp-color-background-alt-reversed);
    --dimension-offset-tooltip-arrow: var(--lp-dimension-spacing-inline-m);
    --dimension-height-tooltip-arrow: var(--lp-dimension-size-xxxs);

    position: absolute;
    width: max-content;
    /* TODO(z-index-tokens): Should we tokenize z-indexes? What would they be used for except the tooltip? */
    z-index: 10;
    background-color: var(--color-background-tooltip);
    font: var(--lp-typography-paragraph-s);
    color: var(--lp-color-text-reversed);
    padding-inline: var(--lp-dimension-spacing-inline-m);
    padding-block: var(--lp-dimension-spacing-block-xs);
```

Keeping `--dimension-offset-tooltip-arrow`, `--color-background-tooltip`, and `--dimension-height-tooltip-arrow` imo makes sense as they are used in more than one place.

packages/svelte/ds-app-launchpad/src/lib/components/Tooltip/Tooltip.stories.svelte
```diff
@@ -0,0 +1,87 @@
...
+        control: false,
+      },
+    },
+  });
+</script>
+
+<script lang="ts">
+  function scrollToCenter(el: HTMLElement) {
+    el.scrollTo({ top: el.clientHeight / 2, left: el.clientWidth / 2 });
+  }
```
steciuk: 1. Let's please use the newer [Svelte Attachments](https://svelte.dev/docs/svelte/svelte-attachments).
2. We could add `behavior: "instant"` so scrolling isn't visible when the story loads.


```suggestion
  const scrollToCenter: Attachment = (el) =>
    el.scrollTo({
      top: el.clientHeight / 2,
      left: el.clientWidth / 2,
      behavior: "instant",
    });
```

packages/svelte/ds-app-launchpad/src/lib/components/Tooltip/Tooltip.stories.svelte
```diff
@@ -0,0 +1,87 @@
...
+    position: "block-end span-inline-end",
+    autoAdjust: true,
+    delay: 0,
+  }}
+>
+  {#snippet template({ trigger: _, children: __, ...args })}
+    <div
+      class="placeholder-box"
+      style="height: 250px; overflow: auto;"
+      use:scrollToCenter
```
steciuk: Pt. 2 of the change suggested [above](https://github.com/canonical/pragma/pull/620/changes#r3128927519).


```suggestion
      {@attach scrollToCenter}
```

packages/svelte/ds-app-launchpad/src/lib/components/Tooltip/types.ts
```diff
@@ -0,0 +1,47 @@
+/* @canonical/generator-ds 0.17.1 */
+
+import type { Snippet } from "svelte";
+import type { Attachment } from "svelte/attachments";
+import type { SvelteHTMLElements } from "svelte/elements";
+import type { PositionArea } from "../../useFunctions/usePositionArea.svelte.js";
+
+type BaseProps = SvelteHTMLElements["div"];
```
steciuk: As Copilot suggested:


```suggestion
type BaseProps = Omit<SvelteHTMLElements["div"], "children">;
```

steciuk (changes requested):
Thank you @jademathre-canonical!

A couple minor nitpicks and ideas for improvement, but overall it's looking great.

jademathre-canonical:
@steciuk Thanks for the suggestions! I've updated it accordingly.

steciuk (approved):
Thank you!

PR #625, opened 2026-04-23
feat(svelte-ds-app-launchpad): Add SearchBox

steciuk:
## Done

- Add a new `SearchBox` component.
- drive-by: fix missing semicolon in `Popover` trigger prop styles.

## QA

- verify `Components/SearchBox` in Storybook

### PR readiness check

## Screenshots

image


packages/svelte/ds-app-launchpad/src/lib/components/SearchBox/SearchBox.svelte
```diff
@@ -0,0 +1,62 @@
...
+    aria-label={ariaLabel}
+    {disabled}
+    class={{ "no-invalid-styles": !invalidStyled }}
+    {...rest}
+  />
+  <ButtonPrimitive
+    type="submit"
+    aria-label={ariaLabel}
+    {disabled}
+    onclick={onSearchButtonClick}
```
MichaelKapustey: This is very limiting. What if the user wants onPress, onMouseLeave?

I would name the parent event "onSearch" to avoid the feel that you're controlling internal implementation. 

Separately if we want to let user know about the properties of inner components and control them I recommend taking a look at `slotProps` (and maybe custom slots) concept from MUI: https://mui.com/x/common-concepts/custom-components/

packages/svelte/ds-app-launchpad/src/lib/components/SearchBox/types.ts
```diff
@@ -0,0 +1,25 @@
...
+   * Click event handler for the search button.
+   */
+  onSearchButtonClick?: MouseEventHandler<HTMLButtonElement>;
+  /**
+   * Whether to apply the invalid styles to the input when it fails native validation (`:user-invalid`) or `aria-invalid="true"` is set.
+   *
+   * @default false
+   * This is opt-in because search landmarks often have submission requirements, such as `required` and `minlength`, where you may want to block submission without showing a visual invalid state.
+   */
+  invalidStyled?: boolean;
```
MichaelKapustey: weird property name. I struggle to understand it. `styled` prefix sends my mind into a direction of `styled-components`. Should it be simply `invalid`? In analogy to `readonly` `disabled` and so on.

packages/svelte/ds-app-launchpad/src/lib/components/SearchBox/styles.css
```diff
@@ -0,0 +1,47 @@
...
+      --color-outline-input-invalid: var(--color-outline-input);
+    }
+  }
+
+  > button {
+    justify-self: end;
+    display: grid;
+    place-items: center;
+    padding: 0;
+    --color-background-button-hover: transparent;
```
MichaelKapustey: years ago I ran into issue where `transparent` color wasn't working correctly on Safari browsers. 
We have couple of options here
- investigate the topic.
- Ask a mac-owner to check this specific case
- use rgba (or introduce css transparent variable that uses rgba) explicitly

packages/svelte/ds-app-launchpad/src/lib/components/SearchBox/types.ts
steciuk: Renamed to `shouldRenderInvalidStyles` in https://github.com/canonical/pragma/pull/625/changes/e89061e9af2567f0a26c7ae78c2625c685be14eb

packages/svelte/ds-app-launchpad/src/lib/components/SearchBox/SearchBox.svelte
steciuk: Made `SearchButton` a subcomponent that can be passed as `SearchBox`'s `children` to gain full controll over its props. (https://github.com/canonical/pragma/pull/625/changes/e89061e9af2567f0a26c7ae78c2625c685be14eb)

packages/svelte/ds-app-launchpad/src/lib/components/SearchBox/styles.css
steciuk: I checked via BrowserStack, and it seems to be working with no issues! :)

packages/svelte/ds-app-launchpad/src/lib/components/SearchBox/types.ts
```diff
@@ -0,0 +1,30 @@
+/* @canonical/generator-ds 0.10.0-experimental.4 */
+
+import type { Snippet } from "svelte";
+import type { HTMLInputAttributes } from "svelte/elements";
+import type { TextInputPrimitiveProps } from "../common/index.js";
+
+export interface SearchBoxProps extends Omit<TextInputPrimitiveProps, "type"> {
```
jmuzina, line 7: There are a lot of props showing up in the storybook docs that don't seem to make sense here. I think it might be due to the way we extend props, there are some HTML native props being included in the docs, even when they aren't really applicable.

This is just a subset of them - it is extra confusing because they have no description so it's not clear what I should be doing with them as a consumer 😅 :

- files
- dirname
- formmethod
- multiple
- src

steciuk: It's a Storybook + Svelte issue with how docgen works for inherited attributes. When we started developing components, just a simple `Omit` was enough to flood your story with all the native junk.

I filed [a PR](https://github.com/storybookjs/storybook/pull/32173) ([issue](https://github.com/storybookjs/storybook/issues/32171)) that made it a bit better, but it isn't perfect. A little too much type jugglery and you end up with what you observed.

Sadly we don't have a solution for that just yet, but I plan to investigate it further :)

MichaelKapustey (approved):

PR #632, opened 2026-05-07
feat(svelte-ds-app-launchpad): Allow explicit TZs in Log component

steciuk:
## Done

`Log` only supported `UTC` and runtime `local` timestamps. In SSR, `local` resolves on the server, which can differ from the browser timezone and lead to inconsistent rendered output.

## Solution

- Extend `Log` to accept explicit timezone values in addition to `UTC` and `local`, and update timestamp formatting to use `Intl.DateTimeFormat`.
- Allow optional `timestampFormatter` `Log` prop to be provided, granting the consumer full control over the timestamp rendering if needed. 
- Add tests covering UTC, local, explicit timezones, invalid timezone fallback.
- Set vitest `TZ` to `"America/Los_Angeles"` to ensure deterministic TZ-related tests.

### PR readiness check


packages/svelte/ds-app-launchpad/src/lib/components/Log/common/Line/utils/formatTimestamp.ts
```diff
@@ -1,38 +1,66 @@
...
+    fractionalSecondDigits: 3,
+    hour12: false,
+  };
+
+  if (timeZone !== "local") {
+    formatterOptions.timeZone = timeZone;
+  }
+
+  try {
+    const formatter = new Intl.DateTimeFormat("en-US", formatterOptions);
```
goulinkh: Would it make sense to make the locale also configurable?

packages/svelte/ds-app-launchpad/src/lib/components/Log/common/Line/utils/TimestampFormatter.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      const paddedDate = new Date("2024-02-03T04:05:06.007Z");
+      expect(formatTimestamp(paddedDate, "UTC")).toBe(
+        "2024-02-03 04:05:06.007",
+      );
+    });
+  });
+
+  describe("local timezone", () => {
+    it("formats using runtime timezone", () => {
+      expect(process.env.TZ).toBe("America/Los_Angeles");
```
goulinkh: Not sure I understand this test. What's the reason behind the runtime environment variable?

goulinkh (changes requested):

packages/svelte/ds-app-launchpad/src/lib/components/Log/common/Line/utils/TimestampFormatter.test.ts
steciuk: Notice TZ set in `packages/svelte/ds-app-launchpad/vite.config.ts`. This is to ensure stable and predictable `local` TZ tests.

This particular assertion is to ensure `vite.config.ts` is setup the way this test expects it to be, so that:

in (UTC): `2024-01-15 14:30:45.123` -> out (America/Los_Angeles): `"2024-01-15 06:30:45.123"`

packages/svelte/ds-app-launchpad/src/lib/components/Log/common/Line/utils/formatTimestamp.ts
steciuk: Good call. I'll add a way to provide custom formatters.

packages/svelte/ds-app-launchpad/src/lib/components/Log/common/Line/Line.svelte
```diff
@@ -1,9 +1,13 @@
 <!-- @canonical/generator-ds 0.10.0-experimental.5 -->
+<script lang="ts" module>
+  import { TimestampFormatter } from "./utils/TimestampFormatter.js";
+
+  const defaultFormatter = new TimestampFormatter();
```
steciuk: Maybe this should be created during `Log` component initialization and provided to `Line` via context, so that its cache doesn't grow indefinitely, e.g., in the server's _global_ scope? 🤔

MichaelKapustey: It should. If you went all the way with caching

packages/svelte/ds-app-launchpad/src/lib/components/Log/common/Line/utils/TimestampFormatter.ts
```diff
@@ -0,0 +1,68 @@
+import type { TimeZone } from "../../../types.js";
+
+export class TimestampFormatter {
```
MichaelKapustey: I'd prefer this to be a pure function with formatter passed outside. And for the formatter -> it also can be a pure function with memoization applied

steciuk: Why would you prefer that?

1. As a class we get the cache-to-instance scoping "for free".
2. The API follows more closely that of `Intl.DateTimeFormat`.
```ts
const dtf = new Intl.DateTimeFormat();
const tf = new TimestampFormatter();

dtf.format();
tf.format();
```
(3. Classes in Svelte [have some runes-related superpowers](https://svelte.dev/docs/svelte/$state#Classes). We don't need/use them here, but generally they made me like classes much more than I used to!)

steciuk: Hmm... but (1.), probably won't apply if we follow the idea from [the other comment](https://github.com/canonical/pragma/pull/632#discussion_r3203111990) anyway.

Let me do that.

MichaelKapustey: I believe that it will look shorter and more readable. For example. here you manage the "cache" yourself but the memoization will handle it for you.

packages/svelte/ds-app-launchpad/src/lib/components/Log/common/Line/Line.svelte
steciuk: > If you went all the way with caching.

If we do that, we no longer really need `DateTimeFormat` caching, as they will be local to the `Log` component initialization context. See changes in https://github.com/canonical/pragma/pull/632/changes/b51c0987fa0dc8679963e77e0b82ea2f78a005f7. :)

packages/svelte/ds-app-launchpad/src/lib/components/Log/common/Line/utils/TimestampFormatter.ts
steciuk: Updated the PR. Let me know what you think of the current `Log` context scoping of the formatter and `use…` function instead of a class.

packages/svelte/ds-app-launchpad/src/lib/components/Log/utils/useDefaultTimestampFormatter.svelte.ts
```diff
@@ -0,0 +1,55 @@
...
+  hour12: false,
+};
+
+/**
+ * Creates a default timestamp formatter that updates when the timezone changes. The formatter formats dates as `YYYY-MM-DD HH:mm:ss.SSS` in UTC, runtime local time, or an explicit timezone.
+ *
+ * @param getTimeZone A function that returns the current timezone.
+ * @returns An object with a reactive `format` function that formats dates according to the current timezone.
+ */
+export function useDefaultTimestampFormatter(getTimeZone: () => TimeZone) {
```
MichaelKapustey, line 20: Can it be just "timezone" not getTimeZone?

steciuk: We explicitly listed using `get…` as [a Svelte code standard](https://github.com/canonical/code-standards/blob/main/docs/svelte.md#do-6) to pass reactive reads to `use…` functions.

Would you like to revisit this topic?

packages/svelte/ds-app-launchpad/src/lib/components/Log/utils/useDefaultTimestampFormatter.svelte.ts
```diff
@@ -0,0 +1,55 @@
...
+  const dateTimeFormat = $derived.by(() => {
```
MichaelKapustey, line 21: I would prefer if you have two separate files. One file is returning Intl object based on the timezone value. A second one that is doing the formatting using intl object and date. Intl object first.
You combine both in Log component. And that's where you use svelte magic (derived)
This will keep formatting functions only responsible for 1) providing right formatter for a giving timezone 2) returning string of correct shape

packages/svelte/ds-app-launchpad/src/lib/components/Log/utils/useDefaultTimestampFormatter.svelte.ts
steciuk: Unless... you mean passing `timezone` (value) instead of a `() => Timezone` (function) and not the parameter name itself. If that's the case, then no. It **has** to be a function to make sure we do the read in a reactive context (`$derived` in this case). Passing value would break the reactivity link.

MichaelKapustey: Is it going to be a reactive value? Because I've seen something like () => 'UTC'. 
May have missed it

packages/svelte/ds-app-launchpad/src/lib/components/Log/utils/useDefaultTimestampFormatter.svelte.ts
MichaelKapustey: I see now what you're trying to do. An effect. Maybe I'll have a third file as an effect combining the first two. Or keep it as is. Up to you.

packages/svelte/ds-app-launchpad/src/lib/components/Log/utils/useDefaultTimestampFormatter.svelte.ts
steciuk:
```ts
  let {
    // ...
    timeZone = "UTC",
    // ...
  }: LogProps = $props();

  const defaultFormatter = useDefaultTimestampFormatter(() => timeZone);
```

Yes, it's used with a prop passed to `Log` component (reactive).

packages/svelte/ds-app-launchpad/src/lib/components/Log/utils/useDefaultTimestampFormatter.svelte.ts
steciuk: Not sure if I understand what you mean by that. What is the effect that you are referring to?

Personally I don't see much value in splitting the `useDefaultTimestampFormatter`. It's concise and IMO has a clear concern. However, I'm not against it if you believe it will help with readability.

steciuk: Oh, I think I know what you meant. Let me do the change.

steciuk: So if I understand, what you wished for is to have something like

```ts
// Log.svelte
  const defaultDateTimeFormat = $derived(getDefaultDateTimeFormat(timeZone));

  const defaultFormatter = {
    format: (date: Date) => defaultTimestampFormat(defaultDateTimeFormat, date),
  };
```

But now that I think of it, I'm no longer sure I like the split. Let me explain why.

`defaultTimestampFormat` uses `Intl.DateTimeFormat.formatToParts` and then picks the part types it requires (e.g., `year`, `month`, `day`, etc.). These are not guaranteed to be defined for **any**  `Intl.DateTimeFormat`. In our case, we use `new Intl.DateTimeFormat` with `"en-US"` and explicitly required options:

```
  year: "numeric",
  month: "2-digit",
  day: "2-digit",
  hour: "2-digit",
  minute: "2-digit",
  second: "2-digit",
  fractionalSecondDigits: 3,
 ```
 
which gives us that guarantee. If the `Intl.DateTimeFormat` creation and parts picking live in two separate functions (or worse - files), we significantly increase the risk of their drift.

An alternative would be to check in `defaultTimestampFormat` whether we obtain the parts we expect, but that would need to happen **on every timestamp formatting**, which I believe to be an unnecessary computational hit for not much of a benefit. WDYT?

Comment (steciuk) [reply packages/svelte/ds-app-launchpad/src/lib/components/Log/utils/useDefaultTimestampFormatter.svelte.ts:21]: `getDefaultDateTimeFormat` could have a test that enforces it to return an `Intl.DateTimeFormat` that provides parts that `defaultTimestampFormat` expects, but then again... imo it creates unnecessary cross-module ties and adds complexity rather than removes it.

Comment (MichaelKapustey) [reply packages/svelte/ds-app-launchpad/src/lib/components/Log/utils/useDefaultTimestampFormatter.svelte.ts:21]: Yeah. Makes sense. Its ok to rely on it if everything is in one file. Once its two separate ones you can't be sure. Leave it as is then

**Review by MichaelKapustey — APPROVED** [review]

File: packages/svelte/ds-app-launchpad/src/lib/components/Log/types.ts [hunk 7]
```diff
@@ -27,12 +19,47 @@ export interface LogProps extends BaseProps {
...
+  timestampFormatter?: never;
 }
 
+interface CustomFormatterLogProps extends CommonLogProps {
+  /**
+   * A custom timestamp formatter. If provided, this will be used instead of the default timestamp formatting.
+   *
+   * `timeZone` prop is ignored when using a custom formatter, as it's assumed that the custom formatter will handle timezones internally if needed.
+   */
+  timestampFormatter: { format: (date: Date) => string };
```
goulinkh: extract this into a type `TimestampFormatter`

packages/svelte/ds-app-launchpad/vite.config.ts
```diff
@@ -2,10 +2,15 @@ import { svelte } from "@sveltejs/vite-plugin-svelte";
 import { playwright } from "@vitest/browser-playwright";
 import { defineConfig } from "vitest/config";
 
+const TimeZone = "America/Los_Angeles";
```
goulinkh:
```suggestion
const TIME_ZONE = "America/Los_Angeles";
```

goulinkh (approved):

PR #352, opened 2025-10-01
feat(React Core): Build `<SkipLink>`

jmuzina:
## Done

- Builds the SkipLink component
- Drive-by: Updates the TSDoc template for `types.ts` in the React component generator. `/**` is necessary for props comments to be surfaced to Storybook documentation tables. 

Fixes [WD-26130](https://warthogs.atlassian.net/browse/WD-26130)

## QA

- Open [SkipLink docs](https://skip-link--67be30a0b85c4e44b6744158.chromatic.com/?path=/docs/skiplink--docs). Verify the documentation makes sense and the component functions as you would expect.
- Open the [hidden "Focused" story](https://skip-link--67be30a0b85c4e44b6744158.chromatic.com/?path=/story/skiplink--focused) (used for visual testing). Verify that it focuses the SkipLink on opening. You should see a [passing interaction in the bottom of the screen](https://github.com/user-attachments/assets/49156a49-a1da-48fd-be04-483e10954ed9)

### PR readiness check

- [x] PR should have one of the following labels:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [x] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [x] All packages define the required scripts in `package.json`:
  - [x] All packages: `check`, `check:fix`, and `test`.
  - [x] Packages with build steps: `build` to build the package for development or distribution, `build:all` to build **all** artifacts. See [CONTRIBUTING.md](../old/CONTRIBUTING.md#24-full-artifact-builds-buildall) for details. 

## Screenshots

[Screencast from 2025-10-01 17-23-01.webm](https://github.com/user-attachments/assets/a4e284ce-ebf0-4ce9-8379-689024c89f3c)

[WD-26130]: https://warthogs.atlassian.net/browse/WD-26130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #659, opened 2026-06-12
feat(svelte-wpe): Add `SkipLink` component

jmuzina:
## Done

Adds a `SkipLink` component. This is a direct Svelte port of the existing [global React `SkipLink`](https://github.com/canonical/pragma/tree/abbe034f4d810ca64d349c78a8504b1a38310fba/packages/react/ds-global/src/lib/SkipLink). 

## QA

- Open [stories](https://skiplink-svelte--6a1f1b4ab915d9a04a321108.chromatic.com/?path=/docs/experimental-skiplink--docs)
- Follow QA from https://github.com/canonical/pragma/pull/352
- Compare the docs / behavior against the [global skiplink](https://ds-react-global.canonical.com/?path=/docs/experimental-skiplink--docs) and verify they are in line

### PR readiness check

## Screenshots

image


packages/svelte/ds-app-wpe/src/lib/components/SkipLink/SkipLink.stories.svelte
```diff
@@ -0,0 +1,119 @@
...
+</script>
+
+<Story name="Default" asChild>
+	<SkipLink />
+	<main id="main" tabindex="-1">
+		<p>
+			Click inside this example box, then press Tab to focus the skip
+			link.
+		</p>
+		<!-- biome-ignore lint/a11y/useValidAnchor: for demo only -->
```
jademathre-canonical: My VS Code is still highlighting this. I was able to suppress it with:
`<!-- svelte-ignore a11y_invalid_attribute -->`.
Also, minor formatting, this should be 2 spaces per indent, not 4.

packages/svelte/ds-app-wpe/src/lib/components/SkipLink/SkipLink.svelte
```diff
@@ -0,0 +1,47 @@
...
+    ...rest
+  }: SkipLinkProps = $props();
+
+  const href = $derived(`#${mainId}`);
+</script>
+
+<a
+	class={[componentCssClassName, className]}
+	href={href}
+	tabindex={0}
```
jademathre-canonical, line 22: A `<a>` with an `href` is already in the tab index, so having this `tabindex` here is redundant, so it _could_ be removed. Removing it would also narrow the type in `types.ts`. That said, this is consistent with the React version, and would also require updating tests. So not needed, but maybe not worth removing?

jademathre-canonical (approved):
Update the ignore comments and it looks good to me

packages/svelte/ds-app-wpe/src/lib/components/SkipLink/SkipLink.svelte
```diff
@@ -0,0 +1,47 @@
...
+	{...rest}
+>
```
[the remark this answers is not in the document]
jmuzina: Yes, this is relying on the browser / environment obeying fragment autofocus. Need to see what the vulnerability of this is, and if we should consider enabling a programmatic focus mode.

jmuzina:
I watched again the video in the description of the [React SkipLink](https://github.com/canonical/pragma/pull/352), and it seems something has since either broken in its styling (maybe due to an underlying token change) or the styling opinions of SkipLink have changed in a way I'm not aware of.

From the initial PR last year:

image

Today on the [global React storybook](https://ds-react-global.canonical.com/?path=/story/experimental-skiplink--default):

image

So, this may be a regression that needs to be corrected in the global React skiplink as well - (CC @advl ). I'll take a look tomorrow

jmuzina:
Had the time to look into that visual regression. I suspect it was introduced in https://github.com/canonical/pragma/pull/457 , a lot of CSS tokens were replaced with "placeholders"  . Digging to see if there is any way to get this component to work with the core tokens (maybe something was missed in that token migration?), otherwise we may need to use the Launchpad tokens for the time being

jmuzina:
Talked with @advl , the underlying spacing tokens were adjusted and this is the right approach now. 

@advl I will merge this as-is, then create separate PRs to create `ds-app` and `ds-global` for Svelte, & move this component (and the others we are building) in there. The React skiplink styling can be fixed then too.

jmuzina:
First adding suggestions from @jademathre-canonical

packages/svelte/ds-app-wpe/src/lib/components/SkipLink/SkipLink.svelte
jmuzina: Like you said, it's consistent with the React version so I will keep it like this. But, at the time of upstreaming it I will bring it up as something we can trim from both implementations.

Issue #662, opened 2026-06-12
Pattern/grid components accept `children` but silently drop them


advl:
## Context

Batch 2 of the remediation program broadened `ds-global`'s vitest `include` patterns, resurrecting 20 test files (in `src/lib/patterns/` and `src/lib/grid/`) that were silently excluded by the `*.tests.*`-only pattern and have never run in CI.

The resurrected scaffold tests revealed that five components **accept `children` in their props but never render them**:

- `patterns/IconSection`
- `patterns/ChatSection`
- `patterns/CategoriesSection`
- `patterns/CategoriesSection/common/Category`
- `grid/GridCard`

Example: `Category` destructures `children` (`Category.tsx:15`) but renders only its prop-driven structure (`<h3>` + `<ul>` from `items`) — `renderToString(<Category items={[]}>Test content</Category>)` produces `<div class="ds category"><h3></h3><ul></ul></div>`.

## Decision needed

For each component, either:
1. **Render `children`** (e.g. as a trailing slot), or
2. **Remove `children` from the props type**, making the prop-driven API explicit.

Accept-but-ignore is the worst of both: consumers get silent data loss with no type error.

## Affected tests

The corresponding `renders children` unit tests are skipped with `TODO(#<this issue>)` markers, and the SSR tests assert structure only, pending this decision. Search for the issue number to find them all.

_Found via the P0 remediation program (Batch 2); relates to the maturity-criteria work (experimental components with unsettled APIs)._

PR #663, opened 2026-06-12
fix(react): SSR-safe TooltipArea and useHead; resurrect skipped ds-global tests

advl:
## Done

**Batch 2 of the remediation program — SSR correctness + dead-test resurrection.**

### 1. TooltipArea: hydration-safe portal (`@canonical/react-ds-global`)

`TooltipArea` decided *at render time* where the tooltip message lives (`typeof window !== "undefined" ? createPortal(…) : inline`), so server HTML (inline) and the first client render (portal) produced different trees — a hydration-contract violation. Now a new internal `useIsMounted` hook keeps the message inline through the first client render (matching server HTML) and relocates it into the `document.body` portal after mount. The inline server render is kept deliberately so the `aria-describedby` target exists in SSR HTML.

New tests: an **unmocked** hydration test (`renderToString → render(…, { hydrate: true })` with a console.error spy asserting no hydration messages, plus the post-mount portal relocation) — the existing `TooltipArea.tests.tsx` mocks `createPortal` and could not catch this — and a `TooltipArea.ssr.tests.tsx` asserting the message and `aria-describedby` are present in server HTML.

### 2. useHead: safe without a DOM (`@canonical/react-head`)

The effect's DOM mutations were guarded only by "is there an SSR collector?" — any environment without a `HeadProvider` **and** without a DOM (edge runtimes, node test environments) threw `ReferenceError: document is not defined`. The DOM work is now an extracted `applyHeadTagsToDOM(tags): Element[]` whose first line is the environment guard (no-op + `[]` without `document`). Behaviour in browsers is unchanged; the package's 100% coverage thresholds still pass. New node-environment test covers the guard and `renderToString` without a provider.

### 3. ds-global: resurrect 20 silently-skipped test files

`vitest.config.ts` included only `src/**/*.tests.*` — the 20 `*.test.tsx`/`*.ssr.test.tsx` files in `patterns/` and `grid/` **have never run**. Both naming conventions are now included (unit + ssr projects) until the repo-wide naming unification lands.

Running them for the first time surfaced 20 failures, triaged as agreed:
- **Fixed in-PR (test bugs):** the scaffold `applies custom className` tests queried the text node's element (`getByText`) instead of the component root — the components apply `className` correctly; assertions now target the root.
- **Skipped with `TODO(#662)` (real API question):** five components (`IconSection`, `ChatSection`, `CategoriesSection`, `Category`, `GridCard`) accept `children` but silently drop them. Whether to render children or remove them from the props type is an API decision tracked in #662; the `renders children` tests are `it.skip`-ed and the SSR tests assert structure only, all marked with the issue number.

**Net test delta for ds-global: 40 → 60 files running, 247 passing + 5 skipped (every skip references #662).**

## QA

```bash
cd packages/react/ds-global && bun run check && bun run test
cd packages/react/head && bun run check && bun run test
```

Both packages: all 3 validations pass; ds-global 60 files / 247 passed / 5 skipped; head 4 files / 17 tests with 100% coverage thresholds intact.

### PR readiness check

  - [x] Packages with build steps: `build` and `build:all` (unchanged by this PR).

## Screenshots

No visual changes (rendering-order and test-infrastructure only; the tooltip relocates to its portal one tick after mount, while still closed/hidden).

---
https://claude.ai/code/session_01TtnNBZp4Kv8EaTuPd5jbvr

---
_Generated by [Claude Code](https://claude.ai/code/session_01TtnNBZp4Kv8EaTuPd5jbvr)_


packages/react/head/src/lib/useHead.ts
```diff
@@ -28,6 +28,70 @@ function createLinkElement(attrs: Record<string, string>): HTMLLinkElement {
...
 }
 
+/**
+ * Apply head tags to the live DOM, returning the elements that were created
+ * so the caller can remove them on cleanup.
+ *
+ * Safe to call in environments without a DOM (edge runtimes, node test
+ * environments): it is a no-op there and returns an empty array.
+ */
+export function applyHeadTagsToDOM(tags: HeadTags): Element[] {
```
advl: separate file, follow https://github.com/canonical/web-code-standards

advl:
Superseded by #822. Most of this PR already landed on `main` independently: the `@canonical/react-head` refactor (`applyHeadTagsToDOM` extraction + `useHead` SSR rewrite + node tests) is byte-identical on main, and the resurrected ds-global grid/pattern tests + vitest config change landed (the config generalised into the shared `@canonical/vitest-config-react` factory). The only residual value — the shared `useIsMounted` hook, the TooltipArea refactor onto it, and real (unmocked-portal) SSR/hydration tests — is carried in #822, which additionally fixes the same hydration bug in ContextualMenu and SubMenu (both still gate their portal on `typeof window` on `main`). Recommend closing this as superseded once #822 lands.

advl:
Closing as superseded — as already noted in this thread (2026-07-15). Everything here either landed independently or became moot: the useHead SSR rewrite and resurrected ds-global tests landed via #759, the hydration-safe portal gating via #822, and TooltipArea itself was removed in #872. Welcome to reopen if anything turns out to be missing.

— Kimi (backlog cleanup, 2026-07-29)

PR #684, opened 2026-06-14
feat(i18n-core): native-Intl framework-agnostic i18n core

advl:
## Done

Introduces **`@canonical/i18n-core`** — the framework-agnostic heart of a native-`Intl` internationalization system. Zero runtime dependencies; pure TypeScript over the platform `Intl` API. Scaffolded with `summon`, conforms to `webarchitect library`.

- **Negotiation** (pure, SSR-safe): `negotiateLocale`, `parseAcceptLanguage`, `directionOf`, `isSupportedLocale`, `readCookie`
- **Messages**: `createTranslator` (`{var}` interpolation + `Intl.PluralRules`), `mergeCatalogs`
- **Formatting**: `createFormatters` — memoized `Intl` number / currency / date / time / relativeTime / list
- **`createLocaleSource`** — the cross-framework runtime channel: a live locale value whose `subscribe` honours the Svelte store contract (so it also drives React `useSyncExternalStore` and Lit controllers); in the browser it persists a cookie and reflects `<html lang dir>`, inert on the server

Design goal: the core owns everything that isn't framework reactivity, so the per-framework bindings stay thin and share one engine.

Drive-by: also includes the originating app-local i18n SSR negotiation seam in `apps/react/boilerplate-vite` that this package generalises.

**PR 1 of 3 (stacked):** **i18n-core** → i18n-react → i18n-lit + i18n-svelte.

## QA

```bash
bun install
bun run --filter @canonical/i18n-core check          # biome + tsc + webarchitect library
bun run --filter @canonical/i18n-core test            # 44 tests
bun run --filter @canonical/i18n-core test:coverage   # 100% statements/branches/functions/lines
```

### PR readiness check

- [ ] Suggested label: `Feature 🎁`
- [x] Follows [code standards](https://github.com/canonical/web-code-standards) — biome, `webarchitect library`, 100% coverage
- [x] Required scripts present: `check`, `check:fix`, `test`, `build`, `build:all`
- [ ] **New package** — first-time `npm publish --access public` + OIDC trusted publisher is a manual human step (2FA); not performed by automation
- [x] Add `no visual change` label (no UI in this package)

## Screenshots

N/A — no UI.

---
_Generated by [Claude Code](https://claude.ai/code/session_014FGRX9ya6w7pdCus26dqee)_


packages/runtime/i18n/.github/PULL_REQUEST_TEMPLATE.md
advl, line 1: Remove from PR, should not be there. Also curious it's generated from summon-package, this must be fixed as well.

packages/runtime/i18n/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../../configs/typescript/tsconfig.json",
```
advl, line 2: use pkg ref, not relative path

packages/runtime/i18n/tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+  "compilerOptions": {
+    "baseUrl": "src",
+    "types": ["node"],
+    "resolveJsonModule": true,
+    "skipLibCheck": true
```
advl, line 7: is this needed here ?

packages/runtime/i18n/README.md
```diff
@@ -0,0 +1,139 @@
...
+Pass the same `locale` to the client (`createLocaleSource({ initial: locale })`)
+so the server and first client render agree.
+
+## Right-to-left & accessibility
+
+`directionOf` resolves writing direction from `rtlLocales` (base languages); a
+sensible default set is `["ar", "he", "fa", "ur"]`. Flipping `dir` is necessary
+but not sufficient — author UI with CSS **logical properties**
+(`margin-inline-start`, `inset-inline`, `text-align: start`) and mirror
+directional icons so RTL is structural, not decorative.
```
advl, line 128: I wonder what is a good heuristic to set up rtl on the html/body based on the language detected

advl: That heuristic is built in: `directionOf(config, locale)` maps a locale's base language to `dir` via the configurable `rtlLocales` set (a sensible default is `["ar", "he", "fa", "ur"]`), and `documentAttrs(config, locale)` returns `{ lang, dir }` to render on `<html>` server-side. On the client, `createLocaleSource` reflects `<html lang dir>` on construction and on every change. So the rule is explicit — base language ∈ `rtlLocales` → `rtl` — rather than guessed, since a language tag alone can't reliably imply direction beyond the known RTL scripts (hence the overridable config set rather than magic).

---

apps/react/boilerplate-vite/src/lib/i18n/negotiateLocale.ts
```diff
@@ -0,0 +1,86 @@
...
+    .map((part) => {
+      const [tag, ...params] = part.trim().split(";");
+      const q = params.map((p) => p.trim()).find((p) => p.startsWith("q="));
+      const weight = q ? Number.parseFloat(q.slice(2)) : 1;
+      return {
+        tag: tag.trim().toLowerCase(),
+        weight: Number.isNaN(weight) ? 0 : weight,
+      };
+    })
+    .filter((entry) => entry.tag.length > 0)
```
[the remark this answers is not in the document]
advl: This `boilerplate-vite` file is the transitional app-local i18n seam that `@canonical/i18n-core` supersedes. I've applied the `q=0` fix in the package's `parseAcceptLanguage` instead (just pushed), and the app will be migrated to consume the package — at which point this copy is removed. Leaving the throwaway seam untouched to avoid churn.

---

packages/runtime/i18n/tsconfig.json
advl: It turns out it is needed. I removed it per this comment, but `build:all` then failed in CI (build-matrix, node 22 and 24) while passing locally — without `skipLibCheck`, `tsc` type-checks `node_modules` `.d.ts` (notably `@types/node`), whose patch version differs between my machine and CI's fresh install, surfacing a lib error there only. Restored it (as `router-core` does). The package-ref `extends` from the other comment is kept.

---

packages/runtime/i18n/tsconfig.json
advl: I tried the package ref, but `@canonical/i18n-core:build:all` then failed in CI (build-matrix, node 22 and 24) while passing locally. `@canonical/typescript-config` exposes its tsconfig only via `main` (no `exports`), so tsc's scoped-package resolution for `extends` behaves differently between my local bun install and CI's. I've reverted to the relative path `router-core` uses, which is CI-green. Happy to switch back to the package ref once the base config gets an `exports` map for its tsconfig — I can do that as a separate change to `configs/typescript` if you'd like.

---

packages/runtime/i18n/src/lib/createFormatters.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  it("reuses memoized formatters across repeated identical options", () => {
+    const formatters = createFormatters("en-US");
+    // The second call with the same options resolves the cached instance.
+    expect(formatters.number(1000)).toBe(formatters.number(1000));
+    expect(formatters.number(1000)).toBe("1,000");
+    expect(formatters.currency(5, "USD")).toBe(formatters.currency(5, "USD"));
+    expect(formatters.currency(5, "USD")).toBe("$5.00");
+  });
```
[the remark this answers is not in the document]
advl: Addressed — the memoization test no longer compares formatted strings. It stubs `Intl.NumberFormat` with a counting subclass and asserts on constructor invocations: three format calls (two with identical default options, one with distinct currency options) must construct exactly two instances, so a caching regression now fails the test (`createFormatters.test.ts` › "constructs one Intl instance per distinct options (memoized)").

---

packages/runtime/i18n/src/lib/parseAcceptLanguage.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it("treats a malformed q-weight as zero", () => {
+    expect(parseAcceptLanguage("en;q=x,fr")).toEqual(["fr", "en"]);
+  });
+
+  it("drops empty segments", () => {
+    expect(parseAcceptLanguage("en,,fr")).toEqual(["en", "fr"]);
+  });
+});
```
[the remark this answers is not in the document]
advl: Addressed alongside the `q=0` fix — the stale expectation is gone: malformed-`q` segments are now dropped rather than kept at weight 0, and there's a dedicated regression case ("drops q=0 and malformed-q entries (not acceptable)") covering both `en;q=0` and `en;q=x`.

---

PR #693, opened 2026-06-25
chore(Svelte): boilerplate Svelte ds-global package

jmuzina:
## Done

Adds the skeleton of the core svelte ds-global package for core versions of Svelte components to live.

## QA

Check that storybook runs properly, you can do this by checking the [published Chromatic storybook](https://6a3ca65a785e76b915a40851-lziusoqkro.chromatic.com/), or cloning and using `bun run storybook`. 

### PR readiness check


packages/svelte/ds-global/src/lib/styles/index.css
advl, line 1: Likely not necessary, the root one should be enough

packages/svelte/ds-global/src/lib/internal.ts
advl, line 1: cleanup

packages/svelte/ds-global/package.json
```diff
@@ -0,0 +1,94 @@
...
+      "types": "./dist/index.d.js",
+      "import": "./dist/index.js",
+      "svelte": "./dist/index.js"
+    },
+    "./styles.css": "./dist/index.css",
+    "./internal": {
+      "types": "./dist/internal.d.js",
+      "import": "./dist/internal.js",
+      "svelte": "./dist/internal.js"
+    }
```
advl: cleanup

packages/svelte/ds-global/package.json
```diff
@@ -0,0 +1,94 @@
...
+  },
+  "files": [
+    "dist",
+    "!dist/**/*.test.*",
+    "!dist/**/*.fixtures.*",
+    "!dist/**/*.stories.*"
```
advl: should not be built in the first place (I think ?)

packages/svelte/ds-global/package.json
```diff
@@ -0,0 +1,94 @@
...
+    "@types/bun": "^1.3.10",
+    "@vitest/browser": "^4.0.18",
+    "@vitest/browser-playwright": "^4.0.18",
+    "playwright": "^1.58.2",
+    "storybook": "^10.3.1",
+    "svelte-check": "^4.4.5",
+    "typescript": "^5.9.3",
+    "vite": "^8.0.0",
+    "vitest": "^4.0.18",
+    "vitest-browser-svelte": "^2.0.2"
```
advl: Ensure version alignment where possible with non svelte packages, also the generator (summon package)

packages/svelte/ds-global/tsconfig.build.json
advl, line 1: Might be worth having a second look. React for reference https://github.com/canonical/pragma/blob/main/packages/react/ds-global/tsconfig.build.json
I would recommend differences to be explainable
I am worried by : 
- moduleresolution bundler
- module esnext should not have to be mentioned, it should be the tsconfig default  nodenext
- it's unclear what the type build and source map output would be - in which folder

packages/svelte/ds-global/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "@canonical/typescript-config-svelte",
+  "compilerOptions": {
+    "verbatimModuleSyntax": true,
+    "skipLibCheck": true,
+    "moduleResolution": "nodenext"
```
advl: could you possibly refresh my mind on this ?

packages/svelte/ds-global/vite.config.ts
advl, line 1: I would recommend a cleanup pass

advl (changes requested):
Please test that the package we are looking at matches 100% `summon package [mypackagee]` with the storybook option. Can be a follow-up

Nit : 
- Update generated files to 0.29 with summon component

Notes for later
- Need convention to import sb-specific styles (e.g. react/global-form is different
- Need for global form to upstream the stories order
- the vite configs need to be refreshed and trimmed at the monorepo level

packages/svelte/ds-global/tsconfig.json
jmuzina: - verbatimModuleSyntax: true - TypeScript 5 feature that enforces ESM import/export syntax. Required for  Svelte component imports, see [svelte docs](https://svelte.dev/docs/svelte/typescript#tsconfig.json-settings). This should probably be added upstream to the Svelte tsconfig package.
  - skipLibCheck: true - Skips type checking of declaration files and dependencies, helps avoid cluttering CI and dev processes with type errors that can be hard or impractical to resolve when working with many dependencies, especially in a monorepo.
  - moduleResolution: "nodenext" - Required for Node.js ESM module resolution. Without this, TypeScript can't resolve packages like @storybook/addon-svelte-csf or vite. It's configured in our base svelte tsconfig, but it seems I get issues resolving the above-mentioned packages if I remove it from the package tsconfig... needs some investigation some other time I think

packages/svelte/ds-global/package.json
jmuzina: No, this is in fact necessary as `svelte-package` does not provide a way to ignore files from being built, so we need to exclude files we don't want to publish here.

packages/svelte/ds-global/vitest.config.ts
```diff
@@ -0,0 +1,33 @@
...
+    test: {
+      projects: [
+        {
+          extends: true,
+          resolve: {
+            conditions: ["browser"],
+          },
+          test: {
+            name: "client",
+            environment: "jsdom",
```
steciuk: Since we identified some time ago that [Playwright is a more prospective solution](https://docs.google.com/document/d/1HHhTD80-nya5CkrWOzWitDsehDdILz0mkPPHzIySahs/edit?tab=t.muzyzbnmoyef#heading=h.ylxem6xryo7x) and likely will become the default repo-wide, I was wondering if it makes sense to configure the package with browser testing right from the start.

Setting it up now would save us from a lot of extra effort down the line. The `ds-app-launchpad` components already have Playwright tests. If we stick with `jsdom` for now, we'd have to rewrite the tests during upstream to `jsdom`, only to port them back to Playwright later when we eventually migrate the rest of the monorepo.

What do you all think? CC: @goulinkh

jmuzina: Makes sense to me. I went with jsdom here only because I saw it's notated as "experimiental" in the launchpad package, but if we're aligned at the DS level I can put it here in a core package as well, wdyt @advl ?

packages/svelte/ds-global/tsconfig.json
advl: For nodenext, I would recommend avoiding the override if we don't understand were it comes from.
verbatimModuleSyntax: true - TypeScript 5 feature that enforces ESM import/export syntax. Required for Svelte component imports, see svelte docs. This should probably be added upstream to the Svelte tsconfig package. >> can be added to the config package if you believe it is idiomatic. Please add a node in the readme about it

packages/svelte/ds-global/vitest.config.ts
advl: Let's do it, but in a follow-up, so we enable it for react at the same time as well.

packages/svelte/ds-global/vitest.config.ts
```diff
@@ -0,0 +1,33 @@
+import { defineConfig, mergeConfig } from "vitest/config";
+import viteConfig from "./vite.config.js";
+
+export default mergeConfig(
+  viteConfig,
+  defineConfig({
```
advl: you might wanna set v8 testing goals as other packages have

packages/svelte/ds-global/tsconfig.build.json
advl: In other words there is unexplained drift with what a "build" tsconfig looks like for react

advl (changes requested):
Comments related to the tsconfig. the rest looks good

packages/svelte/ds-global/vitest.config.ts
jmuzina: added a note identical to what is set in `react/ds-global`

```ts
// Testing posture: Measured — Chromatic is primary visual gate; unit tests cover logic
```

packages/svelte/ds-global/tsconfig.build.json
jmuzina: The type/source outputs drift is due to the difference between svelte-package and tsc. Here's what I found:

(AI+)

> React's tsconfig.build.json sets rootDir/outDir/declarationDir/sourceMap/declaration because React packages build by running tsc directly, which reads and obeys those settings. Svelte packages build via svelte-package -o dist -i src/lib --tsconfig .... I read @sveltejs/package's source (typescript.js: emit_dts): it always overrides declarationDir itself to a private temp folder, and svelte2tsx's emitDts forces rootDir to the -i input path too — regardless of what's in the tsconfig. So React-style rootDir/outDir/declarationDir settings would be silently ignored by svelte-package. ds-global's current tsconfig.build.json (just extends + exclude, no output paths) is correct for this tool, not an oversight.


The other concerns were addressed in https://github.com/canonical/pragma/pull/693/commits/a942f319bf40355de402a492ecdc7863b6105d94

packages/svelte/ds-global/vitest.config.ts
goulinkh: I have created a ticket to track this: https://github.com/canonical/pragma/issues/729

Please keep in mind, the way tests are written with jsdom are very different from vitest browser mode with playwright. So having components with tests will create a tech dept during the migration similar to the current state of existing react packages in pragma.

packages/svelte/ds-global/.storybook/preview.ts
```diff
@@ -0,0 +1,9 @@
+import previewConfig from "@canonical/storybook-config/preview";
+
+import "../src/lib/index.css";
+
+const preview = {
+  ...previewConfig,
```
goulinkh, line 6: Nit, could you please ensure this spread is still needed, I am not sure if I have missed this during the initial implementation, as it could be a fixed bug.

advl (approved):

packages/svelte/ds-global/vitest.config.ts
```diff
@@ -0,0 +1,34 @@
+// Testing posture: Measured — Chromatic is primary visual gate; unit tests cover logic
```
goulinkh: Recent version of vitest docs suggest including the configuration inside the unified `vite.config.ts`, could this be merged into a single `vite.config.ts` similar to https://github.com/canonical/pragma/blob/main/packages/svelte/ds-app-launchpad/vite.config.ts

goulinkh (changes requested):
LGTM, I have included a couple of small comments and filed a follow up issue not covered in this PR: https://github.com/canonical/pragma/issues/729

packages/svelte/ds-global/.storybook/preview.ts
jmuzina: seems like it is not needed, I changed  the preview file to align with the React one without issue.
https://github.com/canonical/pragma/pull/693/commits/808d228715f793fe153ad926a6932ea728bfd48c

packages/svelte/ds-global/vitest.config.ts
jmuzina: I like this, worth noting that it is different from the React global package approach of using two separate files. I adjusted in https://github.com/canonical/pragma/pull/693/commits/190747dc7ae3f92f11c896a2f3ef4114f97cbebf but noting for @advl for the React side

goulinkh (approved):

jpmartinspt (approved):

PR #660, opened 2026-06-12
feat(svelte-ds-app): Port React core layouts

jmuzina:
## Done

Svelte port of https://github.com/canonical/pragma/pull/656 , implementing:

1. `ApplicationLayout`, the outermost shell of an application. Splits a navigation rail and an app's main content side-by-side.
2. `ViewLayout`, a "sub-shell" for the main content area of an application, rendered inside an `ApplicationLayout` typically. Defines a main area and optional aside
3. `ContentLayout`, a content grid for the meat of an application (the main area from `ViewLayout`). The rest of the app consumes `ContentLayout`'s grid.

Depends on #694 for creation of the ds-app Svelte package.

## QA

- Open the [storybook](https://svelte-application-layout--6a3ca6a1c2cb863811b05c34.chromatic.com/). Ensure the docs and behavior of each of the layouts exactly match the [React storybook](https://ds-react-app.canonical.com/).
- Pay special attention to responsiveness. 
  - ContentLayout is responsive and lays out cards according to an intrinsic grid strategy. 
  - ApplicationLayout and ViewLayout are not responsive (reduction in space does not cause navigation aside rails to collapse), following the current React layout behavior.

### PR readiness check

- [x] PR should have one of the following labels:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [x] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [x] The code follows the appropriate [code standards](https://github.com/canonical/web-code-standards)
- [x] All packages define the required scripts in `package.json`:
  - [x] All packages: `check`, `check:fix`, and `test`.
  - [x] Packages with build steps: `build` to build the package for development or distribution, `build:all` to build **all** artifacts. See [CONTRIBUTING.md](../old/CONTRIBUTING.md#24-full-artifact-builds-buildall) for details.
- [ ] If this PR introduces a **new package**: first-time publish has been done manually from inside the package directory using `npm publish --access public` (first-time publishing is not automated). Run `bun run publish:status` from the repo root to verify. Then configure an OIDC [trusted publisher](https://docs.npmjs.com/trusted-publishers) for the package on npmjs.com (repo `canonical/pragma`, workflow `tag.yml`) and set publishing access to disallow tokens, so the automated release workflow can publish future versions.
- [ ] If this PR does not require visual testing, add the `no visual change` label to skip Chromatic.

## Screenshots

image

Issue #451, opened 2026-03-04
Track removal of Svelte-specific Biome overrides after biomejs/biome#8590


steciuk:
We currently keep Svelte-specific lint overrides in the Biome config due to missing "cross-language" lint support in Biome. Once biomejs/biome#8590 is resolved and released, we should remove these temporary overrides.

The current workaround disables Svelte false positives for:

- `noUnusedVariables`,
- `noUnusedImports`,
- `useImportType`,
- `useConst`.

These were centralized in the common Biome config as part of PR #449.

Issue #410, opened 2026-02-04
[RFC] Svelte styling: balancing consistency with Svelte idioms


steciuk:
Following Pragma's guidelines for the development of [Svelte components for Launchpad UI](https://ds-svelte-global.canonical.com/) let me gather some thoughts on the current approach to styling components. I would like to share some concerns and open a discussion on how we can ensure Pragma delivers the best possible maintainer and consumer experience.

Currently, both React and Svelte components are styled by assigning a _"scoped class"_ to the root element of each component and using that class as the main selector in the component's CSS.

```svelte
<script>
  const componentCssClassName = "ds component-name";
</script>

<div class={componentCssClassName}>...</div>

<style>
  .ds.component-name {
    /* component styles */
  }
</style>
```

I understand the intent is to maintain strict consistency across React and Svelte. However, I'd like to understand if we might be fighting against Svelte's compiler strengths here. Since Svelte [handles scoping automatically](https://svelte.dev/docs/svelte/scoped-styles), replicating the React scoping pattern introduces high specificity and friction for tooling.

I have outlined three specific areas where the current pattern might lower the quality and usability of our Svelte components.

> [!NOTE]
> **How does Svelte scope styles?**
>
> 1. It generates a unique hash for each component (e.g., `svelte-abcxyz`).
> 2. This hash is added as a class to the component's HTML elements.
> 3. The compiler transforms your CSS selectors to include this hash (e.g., `button` becomes `button.svelte-abcxyz`).

### 1. High Specificity

The current pattern results in a high specificity selector by default: `(0,3,0) = .ds + .component-name + .svelte-xxxxxx`.

This makes it more difficult to override styles when necessary than it needs to be. As seen in the example below, a consumer trying to style a component will fail, because their specificity `(0,2,1)` is lower than the component's default specificity.

```svelte
<!-- Component.svelte -->
<script>
  const { class: className } = $props();
  const componentCssClassName = "ds component-name";
</script>

<button class={[componentCssClassName, className]}>Content</button>

<style>
  .ds.component-name {
    /* Specificity: (0,3,0) */
    color: red;
  }
</style>

<!-- App.svelte -->
<script>
  import Component from './Component.svelte';
</script>

<div>
  <Component class="custom-class" />
</div>

<style>
  div :global(.custom-class) {
    /* Specificity: (0,2,1) */
    color: blue;
  }
</style>
```
[...]

PR #694, opened 2026-06-25
chore(Svelte): create base Svelte ds-app package

jmuzina:
## Done

Adds the skeleton of the core svelte ds-app package. It depends on the ds-global package from #693 and playwright migration in #738 .

Unblocks #660 

## QA

Check that storybook runs properly, you can do this by checking the published Chromatic storybook, or cloning and using `bun run storybook`. 

### PR readiness check


MichaelKapustey (dismissed):
Many decisions made here were only made in the context of design system components development
Now they are silently put into the boilerplate that every team in canonical is supposed to use.

For exampel I believe that we should use the svelte way of styling instead of relying on css imports as a default. 
importing css has literally no benefits beside "the react does it this way"
svelte styles provide you with a compile time safety, editor support/auto complete, unused css removal and saves you time while onboarding a developer who has previous experience in svelte

Biome is also a questionable choice when it comes to svelte

jmuzina:
To be clear @MichaelKapustey this is not a Svelte boilerplate package that would define the opinions of future Svelte projects, this is a components package for the design system. The word "boilerplate" from the title here was moreso used to indicate that the PR is creating the package itself, I'll rename it to be clearer.

But, you make a fair point on both styling and Biome:

#### CSS

We had a discussion that did not really resolve satisfactorily in https://github.com/canonical/pragma/issues/410 and that becomes apparent again here. I agree with the benefits you listed of using Svelte's styling opinions as mentioned [here](https://github.com/canonical/pragma/issues/410#issuecomment-3853338422). @advl @steciuk @goulinkh , what are your thoughts here.

#### Biome

It's true that Biome is [experimental](https://biomejs.dev/internals/language-support/#html-super-languages-support) for Svelte. Maybe this is a discussion we should have had in the [ds-global PR](https://github.com/canonical/pragma/pull/693), but let's check now - 

@steciuk AFAIK you have the most insight on the state of Biome for Svelte, and you authored https://github.com/canonical/pragma/issues/451 detailing one of the biggest pending issues for using Biome in Svelte and how we're getting around it. What is your perspective here?

MichaelKapustey:
If this package is only supposed to be a container for svelte design system components (like ds-app-launchpad) and is not ment to shape svelte application development, then it's not the right time and place to have those discussions once again. We may have them separately, case by case. Maybe during a working group meeting. Am I understanding the purpose correctly now, @jmuzina ?

jmuzina:
> this package is only supposed to be a container for svelte design system components (like ds-app-launchpad) and is not meant to shape svelte application development

Yes, that's right. It is true though that this package, and ds-global, will be very influential in the landscape of Canonical's Svelte packages as they are at the core of the DS. They may take opinions that are very DS-specific to further the DS's consistency goals, but the work of product teams should be anticipated to feed into this system in the long run as work is upstreamed. Which leads very well to your next point:

>  it's not the right time and place to have those discussions once again

The impetus behind publishing these packages now is to reduce the increasing amount of friction coming from multiple teams having Svelte needs. The DS tiering system assigns additional scrutiny to crossteam packages, so we are likely going to resurface a lot of these questions in the coming weeks. This question is very fundamental: on which problems do we allow framework-specific packages to have framework-specific opinions, and on which do we prefer more consistency between all packages?

It's hard to say whether this is the "right time" for the discussion. Personally I think there is risk in subjecting the core Svelte components to opinions that product team Svelte developers would not typically follow, like file-based CSS imports. It will create more friction at upstreaming time. We can optimistically say that we need not worry about this as much in the world of AI being able to apply any delta between Svelte styles and DS styles much more automatically, thus making CSS consistency more achievable, but this is an untested assumption so far. If we leave the convention unchanged here, this would be something we will have to observe from contributors to see if the styling convention is posing an undue entry barrier.

I will note, from my experience implementing a few of the React ds elements in Svelte these past two days, it does feel very silly to manually reimplement the same `styles.css` file in the different frameworks. If we're going to allow the flexibility of separately declaring style files, why not let them flex to match the needs of each implementation? But, perhaps a discussion for another time.

advl:
Hi @MichaelKapustey 

Thanks for bringing up those questions. Here are some tentative answers that I would submit to your consideration and hope we can discuss together, perhaps planning a dedicated session as early as next week : 

> For exampel I believe that we should use the svelte way of styling instead of relying on css imports as a default. importing css has literally no benefits beside "the react does it this way"

The question of using pure css files is actually not related to React. Instead it is an embodiement of the principle [VI in our constitution](https://github.com/canonical/pragma/blob/main/CONSTITUTION.md#vi-no-magic), "No Magic". My reading is that the compilation step for css doesn't add value to css that is in the first place, well written. If you believe this is a hindrance to having idiomatic svelte, I am happy to take the time to discuss this further with you and do a full architecture benchmark together.

> Biome is also a questionable choice when it comes to svelte

I see Biome as a tool led by a team of efficiently operating, organized, highly motivated engineers that have an understanding of the issues of the classical javascript tooling, and that are willing to invest the time and effort to modernize and unify javascript development tooling so that other teams can benefit from a lighter maintenance workload. I would recommend taking a look at how dynamic the svelte issues are on their github tracker. Biome also offers the advantage of unified tooling at the level of the monorepo. 

I feel it is also important to precise consumers of the DS do not need to use biome if they not wish to.

@jmuzina I value your comment very much. We need to make progress on the Svelte components, and keep the ability to discuss architecture topics in a non blocking way. 

@MichaelKapustey please be welcome to book my agenda for any of those topics for us to look at it more in depth. I am not against revisions to the package boilerplate later based on how our conversation goes.

packages/svelte/ds-app/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "svelte-check": "^4.4.5",
+    "typescript": "^5.9.3",
+    "vite": "^8.0.1",
+    "vitest": "^4.0.18",
+    "vitest-browser-svelte": "^2.0.2"
+  },
+  "peerDependencies": {
+    "svelte": "^5.47.1"
+  }
+}
```
jademathre-canonical, line 97: The global `package.json` has this appended to the end:
```
  "nx": {
    "tags": [
      "playwright"
    ]
  }
```
Should probably be included as well.

packages/svelte/ds-app/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Styles
+
+Import the main styles package in your application entry point:
+
+```tsx
+import "@canonical/styles";
+```
+
+`@canonical/svelte-ds-app/styles.css` imports `@canonical/styles`, which provides the global design tokens (colour, spacing, typography). Each component in this package co-locates its own component-level tokens in a `styles.css` file next to the component source. These component tokens reference the global tokens from `@canonical/design-tokens` and are included automatically when the component is imported.
```
jademathre-canonical: `@canonical/svelte-ds-app/styles.css` doesn't exist

packages/svelte/ds-app/README.md
```diff
@@ -0,0 +1,81 @@
...
+Playwright browsers must be installed once before running client tests:
+
+```bash
+bunx playwright install chromium firefox webkit
+```
+
+Use watch mode during development:
+
+```bash
+bun run test:vitest:watch
```
jademathre-canonical: `bun run test:vitest:watch` isn't a script. should be bun run test:watch. The same typo exists in ds-global's readme.

jmuzina: addressed in both packages: https://github.com/canonical/pragma/pull/694/commits/19f0f75ef348a620447f3b6f37b7cc987a235abc

packages/svelte/ds-app/README.md
jmuzina: addressed: https://github.com/canonical/pragma/pull/694/commits/19f0f75ef348a620447f3b6f37b7cc987a235abc

packages/svelte/ds-app/package.json
jmuzina: Addressed: https://github.com/canonical/pragma/pull/694/commits/19f0f75ef348a620447f3b6f37b7cc987a235abc

jademathre-canonical (approved):

steciuk (approved):

packages/svelte/ds-app/src/lib/Welcome.mdx
```diff
@@ -0,0 +1,20 @@
+import { Meta } from "@storybook/addon-docs/blocks";
+import { PackageInfo } from "@canonical/storybook-helpers";
+
+<Meta title="Introduction" />
+
+## Package
+
+<PackageInfo
+  name="@canonical/svelte-ds-app"
+  version="0.29.0-experimental.0"
```
advl: update

packages/svelte/ds-app/biome.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
+  "extends": ["@canonical/biome-config"],
+  "files": {
+    "includes": ["!.build-cache"]
```
advl, line 5: svelte specific config to not have per package override ?

packages/svelte/ds-app/package.json
```diff
@@ -0,0 +1,97 @@
+{
+  "name": "@canonical/svelte-ds-app",
+  "description": "App-level components for the Pragma design system.",
+  "version": "0.29.0-experimental.0",
```
advl: version

packages/svelte/ds-app/package.json
```diff
@@ -0,0 +1,97 @@
...
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js",
+      "svelte": "./dist/index.js"
+    }
+  },
+  "files": [
+    "dist",
+    "!dist/**/*.test.*",
+    "!dist/**/*.stories.*",
+    "!dist/**/*.mdx"
```
advl, line 19: Wrong pattern - but seems to be tied to svelte tooling ?

packages/svelte/ds-app/package.json
```diff
@@ -0,0 +1,97 @@
...
+    "test": "vitest --run",
+    "test:watch": "vitest",
+    "test:server": "vitest --run --project=server",
+    "test:ssr": "vitest --run --project=ssr",
+    "test:client": "vitest --run --project=client",
+    "test:client:chromium": "vitest --run --project=client --browser=chromium",
+    "test:client:firefox": "vitest --run --project=client --browser=firefox",
+    "test:client:webkit": "vitest --run --project=client --browser=webkit",
+    "test:watch:client": "bun run test:watch --project=client --browser.headless=false",
+    "storybook": "storybook dev -p 6006 --no-open --host 0.0.0.0"
```
advl: 6007 for app

packages/svelte/ds-app/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Prerequisites
+
+- Svelte 5 or higher
+
+## Installation
+
+```bash
+bun add @canonical/svelte-ds-app
+```
```
advl, line 13: smoke test the install in an app with `npm link` or `bun link`
-> 1 in the dependency package (creates ln -s to the global register)
-> 2 in the consumer package

packages/svelte/ds-app/vite.config.ts
advl, line 1: check whether you can consume the shared vitest config

advl (changes requested):
Minor comments + 

To follow :
- Cleanup baseline drift + color (@advl)
- Add a table on the core readme about the storybook ports
- Summon package svelte

packages/svelte/ds-app/vite.config.ts
jmuzina: this should be a separate Svelte package, handled separately. Tracked [here](https://github.com/canonical/pragma/issues/836)

packages/svelte/ds-app/README.md
jmuzina: Verified, working properly

packages/svelte/ds-app/package.json
jmuzina: changed to 6007. Added a [follow-up](https://github.com/canonical/pragma/issues/837) to make sure the ports are unique across the repo and document the ports in README as [requested](https://github.com/canonical/pragma/pull/694#pullrequestreview-4705855661)

packages/svelte/ds-app/package.json
jmuzina: Not so much "wrong" as "unfortunate" :laughing: 

But yes, this is a consequence of `svelte-package` not giving us a way to filter out built files.

packages/svelte/ds-app/biome.json
jmuzina: This is slightly more nuanced: we have been by [convention in the monorepo ](https://github.com/search?q=repo%3Acanonical%2Fpragma%20build-cache&type=code)setting our sveltekit cache output directory to `.build-cache` in our svelte.config.js files. Not sure why we are doing this or where it started, my guess is it originated in the Launchpad package. Then we specifically need to ignore this directory in package Biome config.

So, this can't be centralized as-is as it is a combination of an opinionated svelte config choice and a resulting biome config. A Svelte-specific biome config would be slightly problematic because our current Biome config is framework-agnostic.

jmuzina: Tracked in https://github.com/canonical/pragma/issues/838

packages/svelte/ds-app/package.json
jmuzina: Ports cleanup: https://github.com/canonical/pragma/pull/845

advl (approved):

PR #695, opened 2026-06-25
feat(svelte-ds-app-launchpad): allow SSR-opened dialogs

steciuk:
## Done

Adds support for rendering `Modal`/`SidePanel` dialogs already-open during SSR (non-modal `open` attribute), then upgrading them to a true modal after hydration while keeping a two-way `bind:open` API in sync with user-driven state changes (invoker commands, Escape, outside click).

**Changes:**
- Introduces a bindable `open` prop (two-way sync) on `Modal`, and forwards it through `SidePanel`.
- Adjusts dialog/backdrop transition behavior so SSR-opened (non-modal) dialogs paint immediately, while modal-open animations still work.
- Updates stories and tests to cover controlled `open` behavior and SSR upgrade behavior.
- Removes `showModal` and `close` component instance methods

**Drive-bys**:
- removes the invoker commands' fallback (already [supported in all major browsers](https://caniuse.com/wf-invoker-commands))
- adds `.vitest-attachments` to `.gitignore`.

## QA

- `bun run check && bun run vitest --run`.
- Verify Storybook for [`Modal`](https://698d9cc0a0decf9afb7a90cb-hjrvlvzlch.chromatic.com/?path=/docs/components-modal--docs) and [`SidePanel`](https://698d9cc0a0decf9afb7a90cb-hjrvlvzlch.chromatic.com/?path=/docs/components-sidepanel--docs).
- Import and use the components in a local project that supports SSR.
  - Provide `open: true` prop to `Modal` and `SidePanel` components.
  - Navigate to the page and verify that the dialogs are visible on page load. Verify that after JS kicks in, the dialogs are modal (e.g. focus can't be moved outside the dialogs)
  - Disable JS in your browser and refresh the page. Verify that the dialogs are visible on page load.

### PR readiness check

### DIFF
--- a/packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
@@ -3,7 +3,7 @@
 <script lang="ts">
   import { useIsMounted } from "../../useFunctions/index.js";
   import { isEventTargetInElement } from "../../utils/index.js";
-  import type { ModalMethods, ModalProps } from "./types.js";
+  import type { ModalProps } from "./types.js";
   import "./styles.css";
 
   const componentCssClassName = "ds modal";
@@ -19,14 +19,6 @@
     open = $bindable(),
     ...rest
   }: ModalProps = $props();
-
-  export const showModal: ModalMethods["showModal"] = () => {
-    open = true;
-  };
-
-  export const close: ModalMethods["close"] = () => {
-    open = false;
-  };
 
   const fallbackId = $props.id();
   const id = $derived(idProp || fallbackId);
@@ -104,32 +96,25 @@
   {...rest}
 >
   <div style="display: contents;" bind:this={contentWrapperRef}>
-    {@render children?.(id, close)}
+    {@render children?.(id, () => (open = false))}
   </div>
 </dialog>
 
 <!-- @component
-`Modal` provides a mechanism for displaying content overlaying the main application. 
+`Modal` provides a mechanism for displaying content overlaying the main application.
 
-Modal can be imperatively controlled by the following methods available on the component instance:
-- `showModal`: Shows the modal.
-- `close`: Closes the modal.
+Modal is declaratively controlled by default through the [Invoker Commands API](https://developer.mozilla.org/en-US/docs/Web/API/Invoker_Commands_API) using `commandfor` and `command` attributes supplied via the `trigger` snippet. See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) for more information.
 
-Modal is declaratively controlled by default through the [Invoker Commands API](https://developer.mozilla.org/en-US/docs/Web/API/Invoker_Commands_API) using `commandfor` and `command` attributes supplied via the `trigger` snippet. Imperative methods remain available for cases where opening or closing must be orchestrated in code. See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) for more information.
-
-Modal can also be controlled through the bindable `open` prop. With `bind:open` it stays in sync with the modal in both directions: setting it opens or closes the modal, and it updates to reflect changes made by invoker commands, `Escape`, or an outside click. Setting `open` during SSR renders the dialog open on page load without client-side JS, and it is upgraded to a true modal once hydrated.
+For cases where opening or closing must be orchestrated in code, Modal can be controlled through the bindable `open` prop. With `bind:open` it stays in sync with the modal in both directions: setting it opens or closes the modal, and it updates to reflect changes made by invoker commands, `Escape`, or an outside click. Setting `open` during SSR renders the dialog open on page load without client-side JS, and it is upgraded to a true modal once hydrated.
 
 ## Example Usage
 ```svelte
 <script lang="ts">
-  let modal = $state<ModalMethods>();
-  // Imperative controls on the component instance
-  $effect(() => modal?.showModal())
   let open = $state(false);
 </script>
 
 <p>Modal is {open ? "open" : "closed"}</p>
-<Modal bind:this={modal} bind:open>
+<Modal bind:open>
   {#snippet trigger(triggerProps)}
     <button {...triggerProps}>
       Open Modal
--- a/packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
@@ -6,7 +6,9 @@
   import type { ModalProps } from "./types.js";
   import "./styles.css";
 
-  const componentCssClassName = "ds modal";
+  const componentCssClassNameBase = "modal";
+  const componentCssClassName = `ds ${componentCssClassNameBase}`;
+  const componentCssClassNameNonModalBackdrop = `ds ${componentCssClassNameBase}-non-modal-backdrop`;
 
   let {
     id: idProp,
@@ -63,7 +65,7 @@
 
 <!-- A non-modal dialog has no real `::backdrop` so we need to fake one. Can't be a pseudo-element, because a pseudo lives inside the dialog so `closeOnOutsideClick` wouldn't work. -->
 {#if initialOpen}
-  <div class={`${componentCssClassName}-non-modal-backdrop`}></div>
+  <div class={componentCssClassNameNonModalBackdrop}></div>
 {/if}
 <dialog
   {id}
--- a/packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
@@ -1,6 +1,7 @@
 <!-- @canonical/generator-ds 0.10.0-experimental.2 -->
 
 <script lang="ts">
+  import type { Attachment } from "svelte/attachments";
   import { useIsMounted } from "../../useFunctions/index.js";
   import { isEventTargetInElement } from "../../utils/index.js";
   import type { ModalMethods, ModalProps } from "./types.js";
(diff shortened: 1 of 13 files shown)


packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
```diff
@@ -41,41 +42,70 @@
...
+          dialogEl.close();
+        }
+        dialogEl.showModal();
+      } else {
+        dialogEl.close();
+      }
+
+      // Re-enable the transition after the first sync so that later changes animate.
+      dialogEl.classList.remove("no-transition");
+    });
```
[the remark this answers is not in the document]
steciuk: Not true. It's a no-op if dialog is already modal-opened.

packages/svelte/ds-app-launchpad/src/lib/components/SidePanel/SidePanel.svelte
```diff
@@ -22,7 +26,9 @@
 
 <Modal
   bind:this={modalMethods}
+  bind:open
   class={[componentCssClassName, className]}
+  --modal-backdrop-color="transparent"
   {...rest}
 />
```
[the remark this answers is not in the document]
steciuk: Not true: https://svelte.dev/docs/svelte/custom-properties

packages/svelte/ds-app-launchpad/src/lib/components/SidePanel/styles.css
```diff
@@ -21,7 +21,10 @@
     */
   &:is(:open, [open]) {
     transform: translateX(0);
+  }
 
+  /* Starting styles only for modal dialogs, so the SSR opened shows up immediately */
+  &:modal {
     @starting-style {
       transform: translateX(100%);
     }
```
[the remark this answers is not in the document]
steciuk: Again, not true.

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
```diff
@@ -15,24 +16,24 @@
...
   }: ModalProps = $props();
 
-  let dialogRef = $state<HTMLDialogElement>();
-
-  const fallbackId = $props.id();
-  const id = $derived(idProp || fallbackId);
-
-  const isMounted = useIsMounted();
-
   export const showModal: ModalMethods["showModal"] = () => {
```
goulinkh: can we replace `showModal` and `close` with a simple code path to toggle the state `open`?

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
```diff
@@ -41,41 +42,70 @@
...
       contentWrapperRef &&
       !isEventTargetInElement(e.target, contentWrapperRef)
     ) {
-      close();
+      open = false;
     }
   };
+
+  // Capture the initial value of `open` for SSR paint. After hydration, the dialog's open attribute should never be set manually.
+  const initialOpen = open;
```
goulinkh, line 50: BTW general question, what do you think about making comments of type `/** */` in order to get the comment as a description to the variable without the need to come back to this line to check the comment.

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
```diff
@@ -41,41 +42,70 @@
...
+
+  // Reflect the invoker commands, Escape, outside click changes back onto `open`.
+  const ontoggle: typeof ontoggleProp = (e) => {
+    ontoggleProp?.(e);
+    const newOpen = e.newState === "open";
+    if (newOpen !== open) open = newOpen;
```
goulinkh, line 56: What's the reasoning behind the check instead of direct assigning:
```
open = e.newState === "open";
```

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
```diff
@@ -41,41 +42,70 @@
...
+  };
+
+  const manageOpenState: Attachment<HTMLDialogElement> = (dialogEl) => {
+    // Suppress the transition on mount so that when we upgrade to modal the open fade doesn't play.
+    dialogEl.classList.add("no-transition");
```
goulinkh, line 79: Can this be flattened into the root component instead of using an attachment.
Also we could make `"no-transition"` a default class applied to the component, that get's removed inside the `$effect`?

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
```diff
@@ -41,41 +42,70 @@
...
 {@render trigger?.({
-  onclick: isInvokerCommandsFallbackNeeded ? showModal : undefined,
   commandfor: id,
   command: "show-modal",
   "aria-haspopup": "dialog",
 })}
 
+<!-- A non-modal dialog has no real `::backdrop` so we need to fake one. Can't be a pseudo-element, because a pseudo lives inside the dialog so `closeOnOutsideClick` wouldn't work. -->
+{#if initialOpen}
+  <div class={`${componentCssClassName}-non-modal-backdrop`}></div>
```
goulinkh: Can you edit `const componentCssClassName = "ds modal";` to `const componentCssClassName = "modal"` in order to safely append this postfix? otherwise, future changes may unintentionally break this class.

goulinkh (changes requested):

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
steciuk: We could. `showModal` / `close` are the least useful ones, true. The only value they bring is that the consumer can operate the modal imperatively without becoming a "state-holder". Do you think that it isn't worth keeping?

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
steciuk: > Can this be flattened into the root component instead of using an attachment.

You mean to have it inline? - Yes. Will do.

> Also we could make "no-transition" a default class applied to the component, that get's removed inside the $effect?

We cannot, as it will then stay applied if we never hydrate - meaning no transitions for command-controlled modals.

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
steciuk: > future changes may unintentionally break this class

What kind of changes? I've used this pattern already (e.g. [here](https://github.com/canonical/pragma/blob/main/packages/svelte/ds-app-launchpad/src/lib/components/Timeline/common/HiddenEvents/common/Link/Link.svelte)).

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
steciuk: It's so that when the bound `open` prop is changed from the outside, the value isn't echoed back.

1. `open` changes from outside
2. The attachment triggers dialog state change
3. `ontoggle` fires and performs an assignment
4. The consumer receives the same value they set via the binding

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
steciuk: Actually

> can operate the modal imperatively without becoming a "state-holder"

is not true, as you'll need to grab and hold a component reference to invoke them. I'll remove them.

steciuk: Removed :heavy_check_mark:

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
goulinkh: > > Can this be flattened into the root component instead of using an attachment.

> You mean to have it inline? - Yes. Will do.

No, I meant to have directly in the component without `@attach`:
```svelte
  let dialogEl = $state();

  // Map `open` changes to `showModal`/`close`. First run upgrades the dialog to modal if `open` is true.
    $effect(() => {
      if (open) {
        if (dialogEl.open && !dialogEl.matches(":modal")) {
          // `open === true` during SSR case
          // `showModal` throws when called on open non-modal dialog so we need to close it first.
          dialogEl.close();
        }
        dialogEl.showModal();
      } else {
        dialogEl.close();
      }

      // Re-enable the transition after the first sync so that later changes animate.
      dialogEl.classList.remove("no-transition");
    });

<dialog bind:this={dialogEl} .. />
```

goulinkh: if not possible, I think the previous version with the attachment moved next to the rest of the code was more readable.

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
goulinkh: The issue with this pattern is it's fragile, where any edit to the class could cause this to break:
```svelte
const componentCssClassName = "ds modal"
<div class={`${componentCssClassName}-something`} ../>
```

Where the following any of the following edits could break this without any warnings:
```diff
- const componentCssClassName = "ds modal"
+ const componentCssClassName = "modal ds"
+ const componentCssClassName = "ds modal "
+ const componentCssClassName = "ds base-modal"
<div class={`${componentCssClassName}-something`} ../>
```

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
steciuk: Oh, I misunderstood. I believe that the attachment is way cleaner then.
1. It doesn't require reactive dialog ref floating around - we don't grab it at all.
2. No `?.` before calling methods, as the element is always defined in the attachment's scope.
3. Natural ordering of the `no-transition` set and removal without `isFirstEffectRun` hacks or states being set from within the `$effect`.

It's exactly what we need here.

packages/svelte/ds-app-launchpad/src/lib/components/Modal/Modal.svelte
steciuk: So would you prefer something like:

```ts
  const componentCssClassNameBase = "modal";
  const componentCssClassName = `ds ${componentCssClassNameBase}`;
  const componentCssClassNameNonModalBackdrop = `ds ${componentCssClassNameBase}-non-modal-backdrop`;
  ```
  
  Or do you have any other idea on how to handle this?

goulinkh: yes much better.

goulinkh (approved):

PR #724, opened 2026-07-07
fix(ds-global-form): clear all selections on multiple-combobox reset

advl:
## Done

- Fix the multiple-select Combobox reset (✕): it cleared only the last selection, not all of them.
  - Root cause: the reset handler and the empty-list path emitted `onChange(undefined)`. With a react-hook-form `Controller` that has a `defaultValue`, setting the field to `undefined` reverts it to that default array on the next render; the `value → selectedItems` sync effect (which only ever populated, never cleared) then repopulated the chips.
  - Emit `[]` instead of `undefined` on both the reset and the empty-list paths so RHF stores the cleared state; make the sync effect clear as well as populate so an empty value reinforces the empty selection.
- Add a real regression test — the test file was previously `it.todo` stubs. Renders the RHF-bound `ComboboxField` with three default selections, clicks reset, and asserts **both** that no chips remain **and** that the submitted form value is `[]` (not the resurrected default). Verified by mutation: the value assertions fail against the pre-fix code.

Fixes the "combobox multiple ✕ only resets the last item" bug.

## QA

1. Open Storybook → `subcomponents/ComboboxInput` → the multiple-select story (or any `ComboboxField` with `isMultiple` bound to a form with default selections).
2. Select several options — they render as chips.
3. Click the ✕ reset button.
4. **Expected:** all chips clear (not just the last one), and the form value becomes `[]`.
5. Removing chips one by one down to zero also leaves the value at `[]` (no resurrection on the next render).

### PR readiness check

- [x] PR should have one of the following labels: `Bug 🐛`.
- [x] All packages define the required scripts in `package.json` (`check`, `check:fix`, `test`, `build`).
- [ ] N/A — no new package.
- [x] No visual change — `no visual change` label added (logic-only fix; no CSS/DOM structure change).

## Screenshots

N/A — behavioural fix, no visual change.

---

**Note on date/time Canonical icons:** this PR was originally scoped to also swap the browser-default date/time indicator icons for Canonical ones. That work is blocked upstream (no calendar/clock glyph exists in `@canonical/ds-assets`, and the swap is Chromium-only — Firefox has no equivalent pseudo-element), so it is split out into a tracking issue and not included here.

### DIFF
--- a/packages/react/ds-global-form/src/lib/subcomponent/ComboboxInput/ComboboxInput.tests.tsx
+++ b/packages/react/ds-global-form/src/lib/subcomponent/ComboboxInput/ComboboxInput.tests.tsx
@@ -1,12 +1,148 @@
-// ComboboxInput tests are blocked by ResetButton importing Button from
-// @canonical/react-ds-global, which fails to resolve in vitest.
-// TODO: fix once ds-global exports are resolvable in test environment.
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import type React from "react";
+import { useState } from "react";
+import { FormProvider, useForm } from "react-hook-form";
+import { describe, expect, it, vi } from "vitest";
+import { ComboboxField } from "#lib/component/ComboboxField/index.js";
+import type { Option } from "../types.js";
+import { ComboboxInput } from "./ComboboxInput.js";
 
-import { describe, it } from "vitest";
+const fruits: Option[] = [
+  { value: "apple", label: "Apple" },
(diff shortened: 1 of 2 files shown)


advl:
The date/time Canonical-icons work originally scoped alongside this fix is tracked separately in #725 (blocked upstream: no calendar/clock glyph in `@canonical/ds-assets`, and the swap is Chromium-only). This PR ships the combobox reset fix on its own.

— claude code

PR #739, opened 2026-07-07
feat(svelte-ds-global): Implement global Svelte breadcrumbs component

jmuzina:
## Done

Implements `Breadcrumbs` globally.

Fixes https://warthogs.atlassian.net/browse/WPE-426

Depends on #738 for Playwright migration, https://github.com/canonical/pragma/pull/757 for ds-type changes to `Item` interface identified as necessary during review.

## QA

- Open the [storybook](https://wpe-426-svelte-breadcrumbs--6a3ca65a785e76b915a40851.chromatic.com/?path=/docs/components-breadcrumbs--docs). 
- Ensure the component and its docs behave identically to the [react version](https://ds-react-global.canonical.com/?path=/docs/components-breadcrumbs--docs).
  - Note: component-level docs do not currently render a component description/usage guidelines etc. See [thread](https://chat.canonical.com/canonical/pl/4yie94r3epdczbnffq7a5x51ow).
  - There is no separate CustomItem story because it is impossible in Svelte to declare a separate component inside a TS file like it is in React (see [React example](https://github.com/canonical/pragma/blob/1cf47a20889f1f25208110550398990bc11067e5/packages/react/ds-global/src/lib/component/Breadcrumbs/Breadcrumbs.stories.tsx#L83-L125)). See details element below for more details if curious.

<details>
<summary>
Further explanation on the lack of separate CustomItem story
</summary>
From Claude:
<blockquote>
 Svelte can't inline a component like JSX can, and your package's "files" glob only excludes *.test.* and *.stories.* from the build. A bare CustomItem.svelte sitting in that folder would ship in dist/ as a real, unintended export (nothing in index.ts currently barrel-exports the folder, but it'd still be publishable/importable, unlike the test fixture which is filtered out).

Two reasonable ways to do it without that leak:

1. Reuse CustomItem.test.svelte directly in the story (import CustomItem from "./CustomItem.test.svelte"). Zero new files, but a .test.svelte import inside a .stories.svelte reads oddly and couples story content to test fixture naming/lifecycle.
2. Add CustomItem.stories.svelte as a dedicated story-only fixture (mirroring the .test.svelte convention but for stories), imported only by Breadcrumbs.stories.svelte. This gets excluded from dist for free via the existing !dist/**/*.stories.* glob, keeps test and story concerns separate, and matches the pattern your team already established for the test fixtures.
</blockquote>
Another option: We could also use some naming convention like `<ComponentName>.stories.fixture.svelte` to define these fixture components, and ignore them from publishing by naming convention.
</details>

### PR readiness check

## Screenshots

image

image


jmuzina:
Need to get https://github.com/canonical/pragma/pull/738 into an agreeable state, then update the tests here to use Playwright instead of jsdom and svelte testing library.

jmuzina:
Noticed an annoyance with Svelte storybook component docs not being rendered. See [thread](https://chat.canonical.com/canonical/pl/4yie94r3epdczbnffq7a5x51ow).

jmuzina:
I quite like the responsive behavior of the [Launchpad breadcrumbs](https://698d9cc0a0decf9afb7a90cb-yjwvcpcwfk.chromatic.com/?path=/story/components-breadcrumbs--collapsed-when-too-narrow), which intelligently create a box showing the leading breadcrumb items that could not fit. 

image

But, it seems to be expressly recommended against in the [Coda docs for Breadcrumbs](https://docs.superhuman.com/d/Pragma-Documentation-Database_dNyzE_TLZDh/per-Tier_suu2dwwx#UI-Blocks-per-Tier_tus7JmYw/r18&columnId=c-40YWiDKIJO&view=full):

> Ideally breadcrumbs should be used in the design in a way where they do not overflow. But if they do for some reason overflow they should overflow in the following way: A 4px vertical distance. The slash should always be at the start of the second line (and not at the end of the first line).

So, assuming this will remain the case, I suppose this implementation should remain in alignment with the React core implementation that does not have Launchpad's more advanced responsiveness behavior. CC @diana-stanciulescu @ninarojc1 @EnzoYD for thoughts.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/CustomLink.test.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+  import type { LinkComponentProps } from "./types.js";
+
+  let { href, class: className, children }: LinkComponentProps = $props();
+</script>
+
+<a {href} class={className} data-custom="true">
+  {@render children?.()}
+</a>
```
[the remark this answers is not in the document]
jmuzina: https://github.com/canonical/pragma/pull/739/commits/c63b3514f1bd1b82ca59d13c623196d0b18255e0

packages/svelte/ds-global/src/lib/component/Breadcrumbs/Breadcrumbs.svelte
```diff
@@ -0,0 +1,53 @@
...
+  class={[componentCssClassName, className]}
+  aria-label={ariaLabel}
+  {...rest}
+>
+  <!-- 
+    a11y: If we set `list-style: none` on the `ol` element, 
+    it will remove the default list semantics and the 
+    screen reader will not announce it as a list.
+  -->
+  <ol class="list p" role="list">
```
jmuzina: Deviation from React implementation: explicitly adding `list` role to fix an accessibility problem, described in the comment here.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/Item.svelte
```diff
@@ -0,0 +1,86 @@
...
+      This way, the disabled item is announced as a disabled link.
+    -->
+    <span
+      class="link"
+      aria-current={current ? "page" : undefined}
+      aria-disabled={disabled ? "true" : undefined}
+      role={disabled ? "link" : undefined}
+    >
+      {@render content()}
+    </span>
```
jmuzina, line 52: Deviation from React implementation for a11y fix:

Added `link` role and `aria-disabled` attribute to disabled items so that they are semantically communicated to AT as disabled elements.

jmuzina (commented):
Calling out two deviations from the React implementation to fix a11y issues I identified.

diana-stanciulescu:
> I quite like the responsive behavior of the [Launchpad breadcrumbs](https://698d9cc0a0decf9afb7a90cb-yjwvcpcwfk.chromatic.com/?path=/story/components-breadcrumbs--collapsed-when-too-narrow), which intelligently create a box showing the leading breadcrumb items that could not fit.
> 
@jmuzina @EnzoYD  we can consider upstreaming this functionality - elipsis for overflow was noted as a "next step" for breadcrumbs

packages/svelte/ds-global/src/lib/component/Breadcrumbs/Breadcrumbs.svelte
advl: If possible please include the react fix as a drive-by

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/Item.svelte
advl: Same thing here if possible :)

packages/svelte/ds-global/src/lib/component/Breadcrumbs/Breadcrumbs.svelte
steciuk: Is this actually the case?

The [MDN says](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/list-style#accessibility):
> Safari does not recognize ordered or unordered lists as lists in the accessibility tree if they have a list-style value of none, **unless the list is nested within the [`<nav>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/nav)** navigation element.

and we are indeed inside of a `nav` here.

To confirm, I tested the behavior locally (chromium, firefox and webkit) and  `list-style: none` (with no explicit role) does not seem to strip the list semantics for here.

EnzoYD:
Happy to upstream it. Should I create an issue or is there already a place for this topic?

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/Item.svelte
```diff
@@ -0,0 +1,86 @@
...
+
+  let {
+    children,
+    label,
+    url,
+    current = false,
+    separator = "/",
+    disabled = false,
+    class: className,
+    className: itemClassName,
```
steciuk: Could we omit `className` from `Item` to not have both `class` and `className` here?

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/Item.svelte
```diff
@@ -0,0 +1,86 @@
...
+    LinkComponent = "a",
+    // Destructure Item props we don't spread to DOM
+    key: _key,
+    items: _items,
+    displayItemsType: _displayItemsType,
+    Component: _Component,
```
steciuk: `key` is already destructured from `Item` in `Breadcrumbs` ([here](https://github.com/canonical/pragma/pull/739/changes#diff-1daaa388f813841d865c337ebeccfa43a30568b9d8986ab7ad7c6a500d465f96R31)). Could the other unused ones follow the same pattern and be omitted from the `ItemProps` type?

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
```diff
@@ -0,0 +1,53 @@
...
+import type { LinkComponent, LinkComponentProps } from "./common/Item/types.js";
+
+/**
+ * Breadcrumb-specific item extending navigation Item (WD405)
+ *
+ * Adds breadcrumb-specific properties while maintaining
+ * compatibility with the unified navigation type.
+ */
+// biome-ignore lint/suspicious/noExplicitAny: Component accepts any props
+export interface BreadcrumbItem extends Item<Component<any>> {
```
steciuk: Shouldn't `item.Component` be forced to accept at least the props that we provide to it (not any)? Otherwise, the consumer that uses a custom component is not forced to explicitly handle them.  The props we provide will be present in the component's props at runtime but invisible to the type, risking unexpected behaviors, e.g., when handling `...rest` spread.

A bit unfortunate effect of this would be that the custom component itself would need to be logically tied to our `Breadcrumbs` implementation (e.g. by extending an interface exported from here).

Hence, I'd like to propose an alternative pattern for providing such overrides that ensures strict type safety without sacrificing custom overrides' flexibility: typed snippets instead of a `Component`.

See example: https://svelte.dev/playground/952c2e66b1e844ce90be4df69b7821a5

advl (approved):
Code LGTM
Asking only to make the react fixes as drive-by

Thanks

steciuk (dismissed):

packages/svelte/ds-global/src/lib/types/link.ts
```diff
@@ -0,0 +1,27 @@
+import type { Component, Snippet } from "svelte";
+import type { HTMLAttributes } from "svelte/elements";
+
+/**
+ * Props passed to a custom `LinkComponent` for router integration.
+ *
+ * The single, shared contract for every component that injects one link
+ * renderer for its navigable items (Breadcrumbs, SideNavigation, Tabs, …).
```
alnvdl-work: Nitpick: ellipsis character is usually not used in code/comments.

Maybe just use three dots as is more common `...`?

packages/svelte/ds-global/src/lib/component/Breadcrumbs/Breadcrumbs.svelte
```diff
@@ -0,0 +1,53 @@
...
+      <!-- Each item is spread onto the component -->
+      <ItemComponent {separator} {LinkComponent} {...itemProps} />
+    {/each}
+  </ol>
+</nav>
+
+<!-- @component
+Breadcrumbs are a navigational aid that shows users their current location
+within a site's information hierarchy. They provide a clear path from the
+root of the information hierarchy (IA) to the current page, allowing users
```
alnvdl-work: Is `IA` correct for information hierarchy? Perhaps this is information architecture?

alnvdl-work (approved):
LGTM with the other comments addressed. Just two nitpicks.

diana-stanciulescu:
for "information hierarchy" - I updated coda with "information architecture". I think this automatically will sync to the description you mentioned. for the ellipsis, @advl please take a look.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/Item.svelte
goulinkh: @jmuzina @steciuk Can we agree on a svelte code standard rule to avoid this in the future?

```ttl
cs:svelte.component.class-prop a cs:CodeStandard ;
  cs:hasCategory cs:svelte ;
    cs:name "Native class prop" ;
    cs:description """Accept consumer CSS classes through Svelte's native `class` prop only. Do not also expose a React-style `className` prop: Svelte forwards `class` natively, so a parallel `className` duplicates the same input, is ambiguous when both are set, and leaks a React-ism into Svelte components.""" ;
    cs:do [
        cs:description """
(Do) Destructure the native `class` prop and merge it last in the class array.
```svelte
<script lang="ts">
  const componentCssClassName = "ds breadcrumbs-item";
  let { current = false, disabled = false, class: className, ...rest }: ItemProps = $props();
</script>

<li class={[componentCssClassName, current && "current", disabled && "disabled", className]} {...rest}></li>
```

packages/svelte/ds-global/src/lib/component/Breadcrumbs/CustomItem.test.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
```
goulinkh: Fixtures files should be grouped inside a folder called `fixtured`, if you are ok with this let's also write a rule for this WDYT? @steciuk @jmuzina @advl 
FYI @advl in Svelte you cannot have inside the tests custom templates so the fixtures `.svelte` files are common when complex snippet are required (react equivalent of children):
```ttl
cs:testing.svelte-component-fixtures a cs:CodeStandard ;
    cs:hasCategory cs:testing ;
    cs:name "Svelte component test fixtures" ;
    cs:extends cs:testing.unit.fixtures_pattern ;
    cs:description """Test fixtures that must be real Svelte components (a custom `Component`/`LinkComponent` replacing a default, for example) cannot be `testing/fixtures.ts` named exports because Svelte components compile only from `.svelte` files. Place such fixtures in a `fixtures/` folder colocated with the test that consumes them, and name them `Name.fixture.svelte` — never with a `.test.` segment, which falsely implies a runnable test and collides with the real test suffix. A fixture contains no `describe`/`it`; it only renders markup for a test to mount.""" ;
    cs:do [
        cs:description """
(Do) Colocate the component fixture in a `fixtures/` folder with a `.fixture.svelte` name.
```

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/Item.svelte
jmuzina: The reason both class symbols are here is because, like the React breadcrumbs component, the [ds-types Item interface](https://github.com/canonical/pragma/blob/47b3be9fde607b86ae88ef9e842b455aca2a3cf1/packages/ds-types/src/navigation/types.ts#L41-L45) that we are extending here uses `className`, which to be honest is rather React-specific. `className` isn't necessary in Svelte, native `class` works out of the box.

I could, for now, `Omit` `className` here to clean this up for this component - but I feel a change may be necessary to `ds-types` to sort this out. Svelte shouldn't have to accommodate a React API like this.  Any thoughts group + @advl ?

goulinkh: Zooming out, I don't think the issue is `className` specifically, it's that we're centralizing **component props** at all. A component's props are its framework's public API, and you can only put in a shared props type what *every* consumer can express. That floor is React, which is the least expressive of the frameworks we target, one opaque `children`, one-way data, callback events. So a shared props type quietly caps Svelte at React's capabilities: no named/parameterized snippets, no `{@attach}`, no `bind:` on props, native `class` swapped for `className`. Several of those are things our own `cs:svelte.*` standards actually require, so the shared type is in direct tension with our Svelte conventions.

So the line I'd draw: `ds-types` owns **framework-agnostic data models and narrow contracts**, the item data model, tokens, enums, the `LinkComponentProps` shape and never component Props.
Each framework package owns its props and composes the shared data model with its own snippets, bindings, and attributes. Pulling `className` out of `Item` isn't the fix, since this is a more fondamental flow in the current approach.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
```diff
@@ -0,0 +1,53 @@
...
+   * @default "/"
+   */
+  separator?: Snippet | string;
+  /**
+   * Custom link component for router integration
+   * e.g. a SvelteKit-aware link wrapper
+   * Applied to all items unless overridden per-item
+   * @default "a"
+   */
+  LinkComponent?: LinkComponent;
```
goulinkh: Can we discuss customizability here? `LinkComponent` injection is a React pattern, its only composition primitive is passing a component/render-prop. Svelte gives us native `class`, snippets, attachments and `bind:`, so we inject markup and behavior directly instead of swapping a component type through a narrowed contract (note `LinkComponentProps` strips the anchor attributes down to `href/className/children/aria-current`, a ceiling we wouldn't impose in Svelte). Sorry to sound like a Svelte cult member…

Our own `ds-app-launchpad` Breadcrumbs already does this the Svelte way: no `LinkComponent`, each `Segment` extends the native element type (`SvelteHTMLElements["span"]` / `LinkProps`), so consumers spread `class`, `aria-*`, `data-sveltekit-*`, and handlers straight onto the element. That covers most SvelteKit routing on its own, so the escape hatch mostly isn't needed.

Svelte-native options, least to most reach:

1.  Native attribute spread per item (app-launchpad's approach), `cs:svelte.component.*`.
2.  Subcomponents like `Breadcrumbs.Item`, `cs:svelte.composition.subcomponents`.
3.  A named `item`/`link` snippet with a spreadable props object, `cs:svelte.composition.snippets-vs-subcomponents` + `snippet-attributes`.
4.  `{@attach}` for behavior (preloading, active-link detection), `cs:svelte.composition.attachment-naming`.
5.  `setContext` for cross-cutting injection, `cs:svelte.composition.reactivity-boundaries`.

I'd prefer aligning with the app-launchpad approach over carrying the React injection pattern across.

goulinkh (changes requested):

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/Item.svelte
advl: @goulinkh / @jmuzina it sound good to me, both for the code standard and the fix to the types. Please tag me on the PRs. Thanks !

packages/svelte/ds-global/src/lib/component/Breadcrumbs/CustomItem.test.svelte
advl: @advl Ok let's do this. We need to precise at which level the fixture files should live, esp if they can be shared between files. I think /src/testing/fixtures if global or just ./fixtures if local

steciuk: In `ds-app-launchpad` we used to name the fixture files `test.fixture.svelte` (no name of the component in the file name), living flat in the component directory. 

See examples: https://github.com/search?q=repo%3Acanonical%2Fpragma%20path%3Atest.fixtures.svelte&type=code

The fixtures usually contain exported snippets, and we haven't had a need (yet?) for more than one such fixture per developed component, which makes me believe that if we were to have a dedicated `fixtures` directory, in the vast majority of cases it would contain one file only (bloating file tree unnecessarily?).

I'm not against it and just wanted to flag it here so you are aware of it when deciding on the pattern to standardize :)

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
steciuk: Should we establish a coding standard that suggests using typed snippets rather than components when overrides are needed, following the approach proposed [here](https://github.com/canonical/pragma/pull/739#discussion_r3550749872)?

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
jademathre-canonical: A breadcrumb item must carry a `url` or a `key`. Breadcrumbs keys its `{#each}` on `getItemId(item)`, which can error at runtime for items with neither. Requiring them can turn that runtime crash into a type error with something like:
```
export type BreadcrumbItem = BreadcrumbItemBase &
  ({ url: string } | { key: string });
```

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/Item.svelte
jmuzina: PR created, and this PR has been rebased on top of it.

https://github.com/canonical/pragma/pull/757

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
jmuzina: Thanks for this, I have incorporated it in the [upstream item props PR](https://github.com/canonical/pragma/pull/757) (which simplifies the Item interface and, in the description, shows this pattern as a recommendation for usage in Svelte) and here.

jmuzina: @advl wdyt here? I'm wondering if the `Item` interface should enforce that one of the keys needed for `getItemId` to not throw an error should be guaranteed present by evolving its Typing somewhat - that way this error Jade mentions is not possible at runtime.  

We could solve this at the component level as Jade suggests here but I wonder if it would be helpful to solve it for all components at the level of the `Item` interface.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
jmuzina: I agree, this deserves a coding standard. We can author one as a followup if this is approved.

I've implemented a change that just extends the anchor element rather than accepting a custom `LinkComponent` as I don't think Breadcrumbs needs the escape hatch of a typed snippet currently.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/CustomItem.test.svelte
jmuzina: Per our other requested changes in this branch to directly extend anchor instead of passing a custom component in, this branch no longer has any need for test fixture files. 

Still I created an [issue](https://github.com/canonical/pragma/pull/739#discussion_r3552158191) to track creating this standard.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
jmuzina: Tracked in https://github.com/canonical/web-code-standards/issues/19

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/Item.svelte
jmuzina: I have [updated](https://github.com/canonical/pragma/pull/739/commits/1f9fb7987f77c75dc7a9bb8b152ec0a7b657e58f) the props to discard `items` from `ItemProps`. The other discarded props now no longer exist thanks to refactorings applied by other threads.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/Breadcrumbs.svelte
jmuzina: yup you're right, we were fine here since we are using a `nav` element. Reverted this change

packages/svelte/ds-global/src/lib/component/Breadcrumbs/common/Item/types.ts
```diff
@@ -0,0 +1,41 @@
...
+import type { SvelteHTMLElements } from "svelte/elements";
+
+/**
+ * Native anchor attributes, forwarded to the rendered link (or the
+ * `<span>` standing in for it when current/disabled).
+ * `href`, `class`, and `children` are modeled explicitly instead.
+ */
+export type ItemAnchorProps = Omit<
+  SvelteHTMLElements["a"],
+  "href" | "class" | "children"
```
steciuk, line 12: Is the omit of `class: ClassValue` here and re-add in `ItemProps` as `string` intentional here?

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
```diff
@@ -0,0 +1,58 @@
...
+   * When true, renders as text instead of link.
+   */
+  current?: boolean;
+  /** CSS class applied to this item's `<li>`, in addition to the base classes. */
+  class?: string;
+  /**
+   * Custom renderer for this item, replacing the default `Breadcrumbs.Item`.
+   * e.g. A specialized snippet for complex items.
+   */
+  render?: Snippet<[BreadcrumbItem]>;
```
steciuk: I wonder if this `render` wouldn't be more "ergonomic" on the `Breadcrumbs` root instead of on individual `BreadcrumbItem`s? 🤔

```ts
export interface BreadcrumbsProps extends BaseProps {
  items: ItemProps[]; // subcomponent props directly
  render?: Snippet<[ItemProps]>;
  // ...
}
```

If no `render` provided, each item of `items` is spread on `<Item />`, otherwise, it is handled back to the consumer via the Snippet, so they specify the override to be used only once on `<Breadcrumbs />`.

If there's a need for custom rendering per segment, it could be handled inside the `render` snippet definition E.g.:

```svelte
<Breadcrumbs
  items={[
    { url: "/", label: "Home" },
    { url: "/products", label: "Products" },
    { key: "details", label: "Product Details", current: true },
  ]}
>
  {#snippet render({ current, ...itemProps })}
    <!-- Whatever logic the consumer requires -->
    {#if current}
      <MySpecialCurrentItem {current} {...itemProps} />
    {:else}
      <MySpecialItem {...itemProps} />
    {/if}
  {/snippet}
</Breadcrumbs>
```

WDYT?

jmuzina: I like this, it seems to follow Svelte idioms well. Adjusted in https://github.com/canonical/pragma/pull/739/commits/49853e9729f4051b1b779c57b6508b050774e0da

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
advl: Remove getItemId and use a type check instead of the runtime invariant.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/Breadcrumbs.stories.svelte
```diff
@@ -0,0 +1,76 @@
...
+<Story
+  name="WithCustomAriaLabel"
+  args={{
+    "aria-label": "You are here",
+    items: [
+      { url: "/", label: "Home" },
+      { key: "settings", label: "Settings", current: true },
+    ],
+  }}
+/>
```
goulinkh, line 120: Please consider including more stories such as how to customize items rendered element

packages/svelte/ds-global/src/lib/component/Breadcrumbs/Breadcrumbs.svelte
```diff
@@ -0,0 +1,48 @@
...
+Breadcrumbs are a navigational aid that shows users their current location
+within a site's information architecture. They provide a clear path from the
+root of the information architecture (IA) to the current page, allowing users
+to quickly understand where they are and easily navigate back to previous
+levels in the IA. Breadcrumbs should reflect the IA, not the user's path
+they took to arrive at the current page. They work best in websites or
+applications with multiple levels of hierarchy and are particularly useful
+when users might arrive at deep pages through search or external links.
+
+`import { Breadcrumbs } from "@canonical/svelte-ds-global";`
```
goulinkh, line 45: Consider including `@example` simple usage.

packages/svelte/ds-global/src/lib/component/Breadcrumbs/index.ts
```diff
@@ -0,0 +1,16 @@
+import { default as BreadcrumbsRoot } from "./Breadcrumbs.svelte";
+import { Item } from "./common/index.js";
+
+const Breadcrumbs = BreadcrumbsRoot as typeof BreadcrumbsRoot & {
+  /**
+   * Default breadcrumb item renderer, with link and separator.
+   * Can be replaced per-item by providing a custom `render` snippet
+   * on the item.
```
goulinkh, line 8: Consider including `@example` simple usage

packages/svelte/ds-global/package.json
```diff
@@ -15,6 +15,7 @@
   "files": [
     "dist",
     "!dist/**/*.test.*",
+    "!dist/**/*.fixture.*",
```
goulinkh:
```suggestion
    "!dist/**/*.fixtures.*",
```

See: https://github.com/canonical/web-code-standards/pull/20/changes#diff-64bf15f0ee76845fb34a2117a74c319b051e02595ae3c1f6c505f9ec005719aeR610

goulinkh (approved):
LGTM, thanks for the great work, left some tiny last comments

packages/svelte/ds-global/src/lib/component/Breadcrumbs/types.ts
jmuzina: Doing this as a follow-up, it shouldn't change the code at this callsite.

jmuzina:
Merging after https://github.com/canonical/pragma/pull/846 !

PR #754, opened 2026-07-10
chore(Svelte): boilerplate Svelte ds-global package follow-up

jademathre-canonical:
## Done

Follow-up items related to changes on #693

- Moved types array inside compilerOptions; at the top level TypeScript silently ignores it.
- Added moduleResolution: nodenext in the shared config replaces per-package workarounds for svelte-check's broken inference.
- Rewrote the caveat as a feature note; the workaround is now built into the config rather than the consumer's job.
- Bumped @canonical/styles, @canonical/biome-config, @canonical/svelte-ssr-test to ^0.29.0; ^0.29.0 resolves to the workspace packages.
- Deleleted packages/svelte/ds-global/src/index.ts, it was a redundant re-export shim, nothing references it (build input is src/lib, exports point to dist/).
- Removed vitest.config.ts from excludes; the file doesn't exist.
- Used RenderResult type; simpler than the ReturnType<typeof render<...>> gymnastics.
- retitled Welcome.mdx  to Documentation/Introduction; matches the react ds-global Storybook hierarchy.
- chromatic.ds-svelte-app-launchpad.yml, chromatic.ds-svelte-app-wpe.yml: fixed "Uncommment" typos and clarified which package each TODO refers to.

### PR readiness check


configs/typescript-svelte/README.md
```diff
@@ -22,9 +22,4 @@ This configuration enables the following behavior:
...
+4. `moduleResolution: nodenext`, declared explicitly. The base configuration's [`"module": "NodeNext"`](https://www.npmjs.com/package/@canonical/typescript-config) normally implies this, but `svelte-check` (used by the `check:ts` script in Svelte packages) does not always apply the inference correctly and can then fail to resolve packages that rely on Node's `exports` field (e.g. `vite`, `@storybook/addon-svelte-csf`). Declaring it explicitly here sidesteps the inference, so consuming packages don't need their own override.
```
advl, line 25: @jmuzina cc - is this accurate ?  I feel declaring it in the config wouldnt help either - it should be declared on the consumer package directly

advl:
One minor comment, lgtm aside from that

packages/svelte/ds-global/src/lib/component/Example/Example.svelte.test.ts
jmuzina, line 1: I'd recommend not to make changes that assume `@testing-library` usage, as we are moving towards using Playwright in #738 . maybe the testing changes can be made there or in a branch targeting it

packages/svelte/ds-global/tsconfig.json
```diff
@@ -2,7 +2,13 @@
...
   "compilerOptions": {
     "skipLibCheck": true,
-    "moduleResolution": "nodenext"
+    "types": [
+      "node",
+      "svelte",
+      "vitest/globals",
+      "vite/client",
+      "@testing-library/jest-dom"
+    ]
```
jmuzina: Yup, I noticed I put this in the wrong place too 😆 I'm  applying it too in #738 , fine to do it here as well. Maybe just remove the `@testing-library` type per my other comment

configs/typescript-svelte/README.md
jmuzina: Yes, this is correct. Jade's change is that we explicitly recommend always setting module resolution in a consuming package, rather than optimistically setting it in the base tsconfig and suggesting an override as a mitigation if issues are found.

Declaring it in the config has benefit and downside. Declaring it here more strongly suggests to consumers what config they should be using downstream, even though they have to declare it in the package. However, it's true that due to the stated `svelte-check` issues the consuming package has to separately declare it. 

Still, I think it's best to include it in the base tsconfig: by doing so, downstream projects that don't override module resolution will see type checking issues *only* from `svelte-check`, not other tools they might be using that also rely on tsconfig and don't have this inference bug. This suggests to the dev more strongly that the issue has something to do with `svelte-check` which might lead them more strongly to find this documentation.

configs/typescript-svelte/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "@canonical/typescript-config",
   "compilerOptions": {
     "types": ["svelte", "vite/client"],
-    "verbatimModuleSyntax": true
+    "verbatimModuleSyntax": true,
+    "moduleResolution": "nodenext"
```
advl, line 6: does this still need to be repeated ? see https://github.com/canonical/pragma/blob/a78d05f5be234539624de0b278bb987fab1a5edc/configs/typescript/tsconfig.json

jademathre-canonical: Yes, base config never declares `moduleResolution` explicitly, and svelte-check doesn't apply the inference. Removing this line breaks check:ts.

jmuzina (approved):

steciuk (approved):

PR #757, opened 2026-07-10
fix(ds-types): remove React-styled className from shared Item type

jmuzina:
Refactors the `Item` interface to focus more strongly on its role as a Navigation utility adapter. Properties with no relation to navigation have been removed. This has the added benefit of removing framework-specific properties like `className`, which is a React idiom.

Other updates:
- Updates `utils` package tests and documentation to no longer reference `Item` as a generic.
- Updates `react-ds-core` components that extend `Item` to separately declare `className` and `Component` on their props to apply framework customization at a more appropriate level of abstraction.

##### React
```ts
// types.ts
import type { Item } from "@canonical/ds-types";
import type { ComponentType } from "react";

export interface BreadcrumbItem extends Item {
  current?: boolean;
  className?: string;
  Component?: ComponentType<ItemProps>;
}
```
```tsx
// Breadcrumbs.tsx
{items.map((item) => {
  const ItemComponent = item.Component ?? Item;
  return <ItemComponent key={getItemId(item)} {...item} />;
})}
```

##### Svelte
Optionally apply `class` (rather than `className`), a snippet renderer (in place of a React component type), and Breadctumb-specific behavior like `current`. 

```ts
// types.ts
import type { Item } from "@canonical/ds-types";
import type { Snippet } from "svelte";

export interface BreadcrumbItem extends Item {
  current?: boolean;
  class?: string;
  render?: Snippet<[BreadcrumbItem]>; // Svelte's override idiom: swap the markup
}
```

```svelte
<!-- Breadcrumbs.svelte -->
{#each items as item (getItemId(item))}
  {#if item.render}{@render item.render(item)}{:else}<Item {...item} />{/if}
{/each}
```

### DIFF
--- a/packages/ds-types/src/navigation/types.ts
+++ b/packages/ds-types/src/navigation/types.ts
@@ -39,12 +39,6 @@ export interface Item<C = DefaultComponent> {
   disabled?: boolean;
 
   /**
-   * CSS class name for custom styling of the item.
-   * e.g. 'nav-item-active' to highlight the current page.
-   */
-  className?: string;
-
-  /**
    * Array of child items for nested navigation structures.
    * e.g. Submenu items under a parent like 'Settings'.
    */
--- a/packages/ds-types/src/navigation/types.ts
+++ b/packages/ds-types/src/navigation/types.ts
@@ -1,19 +1,15 @@
 /**
- * Default component type for navigation items.
- * Framework-agnostic — consumers supply a concrete type
- * (e.g. React.ComponentType) via the generic parameter.
- */
-export type DefaultComponent = unknown;
-
-/**
  * Navigation item - Public API (WD405)
  *
  * A unified type for navigation items across the design system.
  * Used by Breadcrumbs, SiteNavigation, FileTree, MegaMenu, etc.
  *
- * @template C - Custom component type for rendering the item
+ * Only fields consumed by the shared navigation utilities belong here.
+ * Styling and rendering customization (CSS classes, custom renderers)
+ * belong on the item types that components declare by extending this
+ * one (e.g. BreadcrumbItem, MenuItem), using each framework's own idioms.
  */
-export interface Item<C = DefaultComponent> {
+export interface Item {
   /**
    * Unique identifier for the item when a URL is not provided.
    * e.g. 'section-header' for a non-navigable grouping item.
@@ -42,19 +38,13 @@ export interface Item<C = DefaultComponent> {
    * Array of child items for nested navigation structures.
    * e.g. Submenu items under a parent like 'Settings'.
    */
-  items?: Item<C>[];
+  items?: Item[];
 
   /**
    * Enum value determining the type of display for items.
    * e.g. 'list' for rendering as a simple list.
    */
   displayItemsType?: "default" | "custom";
-
-  /**
-   * Custom component for rendering the item itself.
-   * e.g. A specialized renderer for complex items.
-   */
-  Component?: C;
 }
 
 /**
@@ -65,8 +55,7 @@ export interface Item<C = DefaultComponent> {
  *
  * Generic over the **whole item type** `T` it annotates, so an enhanced item
  * (extra fields beyond the base WD405 `Item`, e.g. `icon`/`slot`) survives
- * annotation with its fields intact and fully typed. `T` may itself carry a
- * custom component type (`Item<C>`).
+ * annotation with its fields intact and fully typed.
  *
  * @template T - The item type being annotated (defaults to the base `Item`)
  */
--- a/packages/ds-types/src/navigation/types.ts
+++ b/packages/ds-types/src/navigation/types.ts
@@ -39,12 +39,6 @@ export interface Item {
    * e.g. Submenu items under a parent like 'Settings'.
    */
   items?: Item[];
-
-  /**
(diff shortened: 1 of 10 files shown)


packages/ds-types/src/navigation/types.ts
```diff
@@ -1,19 +1,15 @@
...
  * Used by Breadcrumbs, SiteNavigation, FileTree, MegaMenu, etc.
  *
- * @template C - Custom component type for rendering the item
+ * Only fields consumed by the shared navigation utilities belong here.
+ * Styling and rendering customization (CSS classes, custom renderers)
+ * belong on the item types that components declare by extending this
+ * one (e.g. BreadcrumbItem, MenuItem), using each framework's own idioms.
  */
-export interface Item<C = DefaultComponent> {
+export interface Item {
```
jmuzina, line 12: This was intended to be framework-agnostic originally, and it was in the literal sense (`unknown` puts the responsibility of mapping type for item customization up to the framework and component). But, I don't believe it conceptually belongs here.

This was framework-agnostic in the technical sense, but not in the API design intent. The pattern of customizing an item by by passing a Component type directly via props is a composition pattern that works well for React using `ComponentType` or similar but not for Svelte when it would likely be more idiomatic to use a Snippet. 

Also, the core value-add of this type seems to be to create a bridge to the shared navigation helpers in `utils/src/lib/navigation`. `Component` was not actually consumed anywhere in those helpers, nor was `className` - it was up to the framework-specific component author to decide how to use them. As such I propose to remove this generic to keep rendering/composition logic framework-specific, and more strongly align this interface with its use as an adapter for the Navigation utils.

packages/utils/src/lib/navigation/annotateTree.test.ts
```diff
@@ -74,15 +74,18 @@ describe("annotateTree", () => {
...
-      className: "active",
+      icon: "home",
       disabled: true,
     };
     const result = annotateTree(root);
 
-    expect(result.className).toBe("active");
+    expect(result.icon).toBe("home");
     expect(result.disabled).toBe(true);
   });
```
jmuzina, line 88: This change looks unrelated at first glance, but the reason it is here is to keep the spirit of the test (item properties being persisted after tree annotation) intact, but without referencing `className` (which is React-specific), instead referencing a more realistic and general case like adding an `icon` identifier by extending the item interface.

packages/react/ds-global/src/lib/component/Breadcrumbs/types.ts
```diff
@@ -13,13 +13,20 @@ import type { ItemProps, LinkComponentProps } from "./common/Item/types.js";
...
   current?: boolean;
+  /** CSS class name applied to this item's `<li>`, in addition to the base classes. */
+  className?: string;
+  /**
+   * Custom component for rendering this item itself.
+   * e.g. A specialized renderer for complex items.
+   */
+  // biome-ignore lint/suspicious/noExplicitAny: Component accepts any props
+  Component?: ComponentType<any>;
 }
```
jmuzina, line 30: These `className` and `Component` additional properties are now the responsibility of the individual component that extends the Item interface - in this case React. 

We could create framework-specific interfaces that extend `Item` and add more appropriate properties here,  but I think it would distract from the purpose of `Item`, as stated in the PR description. `Item` is not about rendering, it is about navigation utility.

packages/react/ds-global/src/lib/component/Tabs/types.ts
```diff
@@ -8,7 +8,10 @@ import type { LinkComponent } from "../../types/link.js";
  * `label` (visible text), `key` (identity when there is no url), and
  * `disabled`. Nested `items` are ignored (tabs cannot nest).
  */
-export type TabItem = Item;
+export interface TabItem extends Item {
+  /** CSS class name applied to this tab's `<li>`, in addition to the base classes. */
+  className?: string;
+}
```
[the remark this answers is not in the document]
jmuzina: This issue predates this PR. Leaving it as not in scope of this, core DS team or next product team to encounter this issue should solve it.

CC @advl for awareness. The issue is that, even in the main branch, subitems are not type-restricted to the same extension of `Item` that their parents are.

packages/ds-types/src/navigation/types.ts
```diff
@@ -1,19 +1,15 @@
...
  *
  * A unified type for navigation items across the design system.
  * Used by Breadcrumbs, SiteNavigation, FileTree, MegaMenu, etc.
  *
- * @template C - Custom component type for rendering the item
+ * Only fields consumed by the shared navigation utilities belong here.
+ * Styling and rendering customization (CSS classes, custom renderers)
+ * belong on the item types that components declare by extending this
+ * one (e.g. BreadcrumbItem, MenuItem), using each framework's own idioms.
  */
```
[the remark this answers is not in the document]
jmuzina: Good point. removing: https://github.com/canonical/pragma/pull/757/commits/9429bf25c887cfa4bbfdfdb0b304d315b72dafc2

packages/react/ds-global/src/lib/component/Breadcrumbs/types.ts
advl: Fair enough

advl (approved):
lgtm, thanks !

PR #840, opened 2026-07-16
chore(packages): adopt webarchitect where the ruleset asserts something

### LINKED ISSUE #286: Improvements to the webarchitect tool
add `webarchitect --list` to display available schemas
consume `webarchitect` in all the packages of the monorepo
consider renaming webarchitect to `tsarchtiect` or `nodearchitect` as it's not web specific

advl:
## Done

Adopts `webarchitect` in the **eight** packages whose kind has invariants worth asserting.

| Package | Ruleset | What it now checks |
|---|---|---|
| `cli/summon` | `tool-ts` | biome config, package structure, license |
| `react/ssr-adapter-{cloudflare,deno,vercel}` | `library` | biome config, package structure, license |
| `storybook/addon-{form-state,msw,utils}` | `library` | as above |
| `svelte/ssr-test` | `library` | as above |

Each gains the `check:webarchitect` script **and** the wiring into its `check` chain, so the ruleset actually runs.

Violations the adoption surfaced and fixed: `cli/summon` had no `types` and pointed `module` at the compiled binary — both now resolve to a new types-only barrel at `src/index.ts`, deliberately not re-exporting `bin.tsx`, since importing that file runs the CLI. `addon-form-state` and `addon-msw` were missing `module` and had no `test` script at all, so the root fan-out skipped them silently; they now declare an explicit no-op, following the `ds-types` precedent. `addon-utils` was missing `module`.

## What this PR deliberately does NOT do

An earlier revision also adopted `webarchitect` in `styles/{main,debug,typography}` and `ds-implementations` under the `assets` ruleset. **That has been dropped.**

`assets` extends `base`, which declares a `license` variable and no rules at all. `assets` adds exactly one: that `license` equals `LGPL-3.0`. All four packages already declared `LGPL-3.0` before the change, so the rule had no failing case on the day it was added and no plausible one after.

What those four would have gained is a `check:webarchitect` script, a devDependency, a lockfile entry and a CI task, in exchange for one string comparison. And the cost is not merely the ceremony: a reader seeing `check:webarchitect` in a manifest reasonably concludes the package's structure is validated. It would not have been. That converts *unverified* into *appears verified*, which is the more expensive way to be wrong.

This repository already holds the principle — `pr.yml` states that *"a guard whose failing case no longer fails is worse than no guard"*, and runs `check:ranges:test` before `check:ranges` precisely to prove the guard can fail.

**The follow-up that belongs to those four is a question about the ruleset, not the packages:** either give `assets` invariants that are true of asset packages and could actually fail — `files` covers what ships, `main`/`exports` resolve to something that exists, a LICENSE file is present, no stray build script — or delete the ruleset. A ruleset that exists to be adopted rather than to check should not exist.

Adoption is therefore reported as *every package whose kind has invariants worth asserting*, not as a fraction. A coverage figure is a metric that rewards adopting a vacuous ruleset.

## QA

```bash
bun run check         # Successfully ran target check for 62 projects — exit 0
bun run check:ranges  # 62 workspace packages, every sibling range satisfied
```

Root `bun run test` was not obtained: it exits 130 under load on this machine, naming a different project set each run. CI's build matrix is the verdict of record.

### PR readiness check

- [x] PR should have one of the following labels — `chore` is derived from the title automatically.
- [x] All packages define the required scripts in `package.json` — `check`, `check:fix` and `test` verified present in all eight; `addon-form-state` and `addon-msw` gain explicit no-op `test` scripts rather than leaving the root fan-out to skip them silently.
- [x] If this PR introduces a **new package** — n/a.
- [x] If this PR does not require visual testing, add the `Chromatic: skip` label — applied; manifests and one types-only barrel.

## Screenshots

n/a — no visual surface.


packages/cli/summon/package.json
```diff
@@ -18,6 +19,7 @@
     }
   },
   "files": [
+    "src",
```
advl: why src ? Seems it isnt a good pattern

advl: Good instinct to question it — here's the honest picture.

**Why it's here:** `"src"` isn't a free choice in this PR. The `tool-ts` ruleset mandates it as a hard invariant:

```jsonc
// tool-ts.ruleset.json
"module": { "const": "src/index.ts" },
"types":  { "const": "src/index.ts" },
"files":  { "type": "array", "contains": { "const": "src" } }
```

`cli/pragma` already satisfies exactly this (`module`/`types` → `src/index.ts`, `files: ["src","dist"]`). This PR's whole job is to make `summon` conform to the same ruleset, so dropping `src` here would just fail `check:webarchitect` — the opposite of the goal.

**What the pattern is for:** top-level `module`/`types` → raw `src/index.ts` lets in-repo consumers and TS resolve types straight from source with no build step (dev/typecheck ergonomics), and shipping `src` in the tarball gives published source maps a target.

**Where your smell is right:** `summon` also has an `exports` map, and modern (exports-aware) resolvers — Node ESM, bundlers, TS `bundler`/`nodenext` — read `exports` and *ignore* top-level `module`/`types`. So for anyone resolving through `exports`, those two fields are inert, and they point at a *different* entry (the `src/index.ts` barrel that re-exports the UI) than `exports["."]` (the compiled `bin`). Only legacy resolvers that read top-level `module`/`types` see the `src` path at all. Plus shipping `src` roughly doubles the published tarball for source nothing on the published side references. So yes — as a published-package entry-point strategy it reads inconsistent.

**But that's a ruleset-level question, not this PR's.** The inconsistency is inherited from `tool-ts` and already lives in `cli/pragma`; changing it means changing the ruleset so both tools move together, which is out of scope for "adopt webarchitect in the remaining 11 packages." My recommendation is to keep it as-is here (conform, match pragma) and, if we want to revisit the source-entry pattern, do it as a dedicated `tool-ts` change. Happy to file that issue against `tool-ts` (drop `module`/`types` → `src` and `files: src` for tool packages that publish via `exports`→`dist`) so it's tracked rather than lost — say the word.

---

advl: Acted on this — the `src`-entry pattern is now deprecated rather than propagated. Pushed three commits to this branch:

1. **`7b6c58f`** — rewrote the `tool-ts` ruleset to require **compiled `dist` entry points**: `files` must contain `dist`; `module`/`types`, *if present*, must match `^dist/` (so a `src` entry is now a validation error, not a requirement); `build` + `check:ts` required. Conformed `cli/pragma` and `cli/summon` — dropped the top-level `module`/`types` (nothing imports either as a library; consumers use `bin`/`exports`→`dist`) and set `files: ["dist"]`. So the `"src"` you flagged is gone.
2. **`0c45352`** — the one remaining `tool-ts` consumer was `webarchitect` itself, which was run-from-TS. Compiled it: real `tsc` build to `dist/esm` + `dist/types`, `bin`/`module`/`types` → `dist`, `files: ["dist","rulesets"]`. Its ruleset discovery already had a `dist/esm` fallback, so bundled rulesets still resolve. One CI wiring add: a post-build `bun install --ignore-scripts` re-link, because bun's isolated linker won't create the workspace `.bin` symlink until the compiled bin file exists.
3. **`9395501`** — README Developer Tooling note documenting the convention (TS CLI tools compiled & published from `dist`; run-from-source deprecated).

Verified: the **built** webarchitect binary resolves rulesets and validates all three tools (`Summary: All 3 validations passed`), pragma/summon still build and run, and root `bun run check` is green across all 61 projects. `tool-ts` now means "compiled TS tool" for pragma, summon, and webarchitect alike.

---

.github/actions/setup-env/action.yml
```diff
@@ -100,3 +100,10 @@ runs:
     - name: Build packages
       shell: bash
       run: ${{ inputs.build-command }}
+
+      # Bun only links a workspace bin once its target file exists. Compiled CLI
+      # bins (e.g. @canonical/webarchitect's dist/esm/cli.js) don't exist until
+      # the build above has run, so re-link once more after building.
+    - name: Re-link compiled workspace binaries
+      shell: bash
+      run: bun install --ignore-scripts
```
jmuzina, line 118: Might be worth an extra comment here that this needs to happen AFTER the git diff against bun.lock, otherwise somebody not being careful could move CI steps around and cause stale lockfiles to not be caught by CI

jmuzina (approved):

packages/cli/summon/package.json
advl: Rename tool-ts to tool ?

.github/actions/setup-env/action.yml
advl: Good call — added in `ddf4322d`. Put an `ORDER-CRITICAL` note on the post-build "Re-link compiled workspace binaries" step spelling out that it (and every re-link `bun install`) must stay **after** the `git diff --exit-code bun.lock` verification, since a re-link install can rewrite `bun.lock` in place and would otherwise silently repair — and mask — a stale committed lockfile. Also added a shorter back-reference on the first "Re-link workspace binaries" step so a reorder of either is guarded.

---

packages/cli/summon/package.json
advl: Done in `fd7bea61`. Since the bare `tool` ruleset had zero consumers, this was a clean consolidation:

- Deleted the dead `tool.ruleset.json` (was just `package` + GPL) and renamed `tool-ts.ruleset.json` → `tool.ruleset.json` (`name: "tool"`), keeping the compiled-`dist` structure.
- Repointed every consumer (`webarchitect tool`): `cli/pragma`, `cli/summon`, `webarchitect` itself, and the `ADDING_A_PACKAGE.md` example.
- Renamed the concept end-to-end so nothing is half-named: the summon package **type** (`--type=tool`), `getRuleset`/`getEntryPoints`/`getLicense`, the MCP spec enum, docs (both READMEs, CONVENTIONS.md), and tests. `git grep tool-ts` is now empty.

Verified: webarchitect 87 tests + check pass (resolves `tool` + `base`), pragma/summon checks pass, root `bun run check` green across 61 projects.

One pre-existing mismatch I deliberately left (it predates the rename): the summon `tool` generator emits a *src-based, no-build* package, but the `tool` ruleset requires compiled `dist` — so a freshly-generated `tool` package would fail `webarchitect tool`. Reconciling that is a behavioral change to the generator (and its tests), so it belongs in its own follow-up rather than this rename. Want me to file it?

---

PR #846, opened 2026-07-16
fix(storybook/addon-utils): framework-agnostic withUtilStyles — stop wrapping stories in React elements

### LINKED ISSUE #839: `withUtilStyles` decorator crashes every non-React story render
## What's happening

Every Svelte story in Storybook, in every Svelte package, currently fails to render with:

```
TypeError: decorator_Component is not a function
      in <unknown>
      in DecoratorHandler.svelte
      in DecoratorHandler.svelte
      in PreviewRender.svelte
```

React storybooks do not experience this issue, it seems to be isolated to non-React.

## Root cause

`withUtilStyles` (`packages/storybook/addon-utils/src/withUtilStyles.tsx`) is registered globally for every framework via `createConfig.ts`.

https://github.com/canonical/pragma/pull/807#discussion_r3592130016 (merged yesterday) changed `withUtilStyles` to actually wrap the story in a React element created by React's `createElement`.

Svelte's storybook plugin cannot render Svelte components inside React elements, causing this error.

advl:
## Done

- Makes `withUtilStyles` **completely framework-agnostic** again — the react import is gone entirely (`withUtilStyles.tsx` → `.ts`). The decorator is registered for every renderer (react, svelte, lit) via `@canonical/storybook-config`, so it must never wrap the story in any framework's element: #807's `createElement("div", …, StoryFn())` wrapper crashed every Svelte story with `decorator_Component is not a function`, and Lit failed the same way.
- The decorator now returns `StoryFn()` untouched and applies the computed classes + inline layout styles to the story's **real mount container, synchronously at decorate time**:
  - canvas: `#storybook-root` — the story's direct parent;
  - autodocs: `#story--<id>[--primary]-inner`, the per-story element `renderStoryToElement` mounts into (also the direct parent).
- Both targets exist while the renderer is composing the story, so the two properties the React wrapper existed for are **preserved**:
  - **no first-paint flash** — classes are in place before the story's first paint (the pre-#807 post-paint `useEffect` made stories flash their un-gridded layout, then reflow);
  - **subgrid binding** — classes land on the story's *direct* parent in both views, so a story root using `grid-template-*: subgrid` (the Cards group) binds to the decorator grid on the canvas AND the docs page.
- Containers that persist across stories (`#storybook-root`) are fully reset on each application (every producible class and managed style prop removed first), so no story inherits a previous story's modifiers.
- One deliberate behavioral nuance: in canvas **showcase** mode the layout styles now sit on `#storybook-root` (100% tall via `previewHead`), so the showcased item centres in the full preview height rather than in a 60vh band — docs behavior is unchanged (`min-block-size: 60vh` still applies there).

Fixes #839

## QA

Verified live on all three renderers with the built addon (`bun run build` in addon-utils, then `storybook dev`, driven with Playwright):

- **svelte/ds-global** (previously: every story crashed): canvas renders (`comfortable app` on `#storybook-root`), toolbar globals apply (`grid responsive dense app` + `align-content: start`), autodocs renders all stories with classes on each `#story--…-inner` (including the `--primary` duplicate). Zero page errors.
- **react/ds-global** (regression check on #807's fix): Cards subgrid story — `#storybook-root` gets `grid responsive comfortable app` with 12 computed columns; the story root is its direct child and its computed `grid-template-columns` resolves to `subgrid`; all autodocs story containers carry the grid classes with `display: grid`. Zero page errors.
- **lit/ds-prototype** (also broken before, per issue comment): canvas renders with classes applied. Zero page errors.
- `bun run check` (biome + tsc) and `bun run test` green in `packages/storybook/addon-utils`.

### PR readiness check

- [x] Label: `Bug 🐛` (suggested).
- [x] The code follows the code standards.
- [x] All packages define the required scripts (no script changes).
- [x] No new package introduced.
- [ ] Visual testing IS relevant here (the decorator drives story layout) — do **not** add `no visual change`; let Chromatic run.

## Screenshots

Not applicable — the fix restores rendering; see QA notes for per-renderer verification.

https://claude.ai/code/session_016WrkZXTopfNP1ogcpSJRKi

---
_Generated by [Claude Code](https://claude.ai/code/session_016WrkZXTopfNP1ogcpSJRKi)_


packages/storybook/addon-utils/src/withUtilStyles.ts
```diff
@@ -89,70 +157,62 @@ export const withUtilStyles = (
...
+    containerStyles = { "align-content": "start" };
   }
 
-  return createElement(
-    "div",
-    { className: wrapperClass || undefined, style: wrapperStyle },
-    StoryFn(),
-  );
+  for (const el of storyContainers(context)) {
+    for (const cls of CONTAINER_CLASSES) el.classList.remove(cls);
```
jmuzina: nit: `remove()` accepts a list of classes to remove, I think you could remove the loop. Haven't tried this locally though.

```suggestion
    el.classList.remove(...CONTAINER_CLASSES)
```

jmuzina (approved):

packages/storybook/addon-utils/src/withUtilStyles.ts
advl: Done in b290e04 — `classList.remove()` is indeed variadic, so the loop is gone and it now mirrors the spread used for `add()`. Checks and tests green.

---

PR #859, opened 2026-07-16
feat(svelte-ds-global): Upstream `SkipLink` from WPE tier to Global tier

jmuzina:
## Done

- Upstreams the SkipLink implemented in the WPE tier in #659 to the global tier.
- Fixes broken spacing styling in the React implementation, as a drive-by (see [thread](https://github.com/canonical/pragma/pull/859#discussion_r3845492023))

## QA

- Open [Svelte SkipLink](https://6a3ca65a785e76b915a40851-zetjlzyqds.chromatic.com/?path=/docs/work-in-progress-pattern-skiplink--docs&globals=scheme:light) and [React SkipLink](https://67be30a0b85c4e44b6744158-xcjunktyye.chromatic.com/?path=/docs/work-in-progress-pattern-skiplink--docs&globals=scheme:light).
- Verify the documentation makes sense and the component functions as you would expect.
- Open [hidden focused story](https://6a3ca65a785e76b915a40851-zetjlzyqds.chromatic.com/?path=/story/work-in-progress-pattern-skiplink--focused&globals=scheme:light) (used for visual testing). Verify that it focuses the SkipLink on opening. You should see a [passing interaction in the bottom of the screen](https://github.com/user-attachments/assets/49156a49-a1da-48fd-be04-483e10954ed9)

### PR readiness check


.gitignore
```diff
@@ -1,5 +1,6 @@
 # Test screenshots
 __screenshots__/
+.vitest-attachments/
```
jmuzina, line 3: this is a drive-by. When I moved the component over I noticed it brought this folder with it, which needed to be gitignored. It is generated by Vitest when tests fail.

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.svelte.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  it("is keyboard focusable", () => {
+    const page = render(Component);
+    const skipLink = componentLocator(page).element();
+
+    skipLink.focus();
+
+    expect(document.activeElement).toBe(skipLink);
+  });
+});
```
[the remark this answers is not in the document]
jmuzina: restored in https://github.com/canonical/pragma/pull/859/commits/95f219d2b75b9643a35dd0715bedc68d5eeb015e

goulinkh:
@jmuzina Is this PR ready for review? I am not able to find a working storybook preview.

jmuzina:
Hey thanks @goulinkh , back from vacation now, I thought this was ready for review but I will update to fix the storybook.

jmuzina:
Seems like there was something wrong with the CI before, rebased on `main` and I can see the stories now.

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.stories.svelte
```diff
@@ -30,11 +30,13 @@
 	<SkipLink />
 	<main id="main" tabindex="-1">
 		<p>
-			Click inside this example box, then press Tab to focus the skip
-			link.
+			Click inside this example box, then press "Shift+Tab" to make the
+			skip link focused and visible.
```
cadencehagenauer-canonical, line 46: This isn't an issue exclusive to this pr, but the text here is rendering as white on white for me. 

Image

This issue also exists in the [React global strorybook](https://ds-react-global.canonical.com/?path=/docs/work-in-progress-pattern-skiplink--docs), so it could just be a wider storybook issue.

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/styles.css
cadencehagenauer-canonical, line 1: I haven't quite been able to pinpoint why yet, but there are some strange style discrepancies between this version and the React version, namely with the size of the background.

The padding numbers line up, but for some reason in the Svelte version the padding appears as more than 0px

React:

Image

Svelte:

Image

cadencehagenauer-canonical, line 1: Another strange thing is that the svelte skip link does not seem to use the dark background color, even though in the inspector --color-background-layer2 appears to be set to the dark color.

Image

.gitignore
cadencehagenauer-canonical: Good change, I had been noticing this issue as well

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.ssr.test.ts
```diff
@@ -28,7 +28,9 @@ describe("SkipLink SSR", () => {
     it("renders custom children when provided", () => {
       const page = render(Component, {
         props: {
-          children: createRawSnippet(() => ({ render: () => "Go to main" })),
+          children: createRawSnippet(() => ({
+            render: () => "<span>Go to main</span>",
```
cadencehagenauer-canonical, line 32: I'm curious about the motivation to make this change so I can apply that knowledge to my future work.

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/styles.css
cadencehagenauer-canonical, line 1: I'm also curious about the decisions behind using positioning rather than the clipping I usually associate with visually hidden styling.

cadencehagenauer-canonical (commented):
Component code looks good and in line with React. I am noticing some discrepancies in behavior between the two frameworks related to styling. I also have some questions about certain decisions for learning's sake.

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.stories.svelte
jmuzina: Yeah, we've noticed this before, I think it's something to do with the [themes addon](https://github.com/canonical/pragma/tree/main/packages/storybook/addon-canonical-shell-theme), @advl are you aware of this? 

I've been getting around it by opening example stories instead of the docs pages that embed them, it only seems to happen in embedded examples in docs view

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/styles.css
jmuzina: Good find, the React implementation is using tokens that no longer exist, it seems when those tokens were removed the implementation's tokens were not adjusted and thus the styles broke.  The Svelte implementation is using newer tokens, both versions should look like the Svelte screenshot you showed. I'm going to apply this fix to the React styles here.

jmuzina: Fixed in https://github.com/canonical/pragma/pull/859/commits/cbca4ef3a901ab5c5ea0cf75c44f294a513ca483

jmuzina: I learned something today, thanks for putting me down this rabbit hole!

Reading some articles:
- https://dbushell.com/2026/02/20/visually-hidden/
- https://webaim.org/techniques/css/invisiblecontent/

It seems like using clipping might be the more idiomatic / modern way to do this. We should consider adopting it. That being said I won't apply it in this PR so as to keep it scope limited to publishing a working implementation that aligns with the React version. We can explore changing to use clipping afterwards :)

cadencehagenauer-canonical: Sounds good! I had seen the webaim article before, but not the other one with the history of visually hidden!

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.stories.svelte
cadencehagenauer-canonical: Gotcha, good to know there's a workaround. This is probably best fit for another pr but will leave it open in case Allan has anything to say on it.

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/styles.css
cadencehagenauer-canonical: This was also fixed in https://github.com/canonical/pragma/commit/cbca4ef3a901ab5c5ea0cf75c44f294a513ca483

cadencehagenauer-canonical (approved):

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.stories.svelte
jmuzina: That's Adrian, not Allan :laughing: this is a recurring issue due to their tragically similar github usernames

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.ssr.test.ts
jmuzina: I did this because if you invoke `createRawSnippet` on a Svelte client without including a wrapping element (creating a text node), the Svelte client throws an error (see [src](https://github.com/sveltejs/svelte/blob/b433f3cd878814d6d141be359e5dafa9959b5f9e/packages/svelte/src/internal/client/dom/blocks/snippet.js#L89-L91)). This test code is run on the server, which doesn't throw an error when a text node is passed - but I think it's helpful to not create examples of calls to a function named the same that would fail if run on the client. I fear it would give authors the wrong idea and be hard to trace

cadencehagenauer-canonical (approved):

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.stories.svelte
goulinkh: Hi, responding to this thread since I have some knowledge on this as I have suggested the Canonical Storybook Shell Theme linked above.

The Storybook Shell Theme currently only has light theme, meaning it doesn't turn into dark mode or system mode.

This maybe confusing but I will try to explain with a drawing.

There are two theme systems running inside Storybook, the shell them (only light mode) and content theme which supports light/dark theme using Pragma CSS tokens and defaults to system mode (your OS choice of light/dark).

The issue that is shared above happens when the content theme is set to either `(system or dark)` AND `the shell being always on light` AND `the content rendered component storybook has no background`.
image

The content theme is controlled by the user via this bar:
image


**Workaround**

1. The reason this issue exists in the first place is Storybook itself has bad support for the shell to switch from dark to light or vise versa. it works well as static dark only or light only.
  - I do have a working concept of dynamic system theme for the shell theme it can be seen here but it will require an upstreaming effort to Pragma https://svelte-icons.canonical.com/
2. The easiest solution would be change the content theme default setting from current value `system` to `light` and let the user to change it manually to dark.

packages/react/ds-global/src/lib/_work_in_progress/SkipLink/styles.css
```diff
@@ -6,12 +6,12 @@
...
   --skip-link-outline-offset: calc(
     -1 *
-    var(--tmp-focusable-focused-outline-width)
+    var(--dimension-stroke-thickness-large, 3px)
   );
-  --skip-link-focus-inset-inline-start: var(--spacing-horizontal-small);
-  --skip-link-focus-padding-block: var(--spacing-vertical-large);
-  --skip-link-focus-padding-inline: var(--spacing-horizontal-large);
-  --skip-link-focus-top: var(--spacing-vertical-small);
+  --skip-link-focus-inset-inline-start: var(--space-100, 0.5rem);
```
goulinkh: any reason for the default values (`0.5rem`, `1rem`..), since the tokens are defined?

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.stories.svelte
```diff
@@ -2,10 +2,12 @@
 
 <script lang="ts" module>
   import { defineMeta } from "@storybook/addon-svelte-csf";
+  import { expect } from "storybook/test";
+  import type { ComponentProps } from "svelte";
   import SkipLink from "./SkipLink.svelte";
 
   const { Story } = defineMeta({
-    title: "Components/SkipLink",
+    title: "_work_in_progress/pattern/SkipLink",
```
goulinkh, line 10: I am not familiar with this pattern, is there a standard defined somewhere around this? or maybe a short reply clarifying the reasoning would also work.

goulinkh (changes requested):
Left a couple of small changes.

The hiding mechanism `absolute/top:-999px` and showing mechanism `z-index: 99999..` looks messy but I am not sure if there is a better way to approach this with CSS.

If that can be improved that would be great but I wouldn't block the PR on it, let me know once my comments are resolved and I will approve.

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.ssr.test.ts
cadencehagenauer-canonical: Good to know, thanks!

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.stories.svelte
jmuzina: This follows #719 by @advl - some UI blocks have different maturities at the core DS level. The React implementation of SkipLink is marked as work in progress. I'm not sure myself about what decision qualifies a UI block as "work in progress", but the Storybook config changelog seems to indicate it means it is "not-yet-tiered". It also details the other parts of this title hierarchy:

https://github.com/canonical/pragma/blob/e8d25fecbe6e70edfc4cb300bad65f691dbbd26e/configs/storybook/CHANGELOG.md?plain=1#L83-L86

packages/react/ds-global/src/lib/_work_in_progress/SkipLink/styles.css
jmuzina: This was following the style used in the React implementation, which seems to be an outdated Pragma convention to explicitly specify default literals. 

UI blocks added more recently by @advl seem to bind to the token without using a default, so I have removed the default from both Svelte & React implementations: https://github.com/canonical/pragma/pull/859/commits/58437fd6715a947bdc6f56a39a14d70bb4cd8e8e

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.stories.svelte
jmuzina: Thanks @goulinkh . I filed issue #962 to track this

jmuzina:
@goulinkh 

> The hiding mechanism absolute/top:-999px and showing mechanism z-index: 99999.. looks messy but I am not sure if there is a better way to approach this with CSS. If that can be improved that would be great but I wouldn't block the PR on it

Yeah, @cadencehagenauer-canonical and I discussed this [here](https://github.com/canonical/pragma/pull/859#discussion_r3845802149). We think the showing/hiding mechanism can be improved, but I suggest we do that separately as it is a larger change scoped broader than an upstreaming (it changes opinions in the React implementation as well)

packages/svelte/ds-global/src/lib/_work_in_progress/SkipLink/SkipLink.stories.svelte
goulinkh: thanks for the clarification!

goulinkh (approved):

advl:
Merged to avoid the pr becoming stale

PR #880, opened 2026-07-20
fix(router): Outlet constructs elements; routes declare components (AV-340)

advl:
## Done

- **The permanent fix**: `router-react`'s `Outlet` now constructs elements instead of delegating to `router.render()` — `createElement(route.component ?? route.content, {params, search})` folded through `createElement(wrapper.component, null, children)`. Route content and wrappers become **real fibers with their own hook lists**. Null semantics mirror core `render()` exactly; the `useSyncExternalStore` subscription and route-keyed `Suspense` are byte-identical to before.
- **`component:` field** — routes declare components, not render functions. `content:` still works, now `@deprecated`; `route()` enforces exactly-one (redirect routes exempt).
- **Typed route hooks** — `useRouteParams(route)` / `useRouteSearch(route)`, inferring params/search from the typed route table (type-witness contract, documented; stable snapshot identity so navigation-state publishes cause no re-render storms).
- **Per-route `fallback` and `errorComponent`** — composed by `Outlet` inside its route-keyed Suspense (private class boundary; resets on route change).
- **Consumers migrated** (drive-by, same effort): boilerplate-vite routes/decorators/tests, summon-application templates + generator emission + regenerated embedded templates, ds-app / ds-global / addon-utils story utilities. Zero `content:` remains in any non-test `route()` call on main.
- **Docs swept monorepo-wide** (drive-by, same effort): `ROUTER_API.md`, `MIGRATE_TO_PRAGMA_ROUTER.md`, `ROUTER_MIDDLEWARE_COOKBOOK.md`, both package READMEs (new surface + 3-line migration note), summon `CONVENTIONS.md`/`README.md`, the `Outlet` docstring. The deprecated form survives only where it documents the deprecation. CHANGELOGs untouched (historical records); router test files keep `content:` deliberately to cover the still-supported legacy path.

### The bug this fixes

`router-core`'s `render()` invokes a route's `content({params, search})` and every `wrapper.component({children})` as **plain function calls**, and `Outlet` ran that inside its own render — so any hooks those functions call attached to **Outlet's fiber**. Because a function component's signature is type-identical to the content contract, `content: PageComponent` compiled cleanly and worked on SSR and single-URL hydration, then crashed the **first client-side navigation** between pages with different hook counts:

```
Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
```

Found by manual testing on the pragma docsite. Every bare-component route wiring on main — the boilerplate, the summon templates — was latently broken the same way; no suite anywhere navigated client-side, which is why it survived. That test class now exists at both the package and app level.

Fixes AV-340.

## QA

1. `cd packages/react/router && bun run test` → 46 tests / 16 files. The pin is `Outlet hook ownership (AV-340) › navigates between bare-component routes with different hook counts`; it fails against the pre-fix Outlet with the verbatim error above (verified twice, independently, via React 19's `onUncaughtError` root option — `console.error` spies do not catch this).
2. `cd packages/runtime/router && bun run test` → 191 tests / 27 files (includes the exactly-one-slot validation cases and `render()`'s `component ?? content` fallback).
3. `cd apps/react/boilerplate-vite && bun run dev` → click between routes in the running app (previously the latent crash path); `bun run test` → 34 tests / 9 files.
4. Consumers: `summon-application` 39/3, `cli/pragma` 935/110 (+8 perf), `ds-app` 50/18, `ds-global` 349/79 (+5 skipped), `addon-utils` 5/2.
5. Workspace `bun run check` → green across 62 projects.

Reviewed by an adversarial pass before opening: **SHIP, zero HIGH/MED** — null-parity diffed branch-by-branch against core; `fallback: null` confirmed a meaningful no-fallback override; error-boundary reset verified via the keyed subtree; the exactly-one validation confirmed *not* a compat regression (content-less routes were already a type error on main); no re-render storms (match reference stable across navigation-state publishes). Its three doc-level LOWs are fixed in the final commit.

### PR readiness check

- [x] If this PR introduces a **new package**: n/a — no new packages.

## Screenshots

n/a — no visual change. The behavioural change is navigation-time hook attribution; it is covered by the package pin test and by the docsite's `clientNavigation.tests.tsx` integration twin.


advl:
Closing — **the crash this fixes already shipped; what remains is a separate API change that has to be rewritten rather than rebased.**

**The defect is gone from `main`.** This PR's core fix was that `Outlet` invoked `route.content(...)` and `wrapper.component(...)` as plain function calls inside its own render body, so route and wrapper hooks attached to Outlet's fiber and the first navigation between routes with different hook counts threw *"Rendered fewer hooks than expected"*.

`main` now does exactly what this PR proposed — `Outlet.tsx:58-88` builds the element with `createElement(match.route.content, { params, search })` and folds it through `reduceRight` over the wrappers. The regression test landed with it: `Outlet.test.tsx:102` is `describe("Outlet hook ownership (AV-340)")`, including the navigation-between-different-hook-counts case. That arrived via #1073.

**What has not landed** is the rest of this PR: the `component:` route field (deprecating `content:`), `useRouteParams` / `useRouteSearch`, `RouteErrorBoundary`, and per-route `fallback` / `errorComponent`. None of those exist on `main` — `types.ts` still declares `content` as the required and only UI slot.

**Why rewrite rather than rebase.** 32 of this branch's 42 files have been touched on `main` since the merge base, and the 9 that collide directly are all the substantive ones — `Outlet.tsx`, `Outlet.test.tsx`, `route.ts`, `types.ts`, `createRouter.ts`, `ROUTER_API.md` and both READMEs. #1063 rewrote `route.ts` into a type-level validation machine and reworked `types.ts` substantially, so this PR's "exactly one of content/component" check and its type additions have to be re-expressed in that new idiom, not merged into it. Resolving nine substantive conflicts to arrive at code that is already ~80% on `main` is the expensive way to get nothing.

**The work is not cancelled, only the branch.** AV-340 is closed; its successor tracks the remainder, which is now precisely: tighten the React-side route type so a bare component is the idiomatic form, plus the hooks and the error/fallback surface.

**For whoever re-cuts it:** the `content:` → `component:` rename is a consumer-wide breaking sweep and deserves its own PR, kept apart from `errorComponent`/`fallback` and from the typed route hooks. Three small PRs against current `main` will land faster and review better than one 42-file revival.

PR #881, opened 2026-07-20
fix(pragma-cli): restore oxigraph WASM embed in the compiled binary

advl:
## The bug
A clean install of `@canonical/pragma-cli` (a `bun --compile` standalone binary) fails on any machine other than the CI runner it was built on:

```
Error: Internal error: Failed to load oxigraph. Make sure it is installed:
ENOENT … /node_modules/oxigraph/node_bg.wasm
```

Bun's `--compile` does not embed the WASM that oxigraph loads via `readFileSync(`${__dirname}/node_bg.wasm`)` — `__dirname` is frozen to the build-time path, so the file isn't found at runtime.

## Root cause: a regression
This was fixed once (**#584**, `embedWasm.ts`), but the module was **dropped in the CLI rename/rewrite (#879)** while its compile-smoke test (`kernel/runtime/wasmEmbed.test.ts`) survived **orphaned** — so the binary regressed and that protected test has been red.

## Fix
- Restore `kernel/runtime/wasmEmbed.ts`: embed the WASM via `import("oxigraph/node_bg.wasm", { with: { type: "file" } })` (bun's file-embed → `$bunfs`), then patch CJS `readFileSync` to return those bytes for `node_bg.wasm`.
- Wire it as a **side-effect import at the top of `graphpack/read.ts`** — the single funnel every store-boot path (normal load *and* `__store-probe`) reaches before `createStore`, so it runs before oxigraph loads and never touches the storeless hot paths (no cold-start cost).

## Verification
A fresh `bun build --compile` binary boots the store from an **isolated cwd**:
```
$ pragma __store-probe
{"ok":true,"entities":6,"triples":"23"}
```
Confirmed the **embedded** copy is used by moving the on-disk WASM away and re-running (still `ok:true`). The orphaned `wasmEmbed.test.ts` guards this end-to-end in CI (it compiles a binary and boots the store).

Typecheck clean. Separate PR (unrelated to the DS-component work).

### DIFF
--- a/packages/cli/pragma/src/kernel/runtime/graphpack/read.ts
+++ b/packages/cli/pragma/src/kernel/runtime/graphpack/read.ts
@@ -12,6 +12,12 @@
  * fast path.
  */
 
+// Side-effect import — MUST precede the `@canonical/ke` import below so the
+// oxigraph WASM is embedded / its `readFileSync` patched before ke's
+// `createStore` (→ `loadOxigraph`) runs. `read.ts` is the single funnel every
+// store-boot path reaches (normal load and the __store-probe), and it is only
+// dynamic-imported on store boot, so this keeps the storeless hot paths clean.
+import "../wasmEmbed.js";
 import { readFileSync } from "node:fs";
 import { join } from "node:path";
 import { createStore } from "@canonical/ke";
--- a/packages/cli/pragma/src/kernel/runtime/wasmEmbed.ts
+++ b/packages/cli/pragma/src/kernel/runtime/wasmEmbed.ts
@@ -1,0 +1,57 @@
+/**
+ * Embed oxigraph WASM for `bun build --compile`.
+ *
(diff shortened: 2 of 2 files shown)


advl:
Closing — the premise no longer holds. This restores the WASM embed *in the compiled binary*, and the package no longer ships one.

`src/kernel/runtime/wasmEmbed.test.ts` on `main` records the change in its own docblock:

> It used to compile a `bun build --compile` binary and spawn that, because the shipped artifact WAS that binary and the question was whether the WASM survived embedding. The package now ships emitted JavaScript, so the same guard points at the same place it always did — what a consumer runs — which is `node dist/src/bin.js`.

So the process-boundary guard this PR was protecting still exists and still runs; it just points at the artifact consumers actually get. `wasmEmbed.ts` itself is absent from `main` because there is nothing left to embed into.

Reopen if the emitted-JavaScript path has a WASM-resolution gap this branch would still fix.

PR #884, opened 2026-07-21
feat(cli): add the `concept` noun (list/lookup/sample over ds:Concept)

advl:
## What

Surfaces the standalone **`ds:Concept`** documentation — cross-cutting guides, disambiguations, foundations, decisions that aren't bound to a single UIBlock (e.g. "Foundations: Grid", "How to differentiate components") — as a first-class read noun, projected to the CLI, MCP, and generated reference from one bundled pack.

```
pragma concept list                       # SPARQL: name / type / tier, with --search
pragma concept lookup "Foundations: Grid" # renders the Markdown ds:content body + summary/type/tier
pragma concept sample                      # exemplars
```

MCP tools: `concept_list` / `concept_lookup` / `concept_sample`.

## Changes

- **New bundled pack** `src/capabilities/concept/` (pack + module + test), registered in the capability array. `lookup` renders `ds:content` as a fenced **code** section so the long-form Markdown body reads cleanly.
- **Covenant + goldens:** `surface.v2.json` blessed 38 → 41 tools; `TOOL_HINTS`, the generated `docs/reference/*`, the completion goldens (`colophon → concept → config`), and the surface-count test expectations updated. The canonical fixture gains a `Concept` (with a colon in its name) so the parameterized noun suites (errorMatrix / agentSession / roundtrip) cover it.

## Shared kernel fix (beyond concept)

The pack lookup resolver classified **any** name containing a colon as a prefixed IRI — `resolveEntity.looksLikeIri` used `query.includes(":")`. A concept titled `Foundations: Grid` therefore failed with `INVALID_INPUT: Invalid prefix "Foundations"`. It now uses a strict CURIE pattern (`prefix:local`, no space after the colon), so a plain entity name that merely contains a colon is looked up **by name**, while `ds:global.component.button` still resolves as a prefixed IRI. Guarded by the concept lookup test and the colon-bearing canonical fixture name.

## Dependency & verification

Depends on the `ds:Concept` data landing via **canonical/design-system#64**; until then `concept list` is a clean empty result (not an error). The pack is exercised end-to-end against a fixture graph through the MCP projection. Full package `check` (biome / tsc / webarchitect) green; the concept + surface + completion + parameterized-noun suites pass. (Pre-existing create/doctor timeout flakes under full-suite parallel load are unrelated and pass in isolation.)

### DIFF
--- a/packages/cli/pragma/docs/reference/commands.md
+++ b/packages/cli/pragma/docs/reference/commands.md
@@ -121,6 +121,85 @@ pragma colophon  # the toolchain + active domain story
 pragma colophon --format llm  # condensed Markdown for agents
 ```
 
+## concept
+
+### pragma concept list
+
+List design-system concepts (standalone documentation).
+
+List design-system concepts — standalone documentation not bound to a single component (foundations, disambiguations, decision guides). Use when browsing cross-cutting design guidance. Example: concept_list {}.
+
+```
+pragma concept list [options]
+```
+
+**Flags**
+
+| Flag | Value | Description |
+| --- | --- | --- |
+| `--search` | `<string>` | Search in name, summary, type. |
+
+- Store: reads the local store (`pragma sources update` builds it).
+- MCP: exposed as the `concept_list` tool.
+
+**Examples**
+
+```bash
+pragma concept list
+pragma concept list --format llm
+```
+
+### pragma concept lookup
+
+Look up concept details by name, IRI, or glob.
+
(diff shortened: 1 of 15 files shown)


advl:
Superseded — the \`concept\` noun has landed on \`main\` by another route. \`concept_list\` and \`concept_lookup\` are both in the frozen surface covenant (\`surface/surface.v2.json\`), so this branch would now be re-adding what already exists.

Closing rather than rebasing; the branch is ~5 weeks and ~90 commits behind. Reopen if there is anything in here the landed version does not cover.

PR #885, opened 2026-07-21
fix(cli): mutating-command hardening — detect-before-write (create/config-set/sources-update)

advl:
Closes the detect-before-write gaps the mutating-command audit found (AV-356), against the `setup skills` / `upgrade` reference standard. Three fixes, one shared theme: **detect prior state, then act honestly.**

## 1. `create` refuses to overwrite without `--force` — REAL, highest severity (silent data loss) · breaking

`create component|package|application` (+ MCP `create_*`) wrote generated source with **no existence check** — the interpreter's `WriteFile` is an unconditional `fs.writeFile`, and `pathJail` is traversal-only. So `create component src/Button --yes` into a hand-edited path **silently overwrote every file** with the bare template, and the recap reported "Created …". The capability was even declared `destructive: false` on a false premise.

- New `assertNoOverwrite` (`overwriteGuard.ts`): on the non-interactive path, dry-run a **separate** task instance to enumerate the files the run would write, resolve against `rt.cwd`, and refuse with `INVALID_INPUT` (recovery: `--force`) when any exist. Fresh paths unaffected; `--dry-run`/`--undo` skip the guard.
- New **`--force`** flag on all three verbs (stripped before the generator sees answers). Covenant + reference updated. `destructive: false` is now honest by construction.
- `MutationRuntime` gains `undo` (was `preview`-only) so a verb can skip an overwrite guard under `--undo`; wired in the CLI + MCP projectors.

## 2. `config set` detects a no-op and shows `old → new`

`config set <key> <value>` read the config only to merge, never diffed — so `config set tier <same>` rewrote the file and printed "Set tier = X", indistinguishable from a real change. Now `writeConfigField` detects `previous` + `changed` and **gates the write** on `changed`; the recap reports unchanged / `old → new` / reset honestly.

## 3. `sources update` reports skill linking + honest cache-reuse wording

The per-skill created/skipped/replaced actions were computed but never surfaced (silent skill installs), and "Store up to date" meant *pack cache hit*, not "nothing changed". Now `SourcesUpdateData` carries a `skills` summary, the recap surfaces it, and the reuse line names the cache reuse rather than claiming a no-op.

## Deliberately deferred (noted in the audit, kept out to stay low-risk)

- `sources update` still always re-resolves each git package (an offline no-op can fail) — the detect-before-resolve / `--refresh` gate.
- `config set` / `sources update` `--dry-run` still can't show a true no-op — the dry-run interpreter mocks reads, so real-dry-run state is a separate change.
- `upgrade` needed nothing — it already follows the pattern (a second reference model).

## Verification

New tests: 4 overwrite-guard, 3 config-set idempotency, 1 sources skills-breakdown. Full package `check` (biome/tsc/webarchitect) + suite (943) green.


advl:
Closing — superseded in substance, and no longer rebaseable.

The detect-before-write idea landed on `main` in a more general form than this branch proposed:

- **`mutationContract`** in the frozen surface covenant — `--dry-run` / `--undo` / `--yes` are auto-injected onto every verb declaring `mutates: true`, rather than hardened case by case.
- **The setup target table** (`capabilities/setup/targets.ts`) requires `detect` / `plan` / `compose` / `removalPlan` on every row, so detection-before-write is structural rather than per-command.
- `create.verb.ts` carries the interaction-mode refusal path.

This branch is ~5 weeks and ~90 commits behind, and it edits `surface/surface.v2.json`, which has been rewritten since — so rebasing it would be re-authoring it.

**One loose end worth carrying forward rather than losing here:** this branch added `capabilities/create/overwriteGuard.ts`, and no file of that name exists on `main`. The `refuse` path that does exist in `create.verb.ts` is the interaction-mode refusal (non-TTY without `--yes`), which answers a different question from *"this file already exists."* Someone should confirm `pragma create` still declines to clobber an existing component — and if it does not, that is worth its own issue rather than a five-week-old rebase.

PR #896, opened 2026-07-31
[superseded by #908] feat: converged GraphQL base + the /graphql process split (PRD-1b, PRD-3)

advl:
## Done

Two logically independent commits on this branch (separately reviewable — say the word and they become stacked PRs):

### `d737692` — ke-graphql emits the converged Node/_meta base (PRD-1b)

```graphql
interface Node { uri: ID!  _meta: EntityMeta! }
```

Everything conventional (derived by heuristic) moves behind `_meta`; the top level is left for what the ontology declares. They fail differently and now say so: a null under `_meta` means the convention matched nothing, a null above it means the graph asserts nothing.

- **`uri` is the primary key and is the ABSOLUTE IRI.** `Node.id` and `Node.kind` deleted; consumers discriminate on `_meta.type.uri`. `Query.node`'s argument stays named `id`.
- **Collisions are errors, never renames.** `RESERVED_FIELD_NAMES` and the `M002` rename branch deleted.
- **Two knobs**: `prefixing` (real behaviour, default `none` — byte-compatible with the live app) and `mode` (provenance-only, no dead switch).
- **`title`/`label`/`comment`/`definition` live in `EntityMeta`** with `lang` arguments; `title` is `String!` with the fallback computed server-side.

### `4401a13` — split /graphql into its own process (PRD-3)

The graph stops being a Vite plugin inside the web server. `VITE_GRAPHQL_URL=https://elsewhere/graphql bun run dev:bun` spawns no local graph at all — the mechanism that makes "point the app at any provider" real. Deletes `graphqlPlugin.ts`, both double-boot guards, and the two-graphql-versions tightrope. Also fixed two latent failures the split surfaced (the unhandled incremental-stream rejection that would have killed the SSR process on a graph outage, and the launcher orphaning the graph on a spawn failure while logging the opposite).

### Follow-up commits (2026-07-31, on `claude/prism-despecialization-handoff-6lm3jq`, pending fast-forward of this branch)

- `a29c930` — provenance header completed to the contract block: `prefixing` + `validated-store` added, `provider`/`mode` order fixed.
- `bce6f2b` — external-implementation references removed from sources/tests/docs.
- `5481203` — e2e hit counting moved into the harness, incremental per line — immune to the 16 KB log-tail truncation (observed losing 14 of 17 hit lines).
- `69e1546` — the re-review fix round: 23 adversarially confirmed findings fixed, and **error-severity diagnostics are now fatal** (owner ruling: the compile refuses, a boot dies loudly rather than serving a schema minus silently dropped fields). New diagnostics `W001` (root-field occupancy) and `M006` (union member-set conflict).
- `e790b20` — the review tail closed: sdlOutput no longer truncates on artifact boots, collision-remedy texts stop overpromising what `prefixing: "all"` resolves, byte-stability/migration/injection-guard/zero-class pins added, docs aligned with code (16 extraction queries, real header example, honest store-backed-fields claim).
- **PRD-1 contract half, PRA-86 (7 commits, `d21e937` → `03ab6b9`)** — the contract package (created `b9193db`) reworked to the ruled Node/`_meta` base and wired into the workspace: the `packages/docsite/*` glob had been dropped from root `package.json`, so `@canonical/prism-contract` was invisible to `nx affected` and its tests had never run in CI — restored with its lockfile entry. The false browser-escape-hatch promise is deleted (the `??`-live `node:fs` import made it impossible; no speculative `./node` subpath). `OntologyClass` becomes a real resolvable Node in ke-graphql (`_meta.type` = `owl:Class`, frozen TBox-local ClassNode, identity-based `resolveType` branch — closes the R12 blocker), `node(id: <class IRI>)` resolves TBox classes (ABox first, strictly additive) so the promise is keepable, and **gate G-1** proves the live emission satisfies the contract: all 8 fixture emissions conformant, `prefixing: "all"` → 0 violations (the knob-independence ruling, measured), `relay: false` → exactly 1 (`FIELD_REMOVED` on `Query.node` — the teeth control). One recorded deviation from PRA-86's SDL block, rationale in `8cc088b`'s body: `ontologyClass(uri:)` / `ontologyProperty(uri:)` stay `String!` — the live emission's shape, and the deployed `DefinitionsExplorer` operation declares `$uri: String!`; promoting to `ID!` would invalidate it, and fixing that is an embargoed lens edit.
- **PRD-1d annotation-driven binding, PRA-96 (8 commits, `e6cbe77` → `9457aec`)** — the `graphql:` annotation vocabulary lands in ke-graphql: 13 terms extracted by a dedicated namespace probe (both value kinds, deduped, sorted), resolved into a typed overlay by a new pure `compiler/annotations.ts` (new A-band diagnostics A001–A008: conflicts/foreign-targets/malformed values are **fatal** per the R-4/R-9 rulings; consumer config shadowing a TTL annotation is a deliberate warning — the ruled workspace-local asymmetry), and bound across build/map/TBox as `config ?? annotation ?? heuristic`. Descriptive binding (`graphql:titleFrom`/`labelFrom`/`commentFrom`/`definitionFrom`) overrides the fixed chains per class with ancestor inheritance; `graphql:prefix` resolution lands compiler-side (annotation > registered > synthetic) while the app-side harvest removal is explicitly deferred as embargo-unsafe. **The three projection modes are now real** (`auto` = heuristics only, `annotated` = default unchanged, `explicit` = the ruled `graphql:expose` allowlist with aggregated drop diagnostics) — retiring the "provenance-only" clause exactly as ruled. `graphql:searchable` is captured in the IR only — every schema-visible search surface stays with the palette train (no speculative surface). **Proof of byte-compatibility**: commit `e6cbe77` pins all eight fixture emissions byte-for-byte as goldens generated *before* the feature; they hold unchanged through the final commit. `CustomMappings`/`nonNullOverrides` survive deprecated. 543 ke-graphql tests, coverage 100/100/100/100. ⚠ One open input flagged on the issue (O-1): the `graphql:` namespace IRI is keyed to a single constant with a placeholder pending confirmation against the published vocabulary — a one-line swap.

## Review status

The multi-lens review swarm ran to completion this time: 6 lenses over the base, **every finding through an adversarial verifier that confirms only by quoted code** (several by end-to-end reproduction against the real pipeline). 66 raw findings → 63 deduplicated → **all 63 verified** (39 adversarially by independent verifiers, 15 re-verified inline after a usage-limit interruption, plus 9 duplicates/already-fixed). Everything confirmed is fixed in `69e1546`/`e790b20`.

## QA

```bash
cd packages/runtime/ke-graphql && bun run check && bun run test && bun run build
#   biome · tsc · webarchitect 3/3
#   543 tests · coverage 100/100/100/100, no file below 100

cd packages/docsite/contract && bun run check && bun run test
#   biome · tsc · webarchitect 3/3
#   30 tests · coverage 100/100/100/100 (thresholds ratcheted 40 → 100)

cd apps/react/pragma-docs && bun run check
#   biome 376 files · tsc — green
```

### e2e — measured, no longer predicted (2026-07-31, refs layout from in-session sources)

An earlier revision of this description claimed "with a populated refs cache, all cells should pass except dev:express". **Measured: that was wrong.** With the refs precondition satisfied (first time ever — graph boots in ~2 s from 393 TTL sources):

| cell | result | why |
|---|---|---|
| `dev`, `preview` | **PASS** | unchanged |
| `dev:express` | FAIL | pre-existing Node-22 `nodeCssNoop` boot error, orthogonal |
| `dev:bun`, `preview:bun`, `preview:express` | FAIL — **one step later than ever before** | the committed relay schema/artifacts still expect `id: ID!`/`uri: String!` while this branch's compiler emits the converged base; SSR prepare rejects and the page client-renders. This is the PRD-1 migration-window skew, resolved by the (currently embargoed) schema regeneration + query migration — not by this PR |

**First-ever in-test passes**: the refs-cache precondition assertion and the graph-side content-type guard (`GET /playground` on the graph → 404 `application/json`), on all three booting probe cells. The hit-count assertions and the web-side stray-POST guard sit after the current abort point; **verified manually against a live cell**: ssr hits 17 (needs >5), client 0, strict delta exactly +1 on an unmarked POST, stray POST to the web server → `text/html` with the graph counter unchanged. The probe block was not defanged.

## Known gaps — stated, not hidden

- ⚠️ The probe cells stay red until the migration window (committed schema regeneration + lens query migration) — gated on the semantics tree, tracked separately.
- **Behavioural note:** the launcher's signal handler removes the default disposition, so SIGTERM during the graph-boot window exits **1 instead of 143**.

### Pre-existing failures at repo root, none from this PR

`pragma-cli:check` · `pragma-cli:test` · `svelte-ds-global:test` — unchanged baseline. (`svelte-ds-app:test` is additionally red in the verification container only — byte-identical missing-Playwright-executable environmental error as the svelte-ds-global one; nothing in the svelte tree is touched by this branch.)

## Decisions

The four "need a nod" items from the original description — absolute-IRI identity, `OntologyClass` and `Node`, the language-resolution tier, collision fatality — **are all owner-ruled as of 2026-07-31**. The ruling record is `pragma-adrs session/R/R.03.GRAPHQL_RULINGS.md` (companion PR advl/pragma-adrs#20); collision handling was subsequently escalated to fatal by the same rulings and shipped in `69e1546`.

Two further items are flagged for the owner on their issues, each proceeding on a recorded default: the contract lookup args stay `String!` (PRA-86 — emission wins while the lens migration is embargoed), and the `graphql:` namespace IRI placeholder (PRA-96 O-1 — one-line swap on confirmation).

### PR readiness check

- [ ] Label: `Feature 🎁` (both commits) — `Breaking Change 💣` also applies to `d737692`
- [x] Code follows the [code standards](https://github.com/canonical/web-code-standards) — reviewed against the `cs:` ruleset by a multi-lens swarm with adversarial verification; all confirmed findings applied
- [x] All packages define `check`, `check:fix`, `test`; `ke-graphql` defines `build`/`build:all`
- [ ] **New package** (once fast-forwarded): `@canonical/prism-contract` at `packages/docsite/contract` — created `b9193db`, wired into the workspace in `d21e937`; defines `check`, `check:fix`, `test`, coverage 100%
- [ ] `no visual change` — applies, no component rendering changed

## Screenshots

n/a — server, compiler and schema only.

Companion PR: the schema decision record in [`advl/pragma-adrs#20`](https://github.com/advl/pragma-adrs/pull/20) (`R.02.GRAPHQL_SCHEMA.md` + `R.03.GRAPHQL_RULINGS.md`).

https://claude.ai/code/session_01MT2BJ5QRRF4JLrWNmHuV1a


advl:
## ⛔ DO NOT MERGE — four confirmed defects in `d737692`, one of them a live crash

The ke-graphql review completed after this PR was opened and found four blocking issues. **None are fixed** — a fix round was started and died on a quota limit before writing anything. Working tree is clean; nothing is half-applied.

Every finding below was confirmed by an adversarial verifier that **executed the code**, not by reading it. Two were re-graded *down* from high while being confirmed.

---

### 1. 🔴 `lang: null` crashes every `_meta` descriptive field — and nulls the whole response

`resolver/descriptive.ts:130` dereferences `want` unconditionally. `lang` is a **nullable** `GraphQLString` with a default, and graphql-js substitutes a default only for an **omitted** argument — an explicit `lang: null`, or a `$l` variable passed `null`, arrives as `null`.

Verified against the pinned `graphql@17.0.0-rc.0` with a real compiled fixture:

```
{ things { edges { node { _meta { title(lang: null) } } } } }
→ {"errors":[{"message":"null is not an object (evaluating 'want.toLowerCase')",
              "path":["things","edges",0,"node","_meta","title"]}],"data":null}
```

**`data` is `null` entirely** — `title` is `String!` under `_meta: EntityMeta!` under `node: Node!`, so non-null propagation takes the whole response down. Nullable siblings (`label`/`comment`/`definition`) null only their own field.

Two compounding defects:
- `tbox/buildTBoxSchema.ts:392,406,419,433` type the args `{ lang: string }` — asserting a guarantee graphql-js does not give.
- **`shared/constants.ts:144` states the false guarantee outright** in its TSDoc: *"graphql-js coerces the default in, so resolvers always receive a string."* That comment is why this shipped, and it needs correcting alongside the code.

Latent on the current `ds:`/`cs:` corpus only because `tagMatches` short-circuits on `tag !== ""` and every literal there is untagged — so it needs one language-tagged literal in the chain to fire. That is not a defence; it is a fuse waiting on the first tagged label.

**Fix:** make the boundary total (`args.lang ?? DEFAULT_LANG` at the four resolver sites, and/or make `tagMatches` total). `cs:code.fallback.defensive` governs — raw client input. **Tests must cover `lang: null` inline *and* via a variable**, since only the variable path distinguishes omitted-from-null.

### 2. `prefixing` is absent from the SDL provenance header

`compose.ts:88-97` stamps banner / `graphql-schema-spec` / `mode` / `provider` / `revision`.

- `mode` is stamped and provably **inert** — no branch on it anywhere in the pipeline.
- `prefixing` is stamped **nowhere** and provably changes **every field name** (`applyPrefixing`, `map.ts:130-141`, turns `title` into `libTitle`).

So two SDLs with byte-identical headers can have **disjoint field-name sets**, which defeats the header's only purpose. `docs/architecture.md:64` currently documents the omission as deliberate — that is the wrong call and should change with the code.

**Fix:** `prefixing?: FieldPrefixing` on `ComposeOptions`, pass it through `runPasses.ts:44-51`, add the header line. Then update the "five-line" wording at `compose.ts:82`, `compiler/constants.ts:78`, `docs/architecture.md:54-64`, `README.md:84`, and the header assertions in `compose.test.ts:331-370` and `pipeline.test.ts:294-302`.

### 3. The contract SDL is stale — 8 measured violations

`packages/docsite/contract/schema/contract.graphql:34-41` still declares the old six-field `Node`; `OntologyClass.uri` (`:71`) and `OntologyProperty.uri` (`:101`) are still `String!`.

`findBreakingChanges` was actually run (contract as `old`, converged base as `new`): **5× `FIELD_REMOVED`** (`Node.id/kind/label/comment/definition`) **+ 3× `FIELD_CHANGED_KIND`** (`Node.uri`, `OntologyClass.uri`, `OntologyProperty.uri` — `String!`→`ID!`).

The contract's own header names `compose.ts`/`buildTBoxSchema.ts`/`wireRelay.ts` as its source of truth, so this is a **derived artifact the commit left stale**, not incidental docs. Graded medium only because nothing wires it today — but **the published provider gate now rejects the reference implementation it was derived from**, which is precisely the check `sem-graphql` is meant to be validated against.

Also fix the now-false `node(id:)` description at `:138` ("prefixed-URI global ID" → absolute IRI).

### 4. Under `relay: false`, the M005 drop deletes a field for colliding with nothing

`runPasses.ts:43` skips `wireRelay` entirely when `relay === false`, so no structural `uri`/`_meta` is ever injected and the compiler owns neither name. But `map.ts:277-296` picks the structural set from `node.embeddable` alone, with no reference to `options.relay` — so an ontology field is **dropped for colliding with something that will never exist**, and the diagnostic text is false in that mode. Confirmed by compiling both ways.

`buildFields` already reads `state.options` for prefixing (`:132`) and `nonNullOverrides` (`:329`), so the plumbing exists.

---

### Full findings

`blocking`, `low` and `refuted` arrays, each with the verifier's independent evidence and a minimal fix:
`/tmp/claude-0/-home-user/525f51f2-d958-59a2-9e47-1a8b96ec4df6/tasks/wddkiasur.output`

⚠️ **That path is container-local and will not survive.** The four blocking findings are reproduced above in full; the `low` entries are not, and will be lost if the container is reclaimed before someone reads them.

### Note on the first review

An earlier run of this same swarm returned `{"blocking":[],"low":[]}`. **That was not a clean bill of health** — 5 of its 9 agents had died on a session limit, including every verifier, so findings were produced and then dropped. This re-run completed all 10 agents with zero errors. Worth knowing before trusting any empty review result.

### `4401a13` (the process split) is unaffected

Its own review completed, its one blocking finding was fixed, and every fix was demonstrated against a counterfactual. All four defects above are in `d737692` only.

---

advl:
## ✅ All four findings fixed — the DO-NOT-MERGE above is lifted

Four commits, one per finding, pushed to this branch.

| | Finding | Commit |
|---|---|---|
| 1 | `lang: null` nulled the whole response | `a0187e0` |
| 2 | `prefixing` absent from the provenance header | `57c4ecd` |
| 3 | contract SDL stale — 8 violations | `0e00cf1` |
| 4 | M005 dropping fields under `relay: false` | `8f0da29` |

**ke-graphql: 35 files / 415 tests, coverage 100/100/100/100.** Contract package: 22 tests. `check` green in both.

### 1 — closed by counterfactual, the same way it was found

Fixed at the boundary: the four resolvers typed `{ lang: string }` were **asserting a guarantee the runtime does not give**; they now type `string | null` and apply `?? DEFAULT_LANG` (`cs:code.fallback.defensive` — raw client input).

Reverting *only* `buildTBoxSchema` reproduces the crash, and restoring it clears all four cases:

```
                          BEFORE                                    AFTER
A. lang omitted        data={"…title":"Widget"}   errors=[]      data={"…title":"Widget"}
B. lang: null INLINE   data=null  errors=["null is not an        data={"…title":"Widget"}
                              object (evaluating 'want.toLowerCase')"]
C. lang: null VARIABLE data=null  errors=[same]                  data={"…title":"Widget"}
D. lang: "en"          data={"…title":"Widget"}   errors=[]      data={"…title":"Widget"}
```

**The comment was the more important half.** `LANG_ARGS`' TSDoc stated the false guarantee outright, and a test comment repeated it — that belief is why this shipped. Both now say what actually happens and why typing it non-null caused a crash.

Both paths are pinned, across all four descriptive fields. Testing only the inline form would have left half the bug live: **only the variable path distinguishes "omitted" from "explicitly null"**, and only covering `title`/`label` left the `comment`/`definition` branches unexercised (caught by the coverage gate dropping to 99.84).

### 2 — the reasoning is reversed, not just the code

`architecture.md` argued `prefixing` "has real behaviour now and is not part of the header". That is backwards — **having real behaviour is the qualification.** `mode` is stamped and provably inert; `prefixing` renames every generated field. Two SDLs with identical headers could have disjoint field-name sets, which defeats the only thing a provenance header is for. The doc now records why, so the argument isn't re-run. Five lines → six, everywhere.

### 3 — verified by execution, 8 → 0

```
BREAKING (must be 0): 0
dangerous (ignored by the predicate): 4
   · OPTIONAL_ARG_ADDED  EntityMeta.title(lang:) was added.
   · OPTIONAL_ARG_ADDED  EntityMeta.label(lang:) was added.
   · OPTIONAL_ARG_ADDED  EntityMeta.comment(lang:) was added.
   · OPTIONAL_ARG_ADDED  EntityMeta.definition(lang:) was added.
```

`EntityMeta` gains the descriptive fields **without** the `lang` argument, deliberately: adding an optional arg is *dangerous*, not *breaking*, and the predicate ignores dangerous — so a provider that has `lang` satisfies this, while a contract demanding `lang` would reject one that lacks it. **Argument-free is the wider floor**, and those four entries are that asymmetry working as intended.

One updated test had hardcoded `"3 violation(s)"`. That count is incidental to the contract's size, so it now **derives** the count from the predicate and asserts every violation message appears — the invariant is that the error lists them *all*, and a magic number breaks whenever the contract legitimately grows while saying nothing about completeness.

### 4 — pinned in both directions

The guard now reads the same option that decides whether the fields exist. One test asserts the two properties survive under `relay: false` **and** that the same IR under the default still produces two M005s and an empty type — so weakening the guard fails rather than passing quietly.

### Root gate

`packages/cli/pragma` — **111 files / 951 tests pass**; the 2 failures are the pre-existing `--framework` covenant pair (`surface.test.ts`, `docExamples.test.ts`), neither mentioning graphql, and `git status` on that package is empty. `svelte-ds-global` still can't find its Playwright binary. Both predate this branch.

---

advl:
## Superseded by #908 — closing

This PR and #908 are the same line of work. **#908's branch was built directly on top of this one**, so it contains everything here except four commits, and 20 more besides.

Those four (`a0187e0`, `57c4ecd`, `8f0da29`, `0e00cf1`) re-fixed defects that #908 had already fixed. I wrote them in a session that never checked for other branches — I read the issue tracker to learn what was done, when only the repository can tell you what exists. The board was several days stale.

In two places #908's version is also **better**:

| | here | #908 |
|---|---|---|
| provenance header | 6 lines | **7** — also `validated-store`, field order corrected |
| collision diagnostics | error + drop | **fatal** — compile refuses, boot exits non-zero (owner ruling R.03 R-4) |
| `OntologyClass` | `uri: ID!`, not a `Node` | **a real `Node`** — the `_meta` gap that forced the narrower call is closed |
| external attributions | left in place | **scrubbed**, per owner policy |

Two judgement calls recorded in this thread were also overruled after the fact and should not be cited as current: that error diagnostics should merely report rather than abort, and that `OntologyClass` could not implement `Node`. Both were defensible when made; neither holds now.

**What stays useful here:** the defect analysis in the comments above. The four defects were real, and two independent sessions found them from different directions — the reasoning holds even though the commits are discarded. Worth reading before reviewing #908; the `lang: null` case in particular is a good illustration of why an argument default is not a totality guarantee.

Nothing is lost by closing this. → **#908**

---

PR #897, opened 2026-07-31
feat(cli)!: the first install answers reads offline from the embedded pack

advl:
## Done

- The embedded pack is compiled from `pragma.conf.ts` at build time: 8 479 triples / 550 entities ship inside the binary.
- A fresh install answers reads offline: `block list` returns 251 rows cold — no network, no store, no setup step.
- The lock file is **removed**, not kept (execution-account ruling §10.1; breaking changes cost 0 by owner ruling).

**Stack:** PR 1 of 4 (slice 1, ~14 commits), plus a docs-only ADR PR in `advl/pragma-adrs`. Merge order: this PR → `claude/despec-2-identity-create` → `claude/despec-3-vocabulary-stories` → `claude/pragma-cli-despecialization-2nh2ym`. Squash-merge each; successors rebase after each merge. Pre-open sanity review: OPEN_OK.

## QA

- `bun install && bun run check && bun run test` from the repo root.
- `cd packages/cli/pragma && bun run build`, then with networking disabled run `./dist/pragma block list` — 251 rows answered from the embedded pack, cold.

### PR readiness check

- [x] PR label: `Breaking Change 💣` (+ `no visual change`).
- [x] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format.
- [x] The code follows the appropriate [code standards](https://github.com/canonical/web-code-standards).
- [x] All packages define the required scripts in `package.json` (existing package; `check`, `check:fix`, `test`, `build`, `build:all` unchanged).
- [ ] New package: n/a — none introduced.
- [x] No visual testing required — `no visual change` label added (CLI only).

## Screenshots

n/a — CLI-only change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017EwUvEW5ZgjAYc9KWPjHCZ

PR #901, opened 2026-07-31
fix(ci): playwright browsers install from the suites' own resolution; need detected on the warm graph

advl:
## Done

Fixes `main`'s CI, red since at least `7f6fe7c` (2026-07-29): svelte browser suites failing with `Executable doesn't exist` for `chromium_headless_shell-1228` / `firefox-1532`. **CI-only** — no dependency or lockfile changes (owner ruling). Two defects, both fixed:

**1 — The installer never resolved the workspace's playwright.** Bun's isolated layout does not hoist playwright to root `node_modules` (verified: the path does not exist on a fresh CI install), so `bunx playwright install` silently fell back to fetching the **latest** playwright from the registry and installed *that* version's browser builds — which the suites' pinned `playwright-core` cannot find (observed live: installed `1234` = registry-latest, wanted `1228` = the suites' 1.61.1). The install now lives in **`setup-env/install-playwright.sh`**: it resolves through `packages/svelte/ds-global` — a browser suite that declares `playwright` directly, so its resolution is the correct version by construction, in any lock state — logs the version, and **verifies the exact revision directories** its `playwright-core/browsers.json` names. A wrong-version or missing install fails the job at setup, cause named, in seconds. (The `ds-app-wpe` / `ds-global` playwright range split in the tree is deliberately untouched — no longer CI-load-bearing; aligning it is the wpe team's own hygiene call.)

**2 — The need-detection ran in the one place it could silently lie.** The old gate (`check-playwright.sh`) ran inside `setup-env` as the job's **first nx invocation on a cold machine**, where its query resolved an **empty** project set — exit 0, nothing logged — while the test step's own `nx affected -t test`, minutes later on the warmed graph, still ran the browser suites. Detection is kept (owner ruling) but **moved to where it cannot lie**: `pr.yml` detects **after** setup — the build has already exercised nx, so the graph is warm, the same conditions the test step computes under — logs the affected set, the tag universe, and their intersection on every run, and **fails open** on anything suspect (query error, or an empty tag universe while tagged projects exist in-tree). `push.yml`/`tag.yml` install unconditionally — their full suite always includes the browser suites, so detection there is a failure surface with nothing to buy. The composite sheds all browser logic.

- Unblocks CI signal for every open PR including the despecialization stack #897→#900 (lands **first**; the stack re-runs checks on top).
- Touches no code-owned path.

## QA

- **Both detection outcomes proven green on this PR's own runs:** the install path (browsers detected as needed → installed → svelte suites pass on 1.61.1 — the first green `build-gate` since the redness began), and the skip path (a `.github`-only head → detection resolves no affected browser suite → install skipped → full matrix green in ~45 s).
- Locally, against `main`'s untouched lock: `packages/svelte/ds-global/node_modules/playwright` resolves 1.61.1, whose `browsers.json` names exactly `chromium_headless_shell-1228` / `firefox-1532` — the executables every failure named.
- Script `bash -n` + not-requested path exercised; all touched YAML validated.

### PR readiness check

- [x] PR label: `Bug 🐛` (+ `no visual change`).
- [x] The code follows the appropriate [code standards](https://github.com/canonical/web-code-standards).
- [x] All packages define the required scripts in `package.json` — n/a, CI only.
- [ ] New package: n/a — none introduced.
- [x] No visual testing required — `no visual change` label added (CI only).

## Screenshots

n/a — CI change only.

https://claude.ai/code/session_017EwUvEW5ZgjAYc9KWPjHCZ


advl:
## Why the nx-affected install gate is removed rather than restored

For the review record, since this PR deletes `check-playwright.sh` outright — the design decision, with the evidence from this PR's own runs:

**The gate's input provably disagrees with what the tests run.** Twice, measured live in CI here, the gate's setup-time query (`nx show projects --affected --projects tag:playwright`) returned **empty** — exit 0, no output — while the test step's own `nx affected -t test`, in the same job against the same base minutes later, still selected and ran the browser suites (47 projects, including both `tag:playwright` carriers). The same two invocations agree when run locally on the same commits. A skip-optimization whose deciding calculation can disagree with the calculation that selects the tests cannot be hardened into safety — a fail-open wrapper was tried in-branch (`63d7300`) and still couldn't distinguish "legitimately nothing affected" from "graph didn't resolve".

**Its failure mode is the worst kind: silent.** The gate logged nothing about its decision. That is how `main` stayed red for three days with every log showing only the downstream symptom (`Executable doesn't exist`) and nothing about the skipped install.

**What it saved doesn't cover what it cost.** With the browsers cache keyed on `hashFiles('bun.lock')`, a cache-hit "install" costs seconds of restore plus ~30 s of (uncacheable) apt deps; the full download only fires when the lock actually changes the playwright resolution. Weighed against a three-day repo-wide outage, the per-PR minute loses.

**The replacement keeps the intent explicit instead of inferred:** each job declares `install-playwright` (test jobs `'true'`, chromatic `'false'`), the decision is made in shell inside `install-playwright.sh` (composite-step `if:` conditions mis-evaluated silently twice during this investigation and are out of the correctness path), and the script verifies the exact browser revisions the suites' own `playwright-core` names — so any future regression of this class fails at setup, cause printed, instead of 200 browser tests later.

**If the skip is ever wanted back** for browser-unrelated PRs, the safe shape is: compute `nx affected` **once** in `pr.yml`, and feed that single list to both the test command and the install decision — one calculation, structurally no disagreement. Deliberately not built here; can be registered as a follow-up if the ~1 min/PR matters.

---

advl:
Update on the comment above: the owner subsequently ruled detection back in, and the closing paragraph's "safe shape" is now implemented in this PR — need-detection runs in `pr.yml` **after** setup on the warmed graph (the same conditions the test step computes under), logs affected ∩ tag:playwright on every run, and fails open on anything suspect. The analysis above stands as the record of why the detection could not remain where it was (a cold-graph first invocation inside the composite); the skip path and the install path have both since run green on this PR.

---

PR #906, opened 2026-08-03
feat: migrate button in svelte

nbellol:
## Done

[List of work items including drive-bys]

Fixes [list issues/bugs if needed]

## QA

- Open stories
- Compare the behavior and documentation against [ds-global-button](https://ds-react-global.canonical.com/?path=/docs/components-button--docs)

### PR readiness check

## Screenshots

image


packages/svelte/ds-app-wpe/src/lib/components/Button/styles.css
```diff
@@ -0,0 +1,256 @@
...
+  --button-color-text: var(--color-text);
+  --button-color-border: var(--color-border-highlighted, var(--color-border));
+
+  /* Inline padding (Figma):
+     - end (right): always dimension-200 (16px) — the label always sits 16px
+       from the trailing edge.
+     - start (left): dimension-200 (16px) with no icon; dimension-100 (8px) when
+       an icon leads (the icon then sits closer to the edge), with a
+       dimension-100 (8px) gap between the icon and the label. */
+  --button-padding-inline-end: var(--dimension-200);
```
jmuzina, line 27: There seems to be an alignment problem here:

Image

But it is also present in the React implementation, so I will aim to fix it in upstreaming.

packages/svelte/ds-app-wpe/src/lib/components/Button/Button.svelte.test.ts
jmuzina, line 1: Have a look at the [testing standards](https://github.com/canonical/web-code-standards/blob/docs/svelte-standards/docs/testing.md#unit-testing) and the [React version's tests](https://github.com/canonical/pragma/blob/e8d25fecbe6e70edfc4cb300bad65f691dbbd26e/packages/react/ds-global/src/lib/component/Button/Button.tests.tsx) - could you reorg the test cases to follow the standard (`describe` per prop/modifier family/state + major things like accessibility) and have a similar test breadth to the React version?

packages/svelte/ds-app-wpe/src/lib/components/Button/Button.svelte
```diff
@@ -0,0 +1,76 @@
...
+  ]}
+  aria-busy={loading || undefined}
+  disabled={disabled || loading}
+  {...rest}
+>
+  {#if icon}
+    <span class="icon">
+      {@render icon()}
+    </span>
+  {/if}
```
jmuzina, line 53: Good for now, though I'd leave a note that Icons are going to change after we have a Svelte Icon component ([WPE-430](https://warthogs.atlassian.net/browse/WPE-430)), currently blocked by #959 . Once these are done I will adjust this in the Button for upstreaming. Maybe worth a TODO note that the Icon api here is temporary

jmuzina (commented):
Thank you for this!

packages/svelte/ds-app-wpe/src/lib/components/Button/styles.css
nbellol: Should I fix it in both or at least here?

packages/svelte/ds-app-wpe/src/lib/components/Button/Button.svelte.test.ts
nbellol: Thank you for the guidante, I will change this.

packages/svelte/ds-app-wpe/src/lib/components/Button/Button.svelte
nbellol: Added

packages/svelte/ds-app-wpe/src/lib/components/Button/styles.css
jmuzina: If the fix seems trivial, I would suggest to implement it just in our tier here for now to keep the required reviews within WPE. Then I could apply an upstream fix while upstreaming this.

nbellol: I have applied a fix for this

jmuzina: Nice, it looks better 

image

But the label text seems a bit high (or the icon is too low). I suspect this may be due to a [known issue with typography line height tokens](https://github.com/canonical/pragma/pull/961#issuecomment-5421532811) I heard about from @diana-stanciulescu today. 

Let's keep your fix (this is an improvement already) and I can propose it upstream like this!

jmuzina (approved):
thank you!

packages/svelte/ds-app-wpe/src/lib/components/Button/Button.svelte
```diff
@@ -0,0 +1,81 @@
...
+  class={[
+    componentCssClassName,
+    importance,
+    anticipation,
+    variant,
+    loading && "loading",
+    className,
+  ]}
+  aria-busy={loading || undefined}
+  disabled={disabled || loading}
```
jmuzina, line 41: TODO for upstreaming / consideration by the DS team: @advl @stavropouloskonstantinoss @diana-stanciulescu 

This is consistent with the global React implementation, but I think the global implementations should consider [this longstanding Button accessibility issue](https://github.com/canonical/vanilla-framework/issues/4299) that we have the chance to solve now in Pragma before lots of projects rely on it. There are downsides to using the `disabled` attribute but potentially still valid cases for using it.

Maybe there has been some discussion on this already, let me know :)

diana-stanciulescu: we did decide in forms to stop disabling buttons and allow users to click and get an error message . howewer, I still see situations where disabled buttons may be useful, especially for action bars that have an unavailable action for the specific situation.  I don't think we fully solved this but this is the current guidance: https://docs.superhuman.com/d/_dNyzE_TLZDh#UI-Blocks_tuwIHYhx/r12&view=center (under "when to disable buttons")

also, for context: as part of the spacing conversation,  for buttons we are looking at how to best achieve square icon buttons - as a separate component or not (we do want square icon buttons). also, the react implementation currently does not have text an icons vertically simmetrical in the box which we do want to have - that's in the pipeline to be fixed.

packages/svelte/ds-app-wpe/src/lib/components/Button/styles.css
diana-stanciulescu: That issue I mentioned makes line heights smaller than thay should be, not sure if that affects alignment here - the ideal status is symmetry. Also the icon color in all buttons should be the same as the text, so black/white in secondary buttons.

PR #910, opened 2026-08-05
feat(prism): extract the pragma provider into its own package

advl:
## Done

Pragma knowledge does not vanish from the docsite — it **relocates**. This extracts the provider assembly the app was still carrying into a package that sits beside `@canonical/prism-graph-example`, so the two providers are peers and the app is a consumer of neither in particular.

- **New package `@canonical/prism-pragma-provider`** (private, not published) — the ref/sem package lists, the TTL collection and prefix harvesting, the `shim-concept.ttl` exclusion, the `anatomy:uri` custom mapping, and the compiler options, all in one place. 45 tests at 100% coverage.
- **`apps/react/pragma-docs` drops `@canonical/ke` and `@canonical/ke-graphql`** — both imports and both dependency entries. `src/server/graphql.ts` (296 lines) is deleted; `src/server/graph.ts` keeps the transport it always owned — the port literal, the `/graphql`-only routing, the hit counter the e2e pins — and gains one factory call.
- **Gate G-2 lands** — the committed `schema.graphql` is measured against `@canonical/prism-contract` on every run. Green at 0 violations. It throws rather than skips if the file is missing: a gate that goes quiet when its subject disappears is worse than no gate.
- **`sdlOutput` becomes a caller-supplied option with no default.** The path belongs to the app's source tree; a provider deriving it from its own `import.meta.url` would resolve inside `node_modules` with no type error and no failing test. Absent means write nothing.

**Drive-bys:**

- **A hermetic TTL corpus** makes the provider testable without a refs cache, and turns the `anatomy:uri` collision from an unexplained wart into a falsifiable property.
- **R-4's fatality was resolved empirically, and the ruling record was the stale side.** It read *"error+drop implemented; fatality pending"*. Measured against the corpus with identical inputs but for the mapping: **with** it the compile succeeds and emits `anatomyUri`; **without** it `runPasses` **throws `CompilationError`**. It does not drop the field and hand back a schema. Asserted on every run now.
- **The two moved `.sdl.txt` captures are stale** — pre-converged-base, **12 contract violations each**, measured. Their test stays meaningful because it asserts *additivity of the second source root*, a property of the source set orthogonal to the structural head, and the type inventory has not drifted (183 definitions, matching the live schema). They cannot be regenerated in an environment without the full three-package refs cache, and a capture from a partial cache would be worse than a stale one because it looks current. Headers now carry that measurement.
- A dead comparator arm was removed rather than coverage-ignored, with the uniqueness precondition that makes it dead written down.

## QA

```bash
bun install

# the new package
cd packages/docsite/pragma-provider && bun run build && bun run check && bun run test
#   → 5 files / 45 tests, coverage 100/100/100/100, webarchitect library ruleset passes

# the app — 9 tests moved out with the sources test they belong to
cd ../../../apps/react/pragma-docs && bun run check && bun run test
#   → 95 files / 440 tests   (was 96 / 449)
bunx relay-compiler --validate
#   → 23 reader / 10 normalization / 23 operation text

# nothing regressed
cd ../../../packages/docsite/graph-example && bun run test    # → 227 tests, 100%
bunx vitest run src/testing/integration/                      # → acceptance gate: 4 discovered, 0 failing
cd ../contract && bun run test                                # → 32 tests, 100%
cd ../../runtime/ke-graphql && bun run test                   # → 551 tests, 100%

# the app is free of the knowledge engine
cd ../../../apps/react/pragma-docs
grep -rn "from '@canonical/ke" src test      # → no output
grep -n "canonical/ke" package.json          # → no output

# the emitted schema is untouched
git status --porcelain src/relay/schema.graphql   # → empty
```

**Do not run `bun run dev`, `dev:bun`, `graph`, `preview*` or `test:e2e` without a populated `~/.cache/pragma/refs`.** A boot from a partial cache rewrites the tracked `src/relay/schema.graphql`, and the truncation is silent.

### PR readiness check

  - **N/A — `@canonical/prism-pragma-provider` is `"private": true` and is never published**, exactly as its peer `@canonical/prism-graph-example` is. No manual publish, no trusted publisher, no release-workflow involvement.
  - No component renders differently. The diff is a package extraction, a server-side factory call, and a test gate.

## Screenshots

n/a — package extraction, server wiring and a schema gate. Nothing renders differently.

---
_Generated by [Claude Code](https://claude.ai/code/session_019VVct8koTvTWrCzABQp3i2)_


advl:
## CI is red, and it is not this PR

The `build-matrix` failure here comes from the base branch's CI setup, not from anything in this diff.

**What passed.** `@canonical/pragma-docs` ran clean on both matrix cells: **95 test files / 440 tests**, matching the local baseline exactly.

**What failed.** Only two tasks, both Svelte packages unrelated to this work:

```
- @canonical/svelte-ds-global:test
- @canonical/svelte-ds-app:test
```

Neither failed an assertion. Both died before running a browser test:

```
Error: browserType.launch: Executable doesn't exist at
/home/runner/.cache/ms-playwright/chromium_headless_shell-1228/chrome-headless-shell-linux64/chrome-headless-shell
```

**Root cause.** `.github/actions/setup-env` on this base runs `bunx playwright install --with-deps …` from the repository root, and the root package declares no `playwright`. Bun's isolated layout hoists nothing to root `node_modules`, so `bunx` falls back to fetching a copy from the registry and installs *that* version's browser revisions. Measured in a clean checkout:

| Resolution | Version |
| -- | -- |
| `bunx playwright --version` at the repo root | **1.56.1** (registry fallback) |
| `bunx playwright --version` in `packages/svelte/ds-global` | **1.61.1** (the declared devDependency) |
| `playwright-core` that `@vitest/browser-playwright@4.1.2` launches through | **1.61.1** → wants chromium-headless-shell **1228** |

So setup provisions revision 1194 and the runtime looks for 1228.

**This is already fixed on `main`.** `087db67` (#901) replaced the root invocation with `.github/actions/setup-env/install-playwright.sh`, which installs through an anchor package that declares playwright and then verifies the produced revisions against `browsers.json`, so a version-mismatched install can no longer pass setup silently. `main` is 11 commits ahead of `feat/prism-docsite`; the base has simply never received that commit.

**Not fixing it here.** The fix belongs in a merge of `main` into `feat/prism-docsite`, which is a separate change with eight conflicts of its own — five of them real ones in `packages/cli/pragma`. Duplicating #901's fix onto this branch would create a conflicting third copy of the same file. That merge is being prepared as its own PR.

---

PR #911, opened 2026-08-05
chore: merge main into feat/prism-docsite — unblock CI, and fix two breakages neither side had alone

advl:
## Done

Merges `origin/main` (11 commits ahead) into `feat/prism-docsite`. **`.github/` is byte-identical to `main` afterwards** — `git diff origin/main -- .github/` is empty.

**Why now: CI on this base is red for a reason that has nothing to do with the work on it.** PR #910 ran the first CI on this branch. `@canonical/pragma-docs` passed cleanly (95 files / 440 tests). The only failures were two Svelte packages, and **no assertion failed** — both died before running a test:

```
browserType.launch: Executable doesn't exist at
/home/runner/.cache/ms-playwright/chromium_headless_shell-1228/chrome-headless-shell-linux64/chrome-headless-shell
```

`.github/actions/setup-env` on this base runs `bunx playwright install` **from the repository root**, where nothing declares `playwright`. Bun's isolated layout hoists nothing, so `bunx` falls back to fetching a copy from the registry and installs *that* version's browser revisions:

| Resolution | Version |
| -- | -- |
| `bunx playwright --version` at the repo root | **1.56.1** (registry fallback) |
| `bunx playwright --version` in `packages/svelte/ds-global` | **1.61.1** (the declared devDependency) |
| `playwright-core` that `@vitest/browser-playwright@4.1.2` launches through | **1.61.1** → wants revision **1228** |

`main` already fixed this in `087db67` (#901), which installs through an anchor package that declares playwright and then verifies the produced revisions against `browsers.json`. This branch had simply never received it. PR #909, which targets `main`, is fully green — the same workflow on the same runner, differing only in base.

### Conflicts and how each was resolved

| File | Resolution |
| -- | -- |
| `bun.lock` | Took main's, ran `bun install --ignore-scripts`, committed bun's own output. Verified a further clean install leaves `git diff --exit-code bun.lock` empty — the exact check `setup-env` runs. |
| `create/templates.embedded.generated.ts` | Took main's side, then **regenerated** via `bun run scripts/build.ts`. Not hand-merged. |
| `create/create.verb.ts` | Main's structure, this branch's intent. See below. |
| `create/pickGenerator.ts` | Main's structure, this branch's intent. See below. |
| `kernel/config/defaults.ts` | Main's file whole; the branch's `git+ssh` flip re-applied at its new home. |
| `kernel/runtime/refs/resolve.ts` | Main's structure; two of the branch's three fixes ported, one dropped as meaningless. |
[...]

PR #914, opened 2026-08-08
test(prism): the proof — the core lenses render against a provider that has never heard of pragma

advl:
## Done

**Stacked on #910.** Review that one first; this PR's diff is only its own 8 commits.

This is the despecialization train's goal made runnable. The app is booted — its real `dev:bun` script, its real prepare step, its real renderer — against `@canonical/prism-graph-example`, a hand-written GraphQL provider over a fictional metro network whose only relationship to this repo is `@canonical/prism-contract`. Nothing in it knows what a component, a code standard, a job or a persona is.

**What is proven, exactly: two lenses, four routes** — Definitions (`/definitions`, `/definitions/:term`) and Standards (`/standards`, `/standards/:iri`) — plus the frame and the sitemap.

⚠️ **This is narrower than R.01 §8's "the four lenses", and deliberately so.** Components is pragma-specific by ruling. Home is *provably* not neutral: its committed `LobbyQuery` carries `... on Component { uri _meta { curie } name }`, and POSTing that operation text to the metro endpoint returns `Unknown type "Component".` It also links every exemplar to `/components/:uri`, a route metro cannot serve. Home is asserted for its **frame only**, and no artifact here claims otherwise. **R.01 §8 needs an amendment; this PR does not make it.**

### Why this exists when the acceptance gate already passes

`packages/docsite/graph-example`'s gate executes the app's four operation texts against the example provider but **never renders anything**. A query can be neutral while the components reading it are not — a `uri.startsWith`, a hard-coded prefix, a lens that assumes pragma's shape. This is the only check in the repo that runs the real app end to end against a foreign graph.

### The binding seam

`#lib/graphBindings` gains a **closed table** of named deployments and one selector reading `VITE_PRISM_DEPLOYMENT`. Three alternatives were rejected with reasons:

| Option | Verdict |
| -- | -- |
| Vite `resolve.alias` in the harness | **Cannot work.** It reaches the Vite SSR and client registries but not the **native** prepare step, so prepare warms `cs:CodeStandard` while the renderer asks for `metro:Station` — and it fails *quietly*, rendering the Suspense fallback into SSR HTML. |
| `--conditions` on package `imports` | Needs setting in three places across two runtimes. Fragile. |
| SSR-transported bindings | Principled and divergence-proof by construction, but touches six modules and drags R.01 §4's move 3 into this PR. **Recorded in the header as the fix when a real deployment appears.** |
| **Closed table + env selector, defaulting to `pragma`** | **Shipped.** |

It names a **table**, never a class, so a typo yields the default rather than a wrong binding; an unset variable gives byte-identical behaviour to today, so every existing test passes untouched.

The header's "NOT AN ENV VAR, DELIBERATELY" paragraph was **rewritten, not deleted**. Its reasoning turned out to be overstated for the runtime the proof uses — measured, Vite dev injects `VITE_*` straight from `process.env` into the client module, so all three registries read the same value in `dev:bun`. The failure mode is real but specific to a **built** client that did not see the variable at build time; the header now says so, and names the residual build-in-A/run-in-B hazard rather than pretending it away.

### The cross-registry pair — what makes this a proof and not a smoke test

`#lib/graphBindings` is instantiated in **three independent module registries**: Bun native (prepare), Vite SSR (render), and the client bundle. If prepare and render disagree, the store fills while `useLazyLoadQuery` misses and the page serves `Loading the standards…` with a full record map underneath. So every data-bearing route asserts **both** the rendered HTML and the embedded store.

The anneal found that the store half was weaker than it looked, and it is fixed here (`ee8574b`). The original check stringified the whole record map looking for `metro:` — but Relay formats root storage keys from the operation's own **arguments**, so on three of four routes the substring was satisfied by the *request*, whether or not any data came back. It is now two separately-named answers:

- **`askedForMetro`** reads `client:root`'s storage keys, so it still sees the argument the *Bun-native prepare step* chose. That is the cross-registry detector, and it is the only thing in the file that sees that registry.
- **`holdsMetroRecords`** checks for record IDs in `https://metro.example/`. Record IDs are `uri`, the absolute IRI (R-2), and the app only ever *sends* the prefixed CURIE — so such a key can only have been minted by the provider's response. **No request can forge it.**

Verified by mutation: pointing the namespace constant at an absent host turns the proof red, so the identity keys are genuinely present rather than the assertion being vacuous.

### Drive-by: `servers.e2e.ts` was already stale at HEAD

It asserted `react-flow__node-term` / `react-flow__edge-path` in the `/definitions/:term` SSR HTML. `HierarchyWell.tsx` **retired React Flow** (`a612f25`, AV-364) and now emits `hierarchy-node-shell` / `hierarchy-edge` / `hierarchy-cluster-caption`. Measured against a live render: `react-flow` appears **zero** times. Nobody noticed because `grep -rn "test:e2e" .github/` returns nothing — the suite is in no workflow.

⚠️ **One assertion there needed more than a rename, and it wants a reviewer's eye.** `wellEdgeCount < wellNodeCount` was justified by *"each edge is one non-root class's superclass link"* — but AV-364's own commit message says the well **stopped being a tree**; semantic arcs are drawn per domain→range pair and can outnumber the classes. The count is now `hierarchy-edge-structural` only (the same quantity `HierarchyWell.ssr.tests.tsx:66` pins), with semantic edges asserted separately. **This is unverifiable in the agent container** (no refs cache). If CI goes red there, the fix is the floor, not the selector.

## QA

| Suite | Baseline | Final |
| -- | -- | -- |
| `pragma-docs` `test` | 95 files / 440 | **96 / 450** |
| `pragma-docs` `check` | biome 376, tsc clean | biome 379, tsc clean |
| `pragma-docs` `test:proof` | did not exist | **green** |
| `relay-compiler --validate` | 23 reader / 10 norm / 23 op text | identical |
| `graph-example` | 227, 100% | 227, 100% |
| `contract` / `pragma-provider` | 32 / 45 | 32 / 45 |
| `servers.e2e` (pragma cells) | 4 failed / 2 passed | 4 failed / 2 passed, **identical messages** — pre-existing, no refs cache |

`apps/react/pragma-docs/src/relay/schema.graphql` verified clean after every e2e run and at the end; census unchanged at 174 type + 6 interface + 1 enum + 2 directive = **183**.

The proof measured **10 SSR hits / 0 client hits** on the metro provider, asserted as `> 5` / `=== 0` plus a strict-delta check that a direct POST moves the client counter by exactly one — so a counter that never counts cannot pass.

**Reproduce it by hand:**
```bash
cd packages/docsite/graph-example && PORT=5199 bun demo/server.ts
cd apps/react/pragma-docs && VITE_GRAPHQL_URL=http://127.0.0.1:5199/graphql \
  VITE_PRISM_DEPLOYMENT=metro PORT=5198 bun run dev:bun
curl -s http://localhost:5198/definitions/metro%3AStation | grep -c hierarchy-node-shell
```

**CI note, stated plainly:** `test:proof` is added to `pr.yml` and the step is correct, but CI on this base is blocked by an unrelated Playwright defect — see #911, which merges `main`'s fix. **This step's first green run may come after that merge.**

**Known open, for the owner:** the nx edge added so the proof reruns when the provider changes is a real task-graph edge, not just an `affected` hint — `pragma-docs:check` now runs 49 tasks (45 cached). The alternative leaves the proof blind to provider changes. Both directions cannot be closed without a cycle.

### PR readiness check

- [x] PR should have one of the following labels: `Feature 🎁`.
- [x] All packages define the required scripts in `package.json` — unchanged; `pragma-docs` gains `test:proof` alongside its existing `check`/`test`.
- [x] **New package: N/A.** No package is introduced.
- [x] `no visual change` — no component or style file is touched; the diff is the binding table, the e2e harness, two e2e specs, the demo server's request log, and CI config.

## Screenshots

Not applicable — no rendered output changes. The proof asserts on server-rendered HTML, and its assertions are the artifact.

---
_Generated by [Claude Code](https://claude.ai/code/session_019VVct8koTvTWrCzABQp3i2)_

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -30,6 +30,16 @@ jobs:
       - name: Test
         run: bunx nx affected -t test --base=origin/${{ github.event.pull_request.base.ref }} --head=HEAD
 
+      # The despecialisation proof: boot the docsite against
+      # `@canonical/prism-graph-example` and assert the core lenses render.
+      # A separate target from `test` because it spawns two real servers and
+      # takes ~30s — it has no business inside the unit run — and separate
+      # from `test:e2e` because that suite needs the pragma refs cache, which
+      # no runner has. `nx affected` is a no-op for projects lacking the
+      # target, so this step only ever runs the docsite.
+      - name: Despecialisation proof
+        run: bunx nx affected -t test:proof --base=origin/${{ github.event.pull_request.base.ref }} --head=HEAD
+
   build-gate:
     runs-on: ubuntu-latest
     needs: build-matrix
--- a/apps/react/pragma-docs/package.json
+++ b/apps/react/pragma-docs/package.json
@@ -35,6 +35,7 @@
     "test:watch": "vitest",
     "test:coverage": "vitest run --coverage",
     "test:e2e": "vitest run --config vitest.e2e.config.ts",
+    "test:proof": "vitest run --config vitest.e2e.config.ts metro.e2e",
     "storybook": "storybook dev -p 6010 --no-open --host 0.0.0.0"
   },
   "dependencies": {
@@ -85,5 +86,10 @@
     "vite": "^8.0.1",
     "vite-plugin-relay-lite": "^0.12.0",
     "vitest": "^4.0.18"
+  },
+  "nx": {
+    "implicitDependencies": [
+      "@canonical/prism-graph-example"
+    ]
   }
 }
--- a/apps/react/pragma-docs/src/lib/graphBindings/deployments.tests.ts
+++ b/apps/react/pragma-docs/src/lib/graphBindings/deployments.tests.ts
@@ -1,0 +1,101 @@
+/**
+ * The selector's fail-safe rule, and the key-set contract every deployment
+ * has to satisfy.
+ *
+ * WHAT WOULD BREAK WITHOUT THESE. `GRAPH_BINDINGS` is resolved once at module
+ * scope from an environment variable, in three independent module registries
+ * (Bun native prepare, Vite SSR render, the client bundle). The consequence of
+ * getting the rule wrong is not an exception — it is a lens quietly querying
+ * the wrong class, or a store that does not fulfil the operation the renderer
+ * runs. `selectDeployment` takes its name as an ARGUMENT precisely so that
+ * rule can be pinned here rather than discovered in a browser.
+ *
+ * The specific pins on the `metro` table are dataset facts, not preferences:
+ * they are the reason `test/e2e/metro.e2e.ts` can assert what it asserts, and
+ * changing either one silently weakens that proof.
+ */
+
+import { describe, expect, it } from "vitest";
+import {
+  DEFAULT_DEPLOYMENT,
+  DEPLOYMENT_ENV_VAR,
+  DEPLOYMENTS,
+  selectDeployment,
+} from "./deployments.js";
+
+describe("selectDeployment", () => {
+  it("falls back to pragma for an unset, empty or unknown name", () => {
+    // FAIL-SAFE is the whole design. A misspelt variable must degrade to the
+    // shipped default, never to an empty page — which is what a lookup that
+    // returned `undefined` and got spread into a query variable would give.
+    for (const name of [undefined, "", "pragmaa", "metro "]) {
+      expect(selectDeployment(name)).toBe(DEPLOYMENTS.pragma);
+    }
+  });
+
+  it("selects the named deployment", () => {
+    expect(selectDeployment("metro")).toBe(DEPLOYMENTS.metro);
+    expect(selectDeployment(DEFAULT_DEPLOYMENT)).toBe(DEPLOYMENTS.pragma);
+  });
+
+  it("names its own default among the deployments it can select", () => {
+    // `DEFAULT_DEPLOYMENT` is a string, so nothing stops it drifting away
+    // from a real key — after which the "fall back to the default" arm would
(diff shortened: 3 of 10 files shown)


advl:
Closing — **the substance of this PR is already on `feat/prism-docsite`**, landed through the #908/#910 squashes. Judged against the branch this targets rather than against its stale base, the whole PR reduces to four files, +36/−11:

- `packages/docsite/graph-example/demo/server.ts` — the per-hit `[graphql] http hit #N ssr|client` logging (~25 lines). **This is the only unlanded substance**, and it is genuinely useful: it is what lets an e2e harness prove the pages were filled by SSR hits rather than by client JS. It is not lost — it is recorded as a parked item in `canonical/pragma-adrs` `session/B/B.04.EXECUTION_STATE.md`, with this PR named as the place to lift it from.
- Three `package.json` files pinning `@canonical/biome-config`/`typescript-config`/`webarchitect` back to `^0.32.0`. Against the current branch these are **downgrades**, not contributions — the branch has since moved to `^0.33.0`.

The acceptance gate itself — `packages/docsite/graph-example/src/testing/integration/lensOperations.test.ts`, the falsification criterion that scans the lens directory and runs every core lens operation against a provider that has never heard of pragma — **is on the branch and green**. The proof this PR was opened to deliver is delivered.

The base branch (`claude/prism-provider-and-proof`) has no PR of its own and no landing path, which is why the diff shown at the top of this page reads as +1194/−74: it is measured against a fork point that predates two squash merges. Leaving the PR open advertises an unlanded proof that is in fact landed, which misreads the board for anyone planning from it.

PR #936, opened 2026-08-17
feat(constitution): stricter definition of magic

### LINKED ISSUE #789: [RFC] Add a written, decidable definition of "magic" to Constitution VI
This rule was cited as a response to the claim I made recently and I understood that I, when encountering something, cannot confidently understand how “magical” it is.

# The problem

The [constitution](https://github.com/canonical/pragma/blob/main/CONSTITUTION.md) claims operational force: "The principles are not aspirational. They are operational. Code that violates a principle is a bug, not a style choice." An operational rule must be decidable. [Principle XI](https://github.com/canonical/pragma/blob/main/CONSTITUTION.md#xi-structured-data-over-prose) makes the same demand of our specifications: "Specifications that are precise enough for machines to reason about are precise enough for humans to trust."

[VI](https://github.com/canonical/pragma/blob/main/CONSTITUTION.md#vi-no-magic) states: "no conventions that require framework knowledge to interpret and no build-time transforms that alter what ships", and "what the developer writes is not what the browser runs ... The stylesheet in the source is the stylesheet in the browser."

Four constructs this codebase mandates fail that literal reading:

JSX/TSX (every React tier).
 `<Card />` ships as `jsx(Card, {})`. This is a build-time transform that alters what ships, and it is not erasable annotation.

React hooks. 
[Principle V ](https://github.com/canonical/pragma/blob/main/CONSTITUTION.md#v-composability) mandates them ("Shared behaviour lives in hooks") and we publish packages/react/hooks. Hooks bind component state to call order: they must be called unconditionally, at the top level, in the same order on every render. An early return placed above a `useState` produces a bug that cannot be diagnosed by reading the source. Diagnosing it requires knowing React's internal state-by-call-index model. This is the definitional case of a convention that requires framework knowledge to interpret.

Svelte runes, in svelte components:
`$props`, `$derived`, `$bindable` (and more). These are invented syntax compiled into reactive machinery. Nothing here is erasable or platform-native.

`import "./styles.css"`, the styling mechanism VI's own text endorses ("A component imports its stylesheet directly"). 
No browser executes a bare CSS side-effect import. It works only because a bundler discovers the edge in the import graph and extracts or injects the stylesheet. Measured by "what you write is what the browser runs", the endorsed pattern is itself a build-tool convention.

Nobody believes VI was intended to forbid these four. So VI's actual rule should be narrower than its text. But that rule is unwritten, and an unwritten rule cannot be applied symmetrically: any construct can be called magic, and no construct can demonstrate that it is not.

I tried to come up with a few candidates for the rule:

1. Literal:no build-time transforms that alter what ships
Result: Forbids JSX, runes, CSS imports and typescript. The React and Svelte tiers are unconstitutional.

2. Transforms permitted only when they erase cleanly (type stripping)
Result: Saves TypeScript. Still forbids JSX and runes. The Svelte and React tiers cannot exist.

3. The tier's chosen compiler is the accepted contract. No tools beyond it.
Result: 
 - JSX, hooks and runes pass.
 - Sass, PostCSS and CSS-in-JS remain forbidden (added tools, or runtime execution).
 - Svelte style scoping passes: same compiler, zero added tooling.

4. Transforms permitted when total, deterministic, documented, build-time and locally predictable.
Result: similar to 3

The first two readings render our codebase unconstitutional.
# Proposed amendment
Add to VI:

A transform is magic when its output cannot be predicted from the source being read. The platform and the tier's chosen compiler form the execution contract a contributor accepts by working in that tier. Their transforms are acceptable when they are total, deterministic, documented, applied at build time, and inspectable in their output, so that the reader predicts the shipped artifact from local source and documented rules. 

A transform is magic when:
- its effect depends on file placement, file naming, or configuration outside the source being read; when it is heuristic, so that identical source can behave differently by inference;
- when it executes at runtime; 
- when it requires a tool beyond the platform and the tier's chosen compiler. 

The burden of proof does not move: adding a compiler, or widening a tier's contract, remains an inclusion to be justified under [XIII](https://github.com/canonical/pragma/blob/main/CONSTITUTION.md#xiii-preference-towards-minimal-tooling).

MichaelKapustey:
Follows up the discussion in #789. While editing I found some contradictions between the amendment and the surrounding text, and did my best to resolve them following the spirit of that discussion.

Closes #789.

### DIFF
--- a/CONSTITUTION.md
+++ b/CONSTITUTION.md
@@ -86,11 +86,20 @@ The cost is more pieces, more files, and a steeper initial learning curve for de
 
 ## VI. No magic
 
-Software becomes difficult to maintain when its behaviour cannot be understood by reading the source. Frameworks routinely hide complexity behind conventions that feel productive until something breaks, at which point the developer must reverse-engineer hidden behaviour to diagnose the problem. This trade-off---convenience now, opacity later---is rejected.
+Software becomes difficult to maintain when its behaviour cannot be understood by reading the source. Hidden machinery feels productive until something breaks, at which point the developer must reverse-engineer behaviour that no documented rule explains. This trade-off---convenience now, opacity later---is rejected.
 
-Every import is explicit. Every export is listed in a barrel file. Configuration files are plain data. The behaviour of the system is visible in the source, with no conventions that require framework knowledge to interpret and no build-time transforms that alter what ships. A component imports its stylesheet directly rather than relying on a build tool to discover and inject it by naming convention. Package dependencies appear in the manifest rather than being resolved through workspace hoisting.
+A construct is magic when its behaviour cannot be predicted from the source being read. What makes prediction possible is the tier's execution contract: the platform, plus the framework and compiler the tier is explicitly built on---their transforms, their runtime, and the conventions they define and document. Choosing a framework for a tier is choosing this contract in full, and a contributor who works in the tier accepts it in full. The contract's own idioms are therefore not magic, however much framework knowledge they take to read: JSX compiling to render calls, hooks, runes, etc. Not knowing the rules of hooks is the lack of knowledge, not hooks being magical.
 
-The choice of pure CSS over CSS-in-JS or preprocessors follows from the same reasoning. Preprocessors and CSS-in-JS libraries introduce a transformation layer between authoring and execution---what the developer writes is not what the browser runs. Pure CSS eliminates this gap. The stylesheet in the source is the stylesheet in the browser. CSS custom properties provide the theming and composition capabilities that preprocessors once required, without a compilation step that obscures the relationship between input and output.
+The contract's transforms are acceptable because they are total, deterministic, documented, applied at build time, and inspectable in their output, so that the reader predicts the shipped artifact from local source and documented rules.
+
+Magic is what lies outside the contract. A construct is magic when:
+
+- it requires a tool, plugin, or transform the contract does not include---an addition the framework does not need, as CSS-in-JS or a preprocessor is to React;
+- its behaviour depends on file placement, file naming, or configuration the contract does not define and document;
+- it is heuristic, so that identical source can behave differently by inference;
+- it defers a transform to runtime, so that its output never exists as an inspectable artifact.
+
+The principle does not reject frameworks; it rejects unreadability. The current tiers accept deliberately small contracts---React is a rendering library, not a framework; Svelte is little more than a compiler---and every contract must be learned before the source becomes predictable. A bigger framework would not be more magical, only heavier: more conventions to accept, more machinery between the source and what runs. Big frameworks are kept out as complexity, not as magic. Introducing a framework, or widening a tier's contract, is an inclusion to be justified under [XIII](#xiii-preference-towards-minimal-tooling).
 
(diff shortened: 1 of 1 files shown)


stavropouloskonstantinoss (commented):
According to section VII there should not be an invisible coupling between code and framework internals. The conventions arise from the code and should not be inherited from the framework. For this reason, I would suggest not modifying the definition.

CONSTITUTION.md
```diff
@@ -86,11 +86,20 @@ The cost is more pieces, more files, and a steeper initial learning curve for de
...
 
-The choice of pure CSS over CSS-in-JS or preprocessors follows from the same reasoning. Preprocessors and CSS-in-JS libraries introduce a transformation layer between authoring and execution---what the developer writes is not what the browser runs. Pure CSS eliminates this gap. The stylesheet in the source is the stylesheet in the browser. CSS custom properties provide the theming and composition capabilities that preprocessors once required, without a compilation step that obscures the relationship between input and output.
+The contract's transforms are acceptable because they are total, deterministic, documented, applied at build time, and inspectable in their output, so that the reader predicts the shipped artifact from local source and documented rules.
+
+Magic is what lies outside the contract. A construct is magic when:
+
+- it requires a tool, plugin, or transform the contract does not include---an addition the framework does not need, as CSS-in-JS or a preprocessor is to React;
```
advl: ` as CSS-in-JS [...] is to React;` : I think this is more than a react question but is an anti pattern across web engineering. The preprocessor is also unrelated to React; I wouldn't say it is an antipattern or magic tough - just a technology that has been absorbed into the css language itself.

advl (commented):
Thank you for the contribution. It's a clear improvement over the existing.

I left a minor comment, also agree with @stavropouloskonstantinoss

MichaelKapustey:
> According to section VII there should not be an invisible coupling between code and framework internals. The conventions arise from the code and should not be inherited from the framework. For this reason, I would suggest not modifying the definition.

I stated the need for modifying the definition in this issue @stavropouloskonstantinoss https://github.com/canonical/pragma/issues/789. Long story short, without modifying it, magic captures a lot of things were already doing, for example react hooks are magical and therefore forbidden. Lets have a targeted discussion of what you don't like in the proposed definition and improve it together.

CONSTITUTION.md
MichaelKapustey: I tried to provide an example of something extra added to the tier. Would you like to have it removed or rephrased, maybe without mentioning React. In my opinion examples work best when they are specific

advl: Both approaches can work, I'll leave it to your preference ! I was just trying to point at why I thought those examples in specific were to me inaccurate. We could either use other examples or none.

advl:
@MichaelKapustey I think @stavropouloskonstantinoss  were mentioning that there is an edit to VII on top of the edit to VI. VI is in scope of #789 but VII should not be. I also agree with this

stavropouloskonstantinoss:
@MichaelKapustey Thank you for your response. Indeed, I should have made my wording clearer, my suggestion to not modify the definition is aimed to section VII. Your proposed modifications to VI captures the essence of #789, which is about improving the definition of magic.

advl:
@MichaelKapustey gentle bump :)

MichaelKapustey:
@advl updated

PR #951, opened 2026-08-22
ci: the publish lane builds the pragma binary and proves what it published

advl:
## Done

- **The publish lane now provably builds the compiled `pragma` binary.** `@canonical/pragma-cli` gains the `build:all` script every other `packages/{cli,runtime,summon}` package already had — it was the only build-step package missing it, so `setup-env`'s `bunx lerna run build:all` skipped it in every lane and `lerna publish from-package` shipped whatever `dist/pragma` happened to exist (nothing depends on the CLI, so Nx's `^build` never reached it either).
- **Post-publish smoke** (`tag.yml`): installs the just-published `@canonical/pragma-cli@<version>` from the registry in a fresh temp dir (5 propagation retries), asserts `pragma --version` reports exactly the released version, then runs `pragma doctor` — chosen empirically because it exits 0 outside a project, unlike the store-dependent verbs (exit 3).
- **Post-publish status & provenance gate** (`tag.yml`): prints the `bun run publish:status` table, then fails the workflow unless every public package serves its released `name@version` from the registry with a provenance attestation. The direct query exists because `publish:status` cannot gate: it always exits 0 and compares against the `latest` dist-tag, which pre-releases deliberately never move.
- **`docs/CI.md` tells the truth again**: the tag-lane section now describes the real five jobs and the two new verification steps, and the `setup-env` paragraph no longer claims the action "does not build" (its final step has run the `build-command` input — default `bunx lerna run build:all` — all along).
- Drive-by: `@canonical/summon-application` and `@canonical/react-ds-app` declare `publishConfig: { "access": "public" }` like every other public package. Both are already public on the registry, so this is load-bearing only at a future first publish of a renamed/split package.

## QA

- YAML re-parsed cleanly (`js-yaml`), all three new `run` scripts pass `bash -n`, touched manifests parse, biome clean per-package.
- The smoke invocation was executed verbatim against the live registry for the already-published `0.33.0`: install succeeded, the version assertion passed, and `doctor` via `npm exec` exited 0 in an empty directory (also with scrubbed `HOME`/XDG to simulate a fresh runner).
- The provenance field path was validated live: `@canonical/pragma-cli@0.33.0` reports predicateType `https://slsa.dev/provenance/v1`; a nonexistent version trips the FAIL path.
- End-to-end proof comes from the next `tag.yml` dispatch of any release type — the two new steps gate it.

### PR readiness check

- [x] PR has one of the required labels: `Maintenance 🔨` (plus `no visual change`).
- [x] All packages define the required scripts — this PR closes the repo's one remaining gap on that very checklist item: `@canonical/pragma-cli` was the only build-step package without `build:all`.
- [x] No new package is introduced.
- [x] `no visual change` label added to skip Chromatic.

## Screenshots

Not relevant — CI/workflow and manifest changes only.

https://claude.ai/code/session_017EwUvEW5ZgjAYc9KWPjHCZ

---
_Generated by [Claude Code](https://claude.ai/code/session_017EwUvEW5ZgjAYc9KWPjHCZ)_

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -146,4 +146,72 @@ jobs:
             DIST_TAG="latest"
           fi
           echo "Publishing $VERSION to dist-tag '$DIST_TAG'"
+          echo "DIST_TAG=$DIST_TAG" >> "$GITHUB_ENV"
           lerna publish --yes --no-private --dist-tag "$DIST_TAG" from-package
+
+      - name: Smoke the published CLI
+        # `lerna publish` exiting 0 proves the tarballs uploaded — not that the
+        # compiled `pragma` binary inside @canonical/pragma-cli was built from
+        # this tag and actually runs. Install the just-published version from
+        # the registry in a fresh directory and exercise it, retrying while the
+        # registry propagates. `doctor` is the functional probe because it
+        # exits 0 outside a project (store-dependent verbs exit 3 there).
+        env:
+          VERSION: ${{ needs.version.outputs.VERSION }}
+        run: |
+          echo "Smoking @canonical/pragma-cli@$VERSION (published to dist-tag '$DIST_TAG')"
+          cd "$(mktemp -d)"
+          for attempt in 1 2 3 4 5; do
+            if REPORTED="$(npm exec --yes "@canonical/pragma-cli@$VERSION" -- --version)"; then
+              break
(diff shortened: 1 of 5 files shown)


advl:
Won't merge - we can't have the CI become too specific for one particular package. It's an antipattern

PR #953, opened 2026-08-23
feat(ke-graphql)!: adopt the graphql: annotation vocabulary and remove the config layer

advl:
## Done

Nine commits bringing the complete modern ke-graphql to `main` as a package-specific PR — the annotation vocabulary, every review-hardened compiler fix, a standards-mandated rename, and the full removal of the deprecated config surface. Verified by an adversarial review loop that ran until a round confirmed zero findings (arc: 21 → 3 → 3 → 0, all post-sync rounds being documentation-truth residue).

**What arrives (previously only on the feature lineage):**
- **The 13-term `graphql:` annotation vocabulary** — ontology authors declare projection intent (names, cardinality, exposure, descriptive sources, prefixes) on the ontology terms themselves; A-band diagnostics (`A001`–`A008`, `B005`), three projection modes (`auto`/`annotated`/`explicit`), descriptive binding with ancestor inheritance. Namespace: `https://pragma.canonical.com/graphql#`.
- **The hardened compiler**: explicit cardinality outranks per-class SHACL (in the mapper and the TBox browser); `graphql:prefix` binds where the mode is known — under `auto` the overlay is genuinely not consulted; one prefix authority (`resolveEffectivePrefixes`); mode-true `E001` via a Pass-1 → Pass-2 deferral; interface-typed values keep annotated `*From` heads; `_meta.label` is asserted-only; `annotations.ts` de-binarized (it carried two raw NUL bytes — git treated the module as binary).

**Breaking changes (the `!`):**
- **The config-mapping surface is removed entirely**: `CustomMapping`/`CustomMappings`, the `mappings` option, `nonNullOverrides`, `standardVocabFields` — implementation, tests, and docs. Diagnostics **`A005`** (config-shadows-annotation) and **`M003`** (mapping-references-nothing) are retired with their subject (documented in the code table, `X001`-precedent). Declared inverse pairs (`owl:inverseOf`, `graphql:inverse`) are untouched; only the config-only synthetic-inverse form dies. `CardinalitySpec.source` drops the now-unproducible `"custom"`. **No consumer in this repository referenced the removed API** (verified; the committed benchmark was the one straggler and is repaired in this PR — it now drives its rename through `graphql:name` in its own TTL and runs to completion).
- `RawExtraction.deferredSyntheticNamespaces` is new and **optional** at the public boundary.

**Deliberately not done — `ARTIFACT_VERSION` stays 1**: on `main` no released compiler ever folded declarations in Pass 1 (zero vocabulary awareness), so no existing artifact carries the pre-fold shape that would need invalidating; a bump would only break the CLI's committed embedded pack for no protection. The version rule is now documented at the constant: if binding ever returns to Pass 1, the version must rise.

**Rename (code standards):** `effectivePrefixes` → `resolveEffectivePrefixes` per `cs:code.naming.function_verb`; the remaining new-export surface was audited against the rule (only violation found).

**Rebased onto current `main` (`fa2816ae`, 2026-08-30).** The branch had gone un-mergeable: 80 commits of `main` landed under it, conflicting on `bun.lock` and `packages/runtime/ke-graphql/package.json`. Resolved by taking main's side on both — its release bump to `0.36.0` strictly subsumes this branch's own `@canonical/ke` peer-floor raise to `^0.34.0`, so nothing this branch intended was lost. Re-verified after the merge: **39 files / 548 tests green, coverage 100/100/100/100, `check` green.**

## QA

- ke-graphql: `check` green (biome, package-structure, license); `test` **548 passed (39 files)**, coverage **100/100/100/100**, no coverage ignores — re-run after the `main` merge
- All 8 SDL byte-goldens **identical** across the entire branch — default emissions untouched
- `bun run bench` runs to completion and reproduces the README's timing table (compile ~68ms live, ~10ms artifact boot on the reference run)
- `@canonical/pragma-cli:check` **passes**
- Review loop: multi-lens Opus review + per-finding adversarial verification, repeated to a zero-confirmed round

## ⚠️ Known blocker — owner decision pending

**`pragma-cli:test` fails from exactly one root cause**: the anatomy-dsl ontology declares its own `uri` property, colliding with the structural `uri: ID!`. Every failure carries the same diagnostic:

```
M005: http://anatomy-dsl.example.org/ontology#uri maps to NamedNode.uri,
      a structural field the compiler owns — the field is DROPPED.
```

On `main` today the old compiler **silently mis-handles** this (last-value-wins — a latent correctness bug); the hardened compiler raises fatal `M005` and refuses, which is the diagnostic working as ratified. The old escape hatch (`mappings`) is removed by design.

**Post-rebase the count grew, the cause did not.** It was 12 failures before the `main` merge; it is now **19 failed tests plus 5 failed suites** (27 entries). The increase is entirely because `main` added more consumers of the embedded pack — the new `block.shipped.exec`, `block.tierRank.exec` and `listLookup.shipped.exec` parity suites — each of which boots the same colliding graph and correctly refuses it. No second cause appeared, and ke-graphql's own suite is fully green.

**The remedy, in order** (each step assumes the previous):

1. Annotate the upstream ontology — `graphql:name "anatomyUri"` on the offending property in `advl/anatomy-dsl` (`definitions/ontology.ttl`, ≈ line 133) and in the hermetic test corpus.
2. Regenerate the CLI's embedded pack. Two traps: `bun run bundle` alone **writes nothing** (its skip cannot see a toolchain change with unchanged inputs — delete `pack.generated.ts` first), and both packs are declared at `#main`, which has since moved, so pin the sources to the SHAs the committed manifest asserts unless you intend to fold upstream drift into the diff.
3. Delete both stopgaps — `pragma-provider`'s `constants.ts:113-118` and the CLI's `graphpack/read.ts:56-57` — and invert the test that currently pins "this kills the boot". Fix the stale comment at `read.ts:52` while there.

This PR deliberately ships without weakening the diagnostic or hand-editing generated data.

### PR readiness check

- [x] PR should have one of the following labels: **`Breaking Change 💣`** (no label API access from this session — please apply)
- [x] All packages define the required scripts in `package.json` — no script set altered.
- [x] **New package: N/A** — changes are inside the existing `@canonical/ke-graphql`.
- [x] No visual changes — `no visual change` label applies.

## Screenshots

Not applicable — compiler, tests, and documentation only.


advl:
**Atomicity judgment (explicit-atomicity law): this PR stays one unit. Recording why, rather than splitting it.**

Measured, it looks like a violation — +9687/−1069 across 80 files is roughly twice the top of the 500–5000 band, and the title names two acts. The commit history separates cleanly, too: `224cf1029` is the vocabulary adoption (78 files, +10129/−599) and `ca4c66b4f` is the config-layer removal (30 files, +246/−1207).

It is not split, for three reasons:

1. **The removal is entailed by the adoption, not independent of it.** `ca4c66b4f`'s own subject says it: the `graphql:` vocabulary is the transport for projection intent, and the config-side shadow it replaces is deleted whole. Landing the removal alone would take away `mappings`, `nonNullOverrides` and `standardVocabFields` with no replacement in the tree; landing the adoption alone would ship two authorities for the same knobs — two places to look and two ways to disagree — which is the state the owner ruled against when deprecated knobs were declared not-to-be-maintained.
2. **Splitting does not bring the large half into band.** The vocabulary adoption is 10k lines on its own. The split would produce one PR still ~2× over the band and one trivial 1.4k companion, paying two re-merges of `main` (this branch already carries the `fa2816ae` merge) to move nothing across the line that matters.
3. **The band's purpose is served another way here.** The band exists so a review swarm has a fixed target and a fix agent knows what it may touch. This branch's own suite has been green throughout (39 files / 548 tests at the merge), and its failures have had exactly one root cause the whole time.

If a reviewer disagrees, the split point is `ca4c66b4f` and it is clean — that is recorded here so the option stays open rather than being rediscovered.

---

**Blocker status.** The single root cause is unchanged: the anatomy-dsl ontology declares its own `uri`, colliding with the structural `uri: ID!`, and the hardened compiler correctly refuses with fatal `M005`.

The upstream half is now open: **[canonical/anatomy-dsl#5](https://github.com/canonical/anatomy-dsl/pull/5)** annotates `anatomy:uri` with `graphql:name "anatomyUri"` — verified parsing to 188 triples with the annotation reading back correctly, suite green. Two notes that correct the record here: the repo is **canonical/anatomy-dsl**, not `advl`, and pragma consumes the pack from `git+ssh://…#main`, so **no npm release is needed** — once #5 merges, the next embedded-pack regeneration picks it up. Then the two stopgaps (`pragma-provider`'s `constants.ts` and the CLI's `graphpack/read.ts`) come out and the test that pins "this kills the boot" inverts.

advl:
**Correction to my earlier comment on this PR.** I said pragma consumes the anatomy pack from `git+…#main`, so no npm release would be needed once [canonical/anatomy-dsl#5](https://github.com/canonical/anatomy-dsl/pull/5) merged. That was wrong.

#5 is merged, and I regenerated the embedded pack expecting `M005` to clear. It did not. Four hypotheses were refuted before the real cause appeared, and the last one is the useful part:

- **`auto` mode skipping the overlay** — refuted, `DEFAULT_MODE` is `"annotated"`.
- **A stale materialised clone** — refuted, a clean-cache run failed identically.
- **A `graphql:` namespace mismatch across the two lineages** — refuted, both read `https://pragma.canonical.com/graphql#`, matching the annotation.
- **The annotation not working** — refuted decisively. Compiled standalone, and again alongside the design-system ontology, it extracts as exactly one row and **no `M005` is raised**. The rename works.
- **The bundler not reading the declared git source** — ✅ **this is it.** `scripts/embedSources.ts` defines `SOURCE_OVERRIDES`, which resolves `@canonical/anatomy-dsl` from **npm** — `node_modules`, pinned in devDependencies — because the git remote is unreachable from the machine that builds the embed. The installed pin is **0.2.2**, published before the annotation existed.

`pragma.conf.ts` declaring a git source is not wrong; it is the correct source *for a real user*. I reasoned from the declaration instead of from the build, and the override is exactly the kind of thing that punishes that.

Worth noting the override is well-guarded rather than a hack: the release parity gate reads the same table, so the bundler and the gate cannot disagree about which provenance is legitimate. Which also means working around it locally would produce an embed the gate rejects — the honest path is a release.

**The remaining chain, each step assuming the one above:**

1. **Publish `@canonical/anatomy-dsl` 0.2.3.** A reserved act — it needs release credentials, and publishing is outward-facing in a way merging is not.
2. Bump the `@canonical/anatomy-dsl` devDependency pin in `packages/cli/pragma`.
3. Regenerate the embedded pack — proven to work; it fails today only because it is fed 0.2.2.
4. Delete both stopgaps and invert the test pinning "this kills the boot".

So this PR is still blocked on an act rather than on engineering — the same shape as before, one layer further out. Recorded in `canonical/pragma-adrs` `session/B/B.04` §9.4a, including the refuted hypotheses so the search is not repeated.

advl:
Superseded by #1070

PR #959, opened 2026-08-24
wip / POC: Cache-bust icons & add support for consumer-defined icons

jmuzina:
## Done

Draft / POC proposal discussed in React/Svelte WG that attempts to work towards a synthesis of the [Launchpad & core DS approaches to icons](https://docs.google.com/document/d/10UcrQu2Iu6u-Nep9EZQKZs9eYD26S8fs8DNwkxx70iU/edit?tab=t.0). 

**This is a POC**, mostly the result of me spiitballing on the problem with Opus. Don't expect perfect code quality yet, I'm moreso interested in thoughts on the approach, & if it is validated I will advance it to a more mature state.

### Cache busting

Currently [Pragma's React `Icon` component](https://github.com/canonical/pragma/blob/e8d25fecbe6e70edfc4cb300bad65f691dbbd26e/packages/react/ds-global/src/lib/component/Icon/Icon.tsx#L18-L50) binds glyph `use.href` to an SVG sprite assumed to be served at the application's `/icons/<iconName>.svg`. This forces additional configuration on every consumer of the Icon in order to avoid serving stale assets after an Icon is updated. 

This PR takes the belief that it is Pragma's responsibility to provide an easy way to solve this problem without requiring additional setup. `ds-assets` now has a publicly exported function `buildAssetManifest` that, given a directory, hashes the files in that directory and returns (and optionally writes to a file) a key-value map from file name to file name with hash included. This manifest is consumed by `Icon` to determine the `href` for the Icon sprite, so when the underlying asset changes, users get the new icon.

### Custom Icons

Pragma's Icon component doesn't support consumer-defined icons. It [requires all icons to be sourced from the central `ds-assets` package](https://github.com/canonical/pragma/blob/e8d25fecbe6e70edfc4cb300bad65f691dbbd26e/packages/react/ds-global/src/lib/component/Icon/types.ts#L16) which presents problems for downstream projects that may want to use an icon that is not in `ds-assets` (and potentially is too project-specific to ever be in `ds-assets`).  

I added support for consumer-defined Icons. The user can pass their icons through the same `buildAssetManifest` function from ds-assets that is used for the official icons. This custom manifest can be passed into `Icon` to allow it to use custom icons, in addition to the global icons.

## QA

- Clone the PR
- `bun run build` across the entire repo
- `cd packages/react/ds-global`
- `bun run storybook`
- Open Icon stories and verify icons are all still rendered successfully. 
- Open new Custom Icons story and verify the custom icons are also rendered successfully. 
- Using dev tools, inspect two icons of your choice. Note down their `href`s. Change one icon's svg source code somehow, then re-run `bun run build` across the repo and restart storybook. Verify that both icons are still rendered, and that only the changed icon's `href` has changed.

### PR readiness check

- [ ] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [ ] The code follows the appropriate [code standards](https://github.com/canonical/web-code-standards)


packages/react/ds-global/src/lib/component/Icon/types.ts
```diff
@@ -8,12 +8,37 @@ import type { SVGAttributes } from "react";
...
+   * Use this to add custom icons alongside `ds-assets`' own: merge your own
+   * entries in (`{ ...ICON_MANIFEST, "my-icon": "my-icon.<hash>.svg" }`) and
+   * self-host the matching files under `rootPath`. An icon missing from the
+   * override falls back to `ICON_MANIFEST`; an icon missing from both falls
+   * back to the plain `<icon>.svg` naming — functional, but not cache-safe,
+   * so a custom icon you care about invalidating correctly needs its own
+   * manifest entry — hash it with `buildAssetManifest` from
+   * `@canonical/ds-assets/build` to match `ICON_MANIFEST`'s own scheme.
+   */
+  manifest?: Record<string, string>;
```
jmuzina, line 43: I'm not yet sure if I'm on board with this approach of embedding a manifest (a quite logistical concern) in the props of the component. The only reason it is needed is to add support for consumer-provided icons - see the custom icons story.

Maybe an improvement that could be made to remove this prop is to add a `CustomIconProvider` higher up in the tree so this can be configured once at the application level, rather than at the component level. Food for thought, open for opinions

Other options: 

- [Launchpad-ui approach](https://github.com/canonical/launchpad-ui/blob/main/src/lib/components/icons/BotIcon.svelte) of defining an `IconBase` to handle deduplication of icon fetches, and separately bundle each component rather than using a single `Icon` component - meaning consumers could trivially implement their own Icons that use this `IconBase`. We also would avoid the need for explicit asset manifest creation, as the bundler would handle hashing / hash resolution for us. 
- Allow an escape-hatch that just renders `children` so arbitrary content can be rendered, leaving the responsibility of proper markup to the consumer. This is very risky and kind of a non-answer from a systems POV.

packages/ds-assets/docs/ICONS.md
```diff
@@ -57,6 +59,54 @@ interface Props {
...
+```
+
+Then merge it into `ICON_MANIFEST` when passing `manifest` to `Icon`:
+
+```tsx
+import { ICON_MANIFEST } from "@canonical/ds-assets";
+import customManifest from "./custom-icons/manifest.generated.js";
+
+<Icon icon="my-custom-icon" manifest={{ ...ICON_MANIFEST, ...customManifest }} />
+```
```
cadencehagenauer-canonical, line 106: It could be nice to have a central per-project place to do this merging, so it doesn't need to happen in every file where custom icons are needed. That's not hard for a project to do by itself, but a standardized approach could make working across multiple projects easier.

packages/ds-assets/src/build/buildAssetManifest.test.ts
```diff
@@ -0,0 +1,149 @@
...
+  roots.push(root);
+  for (const [fileName, contents] of Object.entries(files)) {
+    writeFileSync(join(root, fileName), contents);
+  }
+  return root;
+}
+
+function hashOf(contents: string, length = 8): string {
+  return createHash("sha256").update(contents).digest("hex").slice(0, length);
+}
```
cadencehagenauer-canonical, line 33: Probably a good idea to define this elsewhere to be used by the tests and the manifest script in case the hashing function needs to be changed.

packages/ds-assets/docs/ICONS.md
jmuzina: Yeah, I agree - see [comment](https://github.com/canonical/pragma/pull/959#discussion_r3858694986), one way to do this centrally would be with a Provider

packages/ds-assets/src/build/buildAssetManifest.ts
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * @note Impure — reads `sourceDir` and writes into `outDir`.
+ */
+export function buildAssetManifest({
+  sourceDir,
+  outDir,
+  manifestPath,
+  hashLength = DEFAULT_HASH_LENGTH,
+}: BuildAssetManifestOptions): Record<string, string> {
+  mkdirSync(outDir, { recursive: true });
```
cadencehagenauer-canonical, line 53: Might be good to check for dir's existence. Very unlikely case, but on Windows this could error even with recursive being true if targeting the root directory.

https://nodejs.org/api/fs.html#fsmkdirpath-options-callback

packages/ds-assets/src/build/buildAssetManifest.ts
```diff
@@ -0,0 +1,94 @@
+import { createHash } from "node:crypto";
+import { mkdirSync, readdirSync, readFileSync, writeFileSync } from "node:fs";
+import { dirname, join, parse } from "node:path";
+
+/** Matches the default hash length bundlers (Vite, webpack) use for their
+ * own emitted assets — short enough to keep filenames readable, long enough
+ * that collisions are not a practical concern for a directory this size. */
+const DEFAULT_HASH_LENGTH = 8;
```
cadencehagenauer-canonical, line 8: With 8 hex values (32 bits), there is a 50% chance a collision will occur after around 65536 files have been hashed. 

There is also around a 1% chance for a collision to occur within 10,000 files being hashed.

This seems like it would probably be enough for most projects.

packages/react/ds-global/src/lib/component/Icon/types.ts
cadencehagenauer-canonical: I think it would be good to avoid embedding the manifest at the component level. 

Something like a CustomIconProvider could be nice.

Taking a look at the IconBase approach, I don't love the extra boilerplate it expects users to create for each icon they want to use. Maybe it could be improved by automating it? Though that could cause IDE/convenience issues depending on when those components are generated.

cadencehagenauer-canonical (commented):
I think this approach generally makes sense! The main decision that needs to be made still is how to centralize/standardize manifest creation. The core approach seems strong and QA worked without issue!

jademathre-canonical:
I'm concerned that this seems to trade stale icons for missing icons. The hashed names are baked into react-ds-global's JS, but the files come from whatever ds-assets the app happens to copy. With ^0.x ranges, you could end up with two copies in node_modules and then every changed icon 404s. We're protecting against a slightly old glyph by risking displaying a blank box.

It also only covers Icon. Spinner and the CSS-referenced icons (Accordion caret, checkboxes) still request plain /icons/x.svg, so an app has to serve both icons/ and dist/icons/, while currently we serve only icons/. 

On custom icons, I agree with your own hesitation about the prop, and I'd add: widening icon to string loses the typo-catching we get from IconName today, and A custom icon would work in a bare `<Icon> ` but not in a something like a Button.

Alternatively a single content hash of the icon set, appended as ?v= to every icon URL. No renamed files, nothing extra to copy, skew degrades gracefully, and a build-time stamp applies the same thing to the CSS references. For custom icons: an augmentable IconName plus an IconProvider that can set the root path and custom-icon URLs once, so <Button icon="custom-branding"> just works.

packages/ds-assets/src/build/buildAssetManifest.ts
```diff
@@ -0,0 +1,94 @@
...
+      .sort(([firstName], [secondName]) => firstName.localeCompare(secondName))
+      .map(([name, fileName]) => `  "${name}": "${fileName}",`)
+      .join("\n");
+    const manifestSource = `// AUTO-GENERATED by buildAssetManifest — do not edit by hand.
+const assetManifest = {
+${manifestEntries}
+} as const;
+
+export default assetManifest;
+`;
```
jacobbudin, line 87: Re the concatenation of strings to generate the object: Did you consider something like this?:
```ts
const sortedManifest = Object.fromEntries(
  Object.entries(manifest).sort(([a], [b]) => a.localeCompare(b)),
);

const manifestSource = `// AUTO-GENERATED by buildAssetManifest — do not edit by hand.
const assetManifest = ${JSON.stringify(sortedManifest, null, 2)} as const;

export default assetManifest;
`;
```

packages/ds-assets/src/build/buildAssetManifest.test.ts
advl: good catch

packages/ds-assets/docs/ICONS.md
```diff
@@ -57,6 +59,54 @@ interface Props {
 }
 ```
 
+## Self-hosting and cache invalidation
+
+Icon filenames are stable across releases: `search.svg` never changes name even when its glyph does. If you self-host `icons/` (e.g. copy it into your app's `public/` directory) and serve it behind normal long-lived asset caching, a `ds-assets` update can ship a fixed or redesigned icon and browsers/CDNs will keep serving the old one indefinitely — the URL never changed, so nothing tells them to refetch. This is a consequence of how you're serving the files, not something `ds-assets` or an `Icon` component can fix on your behalf — asset delivery and cache configuration are your app's responsibility.
+
+What `ds-assets` does provide is a way to make invalidation possible: `dist/icons/` ships each SVG under a **content-hashed filename** (`search.a1b2c3d4.svg`), alongside `ICON_MANIFEST`, which maps every icon name to its current hashed filename:
```
bartaz-canonical, line 66: Is there a reason to put hash in the file name `icon.hash.svg` instead of query param `/icon.svg?v=hash`, so that the files names are not affected and can be easily statically served as they are?

jacobbudin: Two cents: The current method (`icon.hash.svg`) prioritizes correctness over safety (`icon.svg?v=hash`). A failure to generate a hash in the filename is obvious and more likely to lead to the asset not loading, whereas`?v={""|"null"|"undefined"|stale_hash|wrong_hash}` serves as a fallback but at the cost of a bug going unnoticed.

packages/react/ds-global/src/lib/component/Icon/types.ts
steciuk: > I don't love the extra boilerplate it expects users to create

Just for the record, if by "extra boilerplate" you mean the ` <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" >` part - it is accepted, so you can just copy-paste (or [generate from](https://github.com/canonical/svelte-icons/blob/ce9a35d9d36f5f3bb697c3f4fb229895d607520a/scripts/build-icons.ts)) entire SVGs, but it is not required. You could provide only the geometry (starting from `<path d=...` in the linked example), and it should work just fine.

> Though that could cause IDE/convenience issues depending on when those components are generated.

E.g. `<framework>-ds-icons` that would generate the components, build them, and publish a versioned, ready-to-consume package (similar to [@canonical/svelte-icons](https://github.com/canonical/svelte-icons)) should be free of "IDE/convenience issues".

packages/ds-assets/docs/ICONS.md
bartaz-canonical: I believe the reason we used `icon.svg?v=hash` in our static sites (our `versioned_static` template function in Flask) was to keep the links "safe" - make sure they work even if the version doesn't match. So if the browser has cached the HTML with a link to old `icon.svg?v=hash` the icon would still load, even if current hash is different.

So indeed, the hash in file name will prioritise correctness (never load with a wrong hash).

I'm wondering if the correct version is that important for icons? Would we rather load older version of the icon in case of caching issue, or not load at all?

advl:
Please expect a delayed answer on this PR from us

packages/ds-assets/docs/ICONS.md
jmuzina: I went for embedding the hash directly in the filename after reading [this article](https://www.vergecloud.com/blog/what-is-cache-busting/) which suggests:

> a few older proxies and CDNs historically ignored query strings when deciding what to cache. Modern edge networks respect them, but if you need certainty across every intermediary, this is the weakest of the four options

> Content hashing, also called fingerprinting, renames the file using a hash of its contents, producing something like app.a1b2c3.js. The filename only changes when the bytes change, which means unchanged files keep their cached copies and updated files always get a fresh URL. This is the technique most production teams standardize on

That being said, it's a good point that it is probably better to display an outdated icon than nothing. So I think I will change this to use query parameters rather than changing the filename itself.

jmuzina:
> Please expect a delayed answer on this PR from us

Ack, I will place advancing this proposal on hold until I have an opinion from DS team. I am interested in:

- Thoughts on the problem we are seeking to address (is this the right place to address this problem?)
- Thoughts on the caching granularity (per-icon, or for the entire icon set)
- Thoughts on mechanism for including consumer-defined icons
- Thoughts on alternative proposal mentioned [here](https://github.com/canonical/pragma/pull/959#discussion_r3858694986) to use an approach more similar to Launchpad (bundling icons separately)

Until then we are blocked in Svelte projects outside Launchpad from using Pragma icons. Our alternatives right now are:
- Implement icons entirely locally within our project
- Try implementing an upstream Icon component that follows the React implementation's opinions. However we would not be able to use custom icons without additional manual work on our end
- Try optimistically implementing this approach in our tier
- Use the Launchpad icon system

PR #858, opened 2026-07-16
feat(svelte-wpe): Add `Announcement` component

Dampolo03:
## Done

Adds an `Announcement` component. This is a direct Svelte port of the existing [global React `Announcement`](https://github.com/canonical/pragma/tree/main/packages/react/ds-global/src/lib/component/Announcement). 

## QA

- Open [stories](https://6a1f1b4ab915d9a04a321108-rhctjqfryt.chromatic.com/)
- Follow QA from https://github.com/canonical/pragma/pull/746
- Compare the docs / behavior against the [global announcement](https://ds-react-global.canonical.com/?path=/docs/components-announcement--docs) and verify they are in line

### PR readiness check

- [x] PR should have one of the following labels:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [x] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [x] The code follows the appropriate [code standards](https://github.com/canonical/web-code-standards)
- [x] All packages define the required scripts in `package.json`:
  - [x] All packages: `check`, `check:fix`, and `test`.
  - [x] Packages with build steps: `build` to build the package for development or distribution, `build:all` to build **all** artifacts. See [CONTRIBUTING.md](../old/CONTRIBUTING.md#24-full-artifact-builds-buildall) for details.
- [x] If this PR introduces a **new package**: first-time publish has been done manually from inside the package directory using `npm publish --access public` (first-time publishing is not automated). Run `bun run publish:status` from the repo root to verify. Then configure an OIDC [trusted publisher](https://docs.npmjs.com/trusted-publishers) for the package on npmjs.com (repo `canonical/pragma`, workflow `tag.yml`) and set publishing access to disallow tokens, so the automated release workflow can publish future versions.
- [ ] If this PR does not require visual testing, add the `no visual change` label to skip Chromatic.

## Screenshots

_Storybook stories added (Default, Criticality, LongMessage, RightToLeft); render captured via Chromatic on this PR._

PR #961, opened 2026-08-26
feat(svelte-ds-global): Upstream Svelte `Announcement` from WPE tier to Global

jmuzina:
## Done

Upstreams Svelte `<Announcement>` (originally implemented in WPE tier by @Dampolo03 in #858 ) to the global tier.

## QA

- Open [stories](https://feat-upstream-announcement--6a3ca65a785e76b915a40851.chromatic.com/?path=/docs/work-in-progress-component-announcement--docs&globals=scheme:light)
- Follow QA from https://github.com/canonical/pragma/pull/746
- Compare the docs / behavior against the [global announcement](https://ds-react-global.canonical.com/?path=/docs/components-announcement--docs) and verify they are in line

### PR readiness check

## Screenshots

image


diana-stanciulescu:
@jmuzina looks good to me!

diana-stanciulescu (approved):
Looks good to me!

diana-stanciulescu:
@jmuzina at second look I noticed that we had an error in the design specs on typography - the heading of the announcement should use "typography-heading-6-$root" instead of the primary bold (they look the same but mean different things). 

There should be no gap/space between the annoucenment heading and body text. 

FYI there is a bug in the typography tokens regarding line height - that one should be fixed upstream - but for apps , the size would need to resolve to 14px size / 20px line height. 

The rest is per design, thanks!

packages/svelte/ds-global/src/lib/_work_in_progress/Announcement/Announcement.svelte
goulinkh, line 18: Can this component benefit from some of the accessibility attributes dependency on the criticalility.
- `criticality: error/success/warning`: `role="alert"/aria-live="assertive"`
- `ciriticality: information`: `role="status"/aria-live="assertive"`

Any opinions on this @steciuk since you are more knowledgeable in this domain?

packages/svelte/ds-global/src/lib/_work_in_progress/Announcement/types.ts
goulinkh, line 1: Are there plans to make the announcement dismissible?

goulinkh (dismissed):

packages/svelte/ds-global/src/lib/_work_in_progress/Announcement/Announcement.svelte
steciuk: [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role) says:

> The alert role should only be used for information that requires the user's immediate attention, for example:
>
> - An invalid value was entered into a form field
> - The user's login session is about to expire
> - The connection to the server was lost so local changes will not be saved

and

> Warning: Because of its intrusive nature, the alert role must be used sparingly and only in situations where the user's immediate attention is required.

So I don't think we should derive these from the modifier that drives the visuals, as for example a `criticality: error` announcement may be used to surface a persistent system failure message (as seen in the [Storybook example](https://6a3ca65a785e76b915a40851-ugzufouwqu.chromatic.com/?path=/story/work-in-progress-components-announcement--error)), in which case using the alert role would be counterproductive. Even the less-intrusive `status` role has specific use cases that don't necessarily align with visual styling. I think the consumer should be explicit and add these attributes themselves when appropriate.

packages/svelte/ds-global/src/lib/_work_in_progress/Announcement/types.ts
diana-stanciulescu: open question atm - maybe it should be an option > we would need to update the design. (general comment: design for announcement is experimental/partial)

packages/svelte/ds-global/src/lib/_work_in_progress/Announcement/Announcement.svelte
diana-stanciulescu: sounds like "alert" is more appropriate for notifications than announcement. If announcement are meant to communicate something to the user that is independent of user actions or system status , it is just there as text, and the visual treatment is an emphasis. (names we considered for the component were: callout or banner). 
I am also questioning wether the critical;ity modifiers are appropriate for announcement in this case if it should be simply an informational or warning message (but no success or error)

jmuzina: Sounds to me like we should not infer roles / aria-live for `Announcement` and consider this as a suggestion for the `Notification` component.

@diana-stanciulescu , if you would like to change the modifiers accepted by Announcement as you mentioned, let me know. The change would need to be applied to the React version as well.

diana-stanciulescu: not yet - this needs more consideration so ok to keep for now - but was mentioning for context

jmuzina: ack, thanks - @goulinkh @steciuk OK to resolve with un-inferred a11y attributes and a future consideration of adding this guidance on the difference between visual and semantic roles to `Notification`?

advl: Resolving per the consensus above: `Announcement` ships **without inferred `role`/`aria-live`** — per MDN, `alert`/`status` semantics fit notifications reacting to user actions or system state, while an announcement is standing content whose visual treatment is emphasis, so hard-wiring live-region semantics here would misfire for its actual use. The criticality → `role`/`aria-live` mapping is captured as design guidance to consider for `Notification` instead, and the component remains in `_work_in_progress` so this can be revisited if the design direction changes.

---

packages/svelte/ds-global/src/lib/_work_in_progress/Announcement/types.ts
advl: Resolving as an open design question rather than a blocker: dismissibility stays out of scope for this upstream — the Announcement design is still experimental/partial, and adding a dismiss option would need a design update first. The component lives under `_work_in_progress`, so the API can grow a dismiss affordance once the design settles, without a breaking change to a stable surface.

---

jmuzina:
Made the [requested spacing/typography changes](https://github.com/canonical/pragma/pull/961#issuecomment-5421532811), but then noticed this throws off the alignment of the announcement's icon with the header text. Fixing
image

goulinkh (approved):

data/svelte-ds-global/implementationObjects.ttl
jademathre-canonical, line 1: When I generated this, I got a few extra lines:
```
    ds:importStatement "import { Announcement } from \"@canonical/svelte-ds-global\";";
    ds:importStatement "import { Breadcrumbs } from \"@canonical/svelte-ds-global\";";
    ds:importStatement "import { SkipLink } from \"@canonical/svelte-ds-global\";";
```
Image

packages/svelte/ds-global/src/lib/_work_in_progress/Announcement/index.ts
jademathre-canonical, line 1: `export * from "./Announcement/index.js";` can be removed from `packages/svelte/ds-app-wpe/src/lib/components/index.ts`

packages/svelte/ds-global/src/lib/_work_in_progress/Announcement/Announcement.stories.svelte
```diff
@@ -6,7 +6,7 @@
   import Announcement from "./Announcement.svelte";
 
   const { Story } = defineMeta({
-    title: "Components/Announcement",
+    title: "_work_in_progress/Components/Announcement",
```
jademathre-canonical: technically should be `title: "_work_in_progress/component/Announcement",`

packages/svelte/ds-global/src/lib/_work_in_progress/Announcement/Announcement.stories.svelte
jademathre-canonical: ` lang="ar"` :P

jademathre-canonical (approved):
LGTM!

data/svelte-ds-global/implementationObjects.ttl
jmuzina: I think it's been resolved, try running `bun i` before `bun run collect`

advl: this is expected

advl (approved):

PR #980, opened 2026-08-26
feat(pragma-cli)!: create is summon — one generator surface, two CLIs

advl:
## Done

**`pragma create <args…>` ≡ `summon <args…>`** — same grammar, same flags, same wizard, same defaults, same failure modes, byte-identical generated trees. One machine in two binaries.

The hand-maintained mirror tables, the flag-alias map, the rename adapter and the PRA-14 availability gate are deleted in favour of one projection: summon's generators are the single source, and pragma's `create` surface is generated from them.

- **The framework becomes a tree segment.** `pragma create component react src/components/Button` is `summon component react src/components/Button` wearing a different name. The old `--framework` form dies loudly.
- **All three `create` nouns run from a published install.**
- **Non-TTY bare mutation refuses in BOTH CLIs**, identically.
- **MCP `create_*` schemas are prompt-derived**, so an agent and a human see the same surface.
- The wizard is shared through `summon-core` (ruling R5) rather than a resurrected `cli-core`.

## The contract is executed, not written down

There was a generated `parity-contract.md` with a byte-drift guard. It is deleted, and this is the most consequential change in the rebase.

Round 22 found why it had to go: `emitParityContract()` **takes no arguments**, so its output is invariant under every change to bindings, generators, template roots, prompt sets and package versions — and the drift test, which compared committed bytes to that invariant output, was its only reader. The header nonetheless promised *"behavior without doc and doc without behavior are both build failures"*. Five consecutive review rounds corrected its prose without touching the reason the prose kept going stale.

`crossCli.subprocess.test.ts` keeps the guarantee, and always did the real work: it spawns **both** CLIs over argv derived from the same fixture answers and compares what they emit, so an argv hand-drift cannot fake parity. Every reference is repointed at it — including the one pragma's generated reference prints to users.

## Rebased onto the de-compile (#972), and cut to match

#972 removed the compiled binary. A large part of this branch existed to make generators work *inside* that binary, and all of it is now deleted:

- **summon-core's entire `template/embedded/` subsystem** — the manifest store, builder, key scheme, and their tests. `loadTemplateSync` becomes what it always was underneath, a file read, and loses the `prefix` parameter that existed only to compute a manifest key.
- **The `template()` branch that FAILED a disk read "in embedded context"** — a guard against bypassing a manifest that no longer exists.
- **The embedded version fallback and `/$bunfs` guards** in `summon-package` and `summon-application`. Their own docstrings said the disk walk *"stays primary: on disk it can never be stale"* — so with no binary anchor the fallback was unreachable.
- **pragma's generated template manifest**, the codegen that produced it, the roots that declared it, and the injection that installed it before every `create`.
- **`summon-component`'s `shared/loadTemplate.ts`**, a pass-through once the prefix went, plus the `./embedded` package export that existed for pragma's injection.
- **`bunfsPrefix.subprocess.test.ts`**, which compiled a real binary to pin a deleted constant.

`CREATE_GENERATORS` now carries `paths` and nothing else.

**Net: 14,314 → 11,974 insertions.**

### One thing deliberately NOT collapsed

`mount.ts` (548 lines) looked like a candidate to fold into `summon-core`'s `registerGeneratorCommand.ts`. It is not. The registrar owns *everything derivable from the prompts* — command spec, options, grouped help, namespace parents — and injects host-shaped concerns through `GeneratorCliHost`. `mount.ts` is pragma's implementation of that host: its VerbSpec synthesis, its dispatch, its SEC-2 jail, its completion model, its generated reference. Merging them would push pragma-specific concerns into the shared package — re-specialising it, the opposite of the direction this codebase is moving. `registerFromBarrel` in `cli/summon` is already "a thin host over this", so summon and pragma are two hosts over one registrar, which is exactly the intent.

## Round 23 — the five findings the last review left owed

- **F1** the confirm-gate exit code was read behind a frame-only poll; the code is set by a passive effect that flushes *after* the commit painting the error. Round 21 fixed exactly this in `App.invalidAnswer.test.tsx` and left the file it copied the shape from.
- **F2** three sites asserted the gate had painted after one fixed 25 ms tick, when it is the tail of an async chain. Failed 2 in 84 loaded runs, 0 in 30 unloaded — visible only under the parallelism CI runs in.
- **F5** the esc cell promised "previous values pre-filled" and asserted the prompt's *message*, which renders either way.
- **F3** the subcommand rule was wrong in both halves — it credited both summon spellings to one mechanism, and templated one rendering of a two-rendering class.
- **F4** the invariant-emitter problem above.

## Performance

Both fast-path ceilings are raised to 220 ms, and the cause is named rather than absorbed:

| Path | measured | old | new |
|---|---|---|---|
| `pragma --help` | ~151 ms | 130 | **220** |
| `pragma __complete` | ~163 ms | 150 | **220** |

Two costs stacked. #972 roughly doubled the fast paths by shipping JavaScript that node executes. This branch adds ~46 ms of eager import: `capabilities/index` barrels every capability, and `create.verb.ts` statically value-imports `@canonical/summon-core/projection` — the logic shared with summon so the two cannot drift. **Both fast paths pay for it and neither uses it.** `lazy.test.ts` stays green because that subpath is not what it guards; summon-core proper, React, zod and oxigraph are all still absent from the fast-path graph.

Both are marked **PROVISIONAL** in `budgets.ts` and `BUDGETS.md`. Moving that import behind the lazy boundary is the real fix and belongs with follow-up work, not a budget edit.

## QA

```bash
pragma create component react src/components/Button --yes
pragma create package --yes
pragma create application react my-app --yes
summon component react src/components/Button --yes    # byte-identical tree
```

Gates: check 3/3 · pragma **1224** · summon-core 551 · component 90 · package 49 · application 50 · cli/summon 62 · perf **19/19** · `bun.lock` stable under `bun install --ignore-scripts`.

### PR readiness check

- [x] Label: `Breaking Change 💣`
- [x] Conventional Commits title
- [x] Follows the [code standards](https://github.com/canonical/web-code-standards)
- [x] Package scripts unchanged (`gen:parity` removed with its subsystem)
- [x] N/A — not a new package
- [x] `no visual change`

---


advl:
Superseded by a three-PR stack. This was ~14k lines spanning three separable concerns; splitting them makes each reviewable on its own.

- **#997** — `@canonical/task`: the `verbatim` write marker (+51/−4). Independently mergeable.
- **#998** — summon: the projection extracted into summon-core as a data seam plus a Commander adapter.
- **#999** — pragma: the `create` mount that deletes the hand-maintained surface mirror.

#998 and #999 merge together — pragma's mirror asserts against summon's surface, so neither is green alone until the mirror is gone.

The split is lossless: the union of the three branches is byte-identical to the tree verified here.

Since this PR was opened, the seam itself was redesigned — structure is now shared and presentation is host-owned, which removed the prose-then-reverse error protocol and confined all process writes to one function. #998 carries that.

---

PR #1006, opened 2026-08-27
chore(labels): improvements to the repo labels

advl:
## Done

Written by Opus 5 under @advl 's instructions

**This is a proposal, opened for discussion — let's not merge it before we've agreed on the taxonomy.** The code is complete and working, so it's easier to argue about something concrete than about a doc.

### The problem

PR titles in this repo are already strict: `pr-lint.yml` validates every one against Conventional Commits, and **300/300 of the last merged PRs comply**. Labels are the opposite — a checkbox in the PR template that nobody is reminded about:

- **50 of the last 150 merged PRs (33%) carry no type label at all.**
- **78 of 300 have a title prefix that disagrees with their labels** — `feat(…)` merged with no `Feature 🎁`, `refactor(cli)!:` labelled only `Breaking Change 💣`.
- `refactor` has no label at all. Of 32 merged refactors: 14 got `Maintenance 🔨`, 10 got nothing, 7 got only `Breaking Change 💣`.
- `Maintenance 🔨` is a bucket for four different types — `refactor`, `chore`, `test`, `ci` — which together are **23% of all merged work**.
- `Breaking Change 💣` is used *instead of* a type rather than alongside it, so a breaking PR loses its type entirely.
- The lint passes no `types:` input, so it accepts commitlint's full default list. `build`, `perf`, `style` and `revert` are legal titles today with no label to match.
- 13 labels have **zero uses** across ~800 PRs and issues (every `Review: … -1`, `Review: Code +1 (with changes)`, `Don't merge`, `Dependencies`).

The pattern behind all of it: the only label that gets applied reliably is `no visual change` (145 uses), because it's the only one that *does* something — it gates Chromatic. Labels people can't feel don't get applied.

### The proposal

One rule decides everything:

> **A bare lowercase label is derived from the PR title. A `Namespace: value` label is a human judgement.**

The type is already in the title, so CI derives it instead of asking a human to restate it. `feat(router)!: …` gets `feat` + `breaking`; edit the title and the labels follow. The full set, the colour system and the rationale are in **[`docs/references/LABELS.md`](docs/references/LABELS.md)**.

| | |
| --- | --- |
| **Derived** (slate, machine-owned) | `feat` `fix` `docs` `refactor` `chore` `test` `ci` `revert` `breaking` |
| **CI directive** | `Chromatic: skip` (was `no visual change`) |
| **Review** | `Review: {code,QA,design,a11y,UX,Chromatic} {needed,+1}` |
| **Status** | `Status: blocked` `Status: do not merge` `Status: question` |
| **Priority** | `Priority: {low,medium,high}` |
| **Topic** | `Topic: design tokens` — for migration trains that span many PRs |
| **Exception** | `good first issue`, kept verbatim because GitHub treats that exact name specially |

36 labels → 30, and every survivor has either usage evidence or a functional role.

### What's in the diff

- **`.github/workflows/pr-lint.yml`** — pins `types:` to the closed vocabulary, then derives the label from the title in a second job. Adds *and removes* the labels it manages, so an edited title never leaves a stale type behind. Creates a label that doesn't exist yet, so a freshly generated repo labels its first PR correctly. Skips fork PRs, whose token is read-only.
- **`chromatic._template.yml`** — the gate accepts `Chromatic: skip` *and* `no visual change`, so the rename needs no atomic cutover in either merge order. The 16 chromatic workflows inherit this from the one reusable template.
- **`scripts/one-time-only/migrate-labels.sh`** — the whole migration, idempotent. Uses `gh label edit --name`, which **renames in place**: every PR and issue keeps its label and history stays intact. Backfills the type label on all open PRs, and `Topic: design tokens` on the 21 PRs of that train.
- **`docs/references/LABELS.md`** — the taxonomy, the colour system, and a table of what was removed with the evidence for each.
- **PR template, `AGENTS.md`, Renovate config** — the type label is no longer a human step, so it stops being asked for. Renovate drops its hard-coded `Maintenance 🔨`; `chore(deps): …` derives `chore` on its own.
- **`packages/summon/monorepo/`** — the same workflow and PR template ship to every generated monorepo, so new repos start consistent instead of inheriting the drift. (`summon/package` has no `pr-lint`, so it's untouched.)
- **`.jira_sync_config.yaml`** — `enhancement` was a duplicate of `Feature 🎁`; the mapping moves to `feat: Story`.

### Worth arguing about

1. **One colour for all derived types, or a colour each?** They're slate here on the argument that a derived label is an index for filtering, not a signal that needs attention — colour is reserved for human judgements. This is the change most likely to feel like a downgrade, and it's one line in the script.
2. **Deleting the six `Review: … -1` labels.** Zero uses ever; GitHub's "Request changes" is the native mechanism. But it is a process change, so it's the team's call, not mine.
3. **`revert` is in the allowed types** with zero historical uses, because being blocked by the title lint mid-incident is a bad trade. `perf`, `build` and `style` are deliberately left out — `chore` covers them.
4. **Renaming `no visual change`** is the most-used label (145). The gate accepts both names so nothing breaks, but muscle memory is a real cost — happy to drop this one item and keep the name.

## QA

- `pr-lint.yml` and `chromatic._template.yml` parse as valid YAML; both Renovate configs still parse as JSON.
- `packages/summon/monorepo` tests pass (22/22) — the generator asserts the template file list, which is unchanged.
- **The label derivation itself is unverified until this PR runs it.** The first real check is this PR's own CI: the title is `chore(labels): …`, so `apply-type-label` should put `chore` on it and nothing else.

### Merge checklist

- [ ] Team agrees on the taxonomy (points 1–4 above)
- [ ] Merge
- [ ] Run `scripts/one-time-only/migrate-labels.sh` once — until it runs, the old label names simply stay in place and CI creates the new ones as PRs land
- [ ] Follow-up PR drops the `no visual change` clause from the Chromatic gate once the rename is done

---
🤖 Drafted with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/.jira_sync_config.yaml
+++ b/.github/.jira_sync_config.yaml
@@ -33,4 +33,4 @@ settings:
   # (Optional) Dictionary mapping GitHub issue labels to Jira issue types.
   # If label on the issue is not in specified list, this issue will be created as a Bug
   label_mapping:
-    enhancement: Story
+    feat: Story
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,15 +10,15 @@ Fixes [list issues/bugs if needed]
 
 ### PR readiness check
 
-- [ ] PR should have one of the following labels:
-  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
-- [ ] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
+- [ ] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format, using one of the allowed types: `feat`, `fix`, `docs`, `refactor`, `chore`, `test`, `ci`, `revert`.
+  - The matching type label is applied automatically from the title — there is no type label to add by hand.
+  - Breaking changes are marked with `!` in the title (e.g. `feat(router)!: …`), which adds the `breaking` label.
 - [ ] The code follows the appropriate [code standards](https://github.com/canonical/web-code-standards)
 - [ ] All packages define the required scripts in `package.json`:
   - [ ] All packages: `check`, `check:fix`, and `test`.
   - [ ] Packages with build steps: `build` to build the package for development or distribution, `build:all` to build **all** artifacts. See [CONTRIBUTING.md](../old/CONTRIBUTING.md#24-full-artifact-builds-buildall) for details.
 - [ ] If this PR introduces a **new package**: first-time publish has been done manually from inside the package directory using `npm publish --access public` (first-time publishing is not automated). Run `bun run publish:status` from the repo root to verify. Then configure an OIDC [trusted publisher](https://docs.npmjs.com/trusted-publishers) for the package on npmjs.com (repo `canonical/pragma`, workflow `tag.yml`) and set publishing access to disallow tokens, so the automated release workflow can publish future versions.
-- [ ] If this PR does not require visual testing, add the `no visual change` label to skip Chromatic.
+- [ ] If this PR does not require visual testing, add the `Chromatic: skip` label to skip Chromatic.
 
 ## Screenshots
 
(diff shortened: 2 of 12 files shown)


stavropouloskonstantinoss (approved):
I think the proposed change is in the right direction since the type is already in the title and CI enforces it, so deriving the label removes a step that could be forgotten.

1. One idea is to have one colour for all derived types, so they're uniform and there's no information duplication between the colour and the label. But I am not sure though which will be better in practice.

2. Agree — better to use GitHub's Request changes mechanism for a PR.

3. Agree.

4. Muscle memory is real memory that it takes effort to change, so I think the gate should accept both names.

PR #1009, opened 2026-08-27
feat(collect): export the implementation graph as turtle in the release

advl:
## Summary

The release now produces a Turtle (RDF) dataset of the implementation libraries and their implementations, for ingestion into the design-system knowledge graph.

### Collector (`scripts/collect-implementations.ts`)
- Imports the collect engine from the published `@canonical/design-system@0.2.2` instead of a cross-repo relative path (which could not run in CI).
- New `--version` flag (defaults to the `lerna.json` version): stamps `ds:version` on each library and emits `ds:versionedLink` blob URLs pinned to the release tag; `ds:headLink` is now a full `blob/main` URL. New `--out` flag overrides the output directory.
- Implementations are named IRIs (`ds:implementation.{librarySlug}.{blockLocalName}`) instead of blank nodes, so they keep stable identity across releases in the knowledge graph (engine change in design-system 0.2.2).
- The aggregate index emits absolute `rdfs:seeAlso` IRIs — the previous relative ones fail to parse without a base IRI and would have broken ingestion.
- Regenerated `data/` (fixes the stale `react-ds-global` paths); output is deterministic and sorted.

### Coverage
`packages/svelte/ds-global` joins collection (new `design-system.json`). Deferred as follow-ups until their annotations match the spec'd block slugs: `react/ds-app` (dashes vs the spec's underscores, e.g. `side_navigation`), `lit/ds-prototype` (unspec'd prototype blocks), `svelte/ds-app-wpe` (`ds:site` vs `ds:sites` tier drift).

### Distribution
- The canonical data stays committed at root `data/` — a git pack ref scans the repo root's `definitions/**` + `data/**` (no monorepo-subdir support), so this is what `pragma sources update` consumes via the new `@canonical/ds-implementations` default pack in `pragma.conf.ts`.
- New `packages/ds-implementations` package snapshots `data/` at `prepack` for the npm channel (published by the existing `lerna publish`).

### Release wiring
- `lerna-version` action regenerates `data/` with the new version before `git-commit.sh`, so the tagged commit carries the stamped TTL (published npm content ≡ tagged source).
- New `.github/actions/github-release` composite action (`action.yml` + `release.sh`, idempotent, `--verify-tag`): the publish job now creates a GitHub Release for the tag and attaches a concatenated `pragma-implementations.v{VERSION}.ttl`; the job gains `contents: write`.
- New PR-CI freshness job runs `bun run collect:check` — fails when committed `data/` drifts from the in-tree `@implements` annotations.

## Test plan

- [x] design-system engine: 281/281 tests (12 new for named IRIs, version stamping, links, determinism)
- [x] Deterministic double-run of `bun run collect` (identical hashes)
- [x] All 7 generated TTL files and the concatenated release artifact parse in oxigraph (217 quads)
- [x] End-to-end: scratch project with a `file:` pack ref at the repo root → `pragma sources update` builds the store; SPARQL returns 37 `ds:ImplementationObject`s with correct `ds:versionedLink`s
- [x] CLI suite: 1144 passed (pack-roster pins updated to four packs); `tsc` and biome clean on changed files
- [x] `bun run collect:check` passes on the committed data
- [ ] End-to-end release proof: next experimental `tag.yml` dispatch

---


advl:
Superseded by #1010 — same change, reopened from a branch following the AGENTS.md naming convention (`feat/implementation-graph-turtle-export`) and with the PR template applied.

---

PR #1021, opened 2026-08-27
feat(pragma-cli)!: the entity read returns a typed, related, bounded neighbourhood, as Turtle

advl:
## Done

First of the stack taking the MCP resource surface to S-grade. This one is the **read projection** — what an agent actually receives when it resolves `pragma:{+uri}` or calls `graph_inspect`.

Everything lands in the shared `readEntity`, so `graph inspect`, the `graph_inspect` tool and the resource read stay in lockstep and the mirror contract holds by construction.

### 1. The read became the entity's *neighbourhood*

- **Typed terms.** Reads ke's `termBindings` (the term-preserving view of the same rows) instead of the stringly `bindings`. This is what makes prefix compaction *safe*: an IRI object and a string literal that merely looks like one were previously indistinguishable, and `datatype`/`language` were silently dropped. Compaction now runs against the store's **own merged prefix map**, not the compiled-in display map.
- **Inbound relations.** Everything asserting a predicate *about* the subject. This is where the relations actually live: `ds:global.component.button` could not answer *which package implements it* (`ds:implementsBlock`) or *what inherits from it* (`ds:inheritsFrom`), because a subject-only read never sees the other side of the arrow.
- **Blank nodes inlined.** A blank node's label is store-local and re-mints on every load, so `_:c8436e76…` was a dead end no `pragma:{+uri}` read could resolve. They are now inlined one level as records — a changelog entry's type, text and timestamp.
- **Neighbour titles** — but only when the label says something the URI's local name did not. Measured: all 21 titles in a button's neighbourhood merely restated its own URI in different casing, so emitting them spent context to say the same word twice.

### 2. Bounded by what an inbound group *is*

Not by a fixed page size. A **relation** fans in narrowly and every subject is part of the answer, so it is listed. A **roster** grows with the data rather than the model — a class extension, a tier's membership — and is already answered by `pragma/instanceCount` and the noun's list verb, so it is **sampled**: a few exemplars flagged `sampled`, never a page.

The two are told apart by **fan-in, not predicate name**, so the kernel names no vocabulary. Keying on `rdf:type` would have fixed classes and left a `detailed` read of one tier at 20.9 KB — `ds:tier` fans in 141 deep here and is the same pathology under a different predicate.

`count` is always the true total either way. A sample silently passing for the whole set is how "1 implementation" becomes a wrong answer.

### 3. Turtle, and a budget for prose

The JSON projection cost ~95 bytes of wrapper to convey a 13-character IRI, paid per term — ~6.1 KB of one button was pure structure. The resource read now serves **`text/turtle`**, which is also simply what the body *is*; it was claiming `application/json` for a graph.

It is not merely shorter — it is the shape the data already has. IRI versus literal is *syntax* (`ds:Foo` against `"Foo"`), which is the thing the typed terms needed a `termType` field to say. A blank node is `[ … ]` inline. Datatypes and language tags are native. And an inbound edge is just a triple written the other way round — the whole `InboundGroup` structure exists only because JSON cannot point an arrow backwards.

The larger cost turned out not to be structure at all: that button spent **8,572 of its 9,734 literal characters on two fields** (`ds:guidelines` 5,814, `ds:usage` 2,758). Long-form documentation is what `detailed` is *for*, and the noun's own lookup verb already serves those fields under its own disclosure. Below `detailed` a literal is previewed and marked with its true length.

Counts, samples and truncations ride as `#` comments — the one place a serialization can say something *about* the data without asserting it as data.

## Measured

| entity | before (JSON, `main`) | after (Turtle, `standard`) |
|---|---:|---:|
| `ds:global.component.button` | 12,015 B | **4,518 B** (−62%) |
| `ds:Property` (330 instances) | — | **747 B** |
| `ds:global` (141 members) | — | **608 B** |

The button shrinks 62% *while gaining* its inbound relations, seven readable nested records, and titles. The hub figures have no "before" because a subject-only read carried no inbound edges at all; within this PR, listing rosters cost them 19.5 KB and 20.9 KB at `detailed` before the fan-in rule.

Three queries per read, 3–11 ms warm even on the 335-deep hub — well inside the 300 ms budget.

```turtle
@prefix ds: <https://ds.canonical.com/> .

ds:global.component.button
  a ds:Component ;
  ds:name "Button" ;
  ds:tier ds:global ;
  # ds:guidelines — 5,814 chars, showing 400
  ds:guidelines """# Button hierarchy

Each button variant signals a different level of emphasis…""" ;
  ds:changeLog
    [ a ds:ChangeLogEntry ; ds:change "Decided to split the button into multiple components…" ; ds:changeType "decision" ; ds:timestamp "2026-02-20T00:00:00.000+01:00" ] .

# ── referenced by ──
# ds:implementsBlock — 1
ds:implementation.react-ds-global.global.component.button ds:implementsBlock ds:global.component.button .
# ds:inheritsFrom — 1
ds:global.component.dropdown_button ds:inheritsFrom ds:global.component.button .
```

## Bugs fixed along the way

- **Nondeterministic output.** Nested record order followed the blank-node label, which oxigraph re-mints on every load — so two reads of the same unchanged entity disagreed. Now ordered by content. Caught by the mirror test.
- **Blank nodes exhibited as exemplars.** `ds:Property`'s 330 instances are all blank nodes, and the sampler dutifully rendered three `[] rdf:type ds:Property .` lines — identical, useless, and *wrong*, since every `[]` in Turtle mints a fresh node. Blank nodes are now excluded from exemplars while still counting; a group with no nameable member says `none individually addressable`.
- **Two serializer bugs that would have shipped broken documents:** a trailing note swallowed the `;` separator (Turtle comments run to end-of-line), and a literal ending in `"` collided with the `"""` delimiter — reachable on ordinary prose once truncation is in play.

## Deviation from plan

`nested` values are `ReadTerm`, not the `PackChildRow` strings originally specified. Flattening them would reintroduce the IRI-versus-literal ambiguity the term projection exists to close, and leave the serializer unable to tell `ds:Foo` from `"ds:Foo"`.

## QA

```bash
# from the repo root
bun run check && bun run test
```

```bash
cd packages/cli/pragma

# the neighbourhood, as Turtle
bun run src/bin.ts graph inspect ds:global.component.button --format llm

# inbound relations that were previously invisible
bun run src/bin.ts graph inspect ds:global.component.button --format json \
  | jq '.inbound[] | {predicate: .predicate.prefixed, count, sampled}'

# a roster is sampled and says so; a narrow relation is listed whole
bun run src/bin.ts graph inspect ds:Component --format llm | tail -8

# a literal that looks like an IRI is NOT compacted
bun run src/bin.ts graph inspect ds:global.component.button --format json \
  | jq '.groups[] | select(.predicate.prefixed=="ds:anatomyDsl") | .objects[0].termType'
# → "Literal"

# raise the ceiling
bun run src/bin.ts config set detail detailed
```

The Turtle is verified by **parsing it with the same Oxigraph engine the store uses** (`resources.test.ts`), not by matching shapes — a document that only looks right is how a reader gets a syntax error instead of an entity.

- CLI package suite: **1168 passed**, 13 skipped, 5 todo.
- Root `bun run check`: passes, 62 projects.
- Root `bun run test`: the `@canonical/pragma-cli` `create` failures are the documented 5 s-timeout flake under parallel load — all 43 pass solo. `svelte-ds-*` / `react-ds-global` fail on missing Playwright browser binaries, environment-only.

**Not in scope, deliberately:** `graph query` still renders untyped string bindings. Blank-node inlining is depth 1 — measured 11 nested blank nodes out of 922 graph-wide, so one level is what there is to inline.

### PR readiness check

  - [x] Packages with build steps: `build`, `build:all`.
- [x] If this PR introduces a **new package**: n/a.

## Screenshots

n/a — CLI/MCP payload change, no visual surface. Sample output shown above.

---

## BREAKING CHANGES

1. The `pragma:{+uri}` resource read returns **`text/turtle`**, not `application/json`.
2. `graph inspect --format llm` returns **Turtle**, not Markdown.
3. `graph_inspect` and `graph inspect --format json` change shape: `groups[].predicate` and `groups[].objects[]` are term objects (`{termType, value, prefixed?, title?, datatype?, language?}`) rather than strings; the result gains `prefixed`, `inbound`, `nested` and `detail`; `nested` values are terms; literals are previewed below `detailed`.

---


advl:
Good catch — class→instance was the dominant cost. Pushed 67b673c.

The fix generalises past `rdf:type`, because keying on it would have fixed classes while leaving the same pathology under another predicate. The read now tells two kinds of inbound edge apart **by fan-in**, not by name:

- A **relation** fans in narrowly and every subject is part of the answer (`ds:implementsBlock`, `ds:inheritsFrom`) → listed.
- A **roster** fans in without bound because it grows with the data rather than the model, and is already answered by `pragma/instanceCount` and the noun's list verb → **sampled**: a few exemplars flagged `sampled`, never a page.

| entity | `detailed` before | after |
|---|---|---|
| `ds:Property` (330 instances) | 19.5 KB | **2.4 KB** |
| `ds:global` (141 members via `ds:tier`) | 20.9 KB | **2.1 KB** |
| `ds:global.component.button` (2 relations) | 15.8 KB | unchanged — still listed in full |

`ds:global` is the case that argued against a `rdf:type` special-case: it is a tier, not a class, and it was the *worst* single read in the graph. Deriving the rule from shape rather than vocabulary also keeps the kernel free of RDF terms, so a fork whose rosters hang off different predicates gets the same bounding for free.

`count` stays the true total either way. `sampled` is deliberately distinct from `truncated`: "sample of 5" tells a reader that paging will not produce the rest and the list verb is where the full set lives, where "showing 5" invites them to ask for more.

Two new cases pin both branches — a 22-deep roster is sampled with an exact count, and a fan-in-of-one relation is listed whole with neither flag set.

🤖 Posted with [Claude Code](https://claude.com/claude-code)

advl:
All eleven addressed in `06705a0b`. Two were correctness bugs in the promises this PR itself makes — thank you, those were the valuable ones.

**Inbound counts were not the true total.** Exactly as described: one `LIMIT 500` answered counting and exhibiting at once and answered neither. The first predicate to fill it under-reported its own `count`, *and* every predicate ordered after it vanished from the read — while `InboundGroup.count` documented the exact total. Counts now come from a `GROUP BY` aggregate the store computes; only exemplars are bounded, in a second pass that filters blank nodes **in the query**, so the budget is spent on subjects that can actually be shown. Pinned by a fixture where a 22-deep predicate sits beside a 1-deep one and both must survive.

**A blank node linked from two predicates lost one edge.** Confirmed, and your note about Turtle made it worse than it looked — the ordinary blank object is filtered out of the serialization too, so nothing remained to notice the loss by. Keyed by predicate *and* node now, with a doubly-linked fixture node.

- **RDF 1.2 `direction`** — correct, and it is the exact flattening reading the term view was meant to stop. Carried on `ReadTerm`, emitted as `@lang--dir`, fixture literal added.
- **`String(value)` → `[object Object]`** — correct, and the whole plain-format record was unreadable. Rendered through the term formatter.
- **`prefixesOf` was incomplete** — correct, and your suggested remedy is the one I took: `InspectResult` now **carries** the map it was compacted against, so the reconstruction is gone entirely rather than patched. It could never have learned a prefix used only by the subject, a nested record key, or a datatype.
- **Absolute datatype IRIs unwrapped** — correct; `"x"^^http://example.com/type` is invalid. Wrapped as `<iri>` in every position, nested predicate keys included.
- **"none individually addressable" at `summary`** — correct and mine to own: at that level *every* group is empty because the level exhibits none, so it libelled perfectly nameable subjects. A group now records what it *would* have shown, so the two cases are distinguishable.
- **`graph_inspect` had no `disclosure`** — correct, the documented `detail` argument was unreachable. Rather than drop the claim I declared the ladder, so the knob is now real; reference regenerated.
- **Bounding table said 200** — that was the PR description lagging a mid-flight design change to the fan-in rule. The description has been rewritten.
- **`mcp-integration.md` described a JSON shape** — correct; rewritten for the Turtle document, with worked examples of the `#` comment bounds.
- **Dead `withoutBlankNodeLabels`** — removed.

CLI suite **1176 passed**; root `bun run check` passes across 62 projects. The single root-run failure is the documented `create` 5s-timeout flake — all 43 pass solo.

advl:
Closing — this work is **already on `main`**, merged inside #1025's squash commit (`e535e459`).

The two PRs were stacked: #1025 was branched off this one, so when it squash-merged it carried all five of these commits with it. That is also why this PR then showed as conflicting — not staleness, but redundancy.

Verified rather than assumed:

- `src/kernel/render/turtle.ts` is present on `main`, and `readEntity.ts` carries every marker from this branch (`ROSTER_THRESHOLD`, `LITERAL_CAP`, `exhibits`).
- Every file only this PR touched — `inspect.verb.ts`, `inspect.render.ts`, `inspect.test.ts`, `blockGraph.ts`, `turtle.ts`, `readEntity.ts`, `docs/mcp-integration.md`, `docs/reference/tools.md` — is byte-identical between this branch and `main`.
- Diffing this branch against `main` yields 162 insertions / 1030 deletions, and every affected file belongs to #1025. In other words the branch is `main` *minus* the listing feature: merging it would remove work, not add any.

What shipped, for the record:

| | before | after |
|---|---:|---:|
| `ds:global.component.button` | 12,015 B JSON | 4,518 B Turtle |
| `ds:Property` (330 instances) | 19,508 B | 747 B |
| `ds:global` (141 members) | 20,941 B | 608 B |

Plus inbound relations, blank nodes inlined as records, fan-in-derived roster sampling with exact counts, literal budgeting, and the eleven review fixes in `06705a0b`/`0d605fbf`.

⚠️ One thing to carry forward: the `Breaking Change 💣` label lived here, while #1025 merged as `Feature 🎁`. The breaking changes went out under it — the `pragma:{+uri}` resource read now returns `text/turtle` rather than `application/json`, `graph inspect --format llm` returns Turtle rather than Markdown, and the `graph_inspect` / `--format json` payload changed shape (terms instead of strings, plus `prefixed`/`inbound`/`nested`/`detail`/`prefixes`). Worth reflecting in the release notes even though the PR is closed.

🤖 Closed with [Claude Code](https://claude.com/claude-code)

PR #1024, opened 2026-08-27
fix(pragma-cli): name the typed URI in resolution errors, accept encoded ones

advl:
## Done

Three fixes to URI resolution in the CLI's pack kernel, all reachable from `graph inspect` and the `pragma:{+uri}` MCP resource read.

1. **Errors now echo what the caller typed, not the expansion.** `resolveUri` expanded a prefixed name and then validated the result, so the diagnostic quoted a string the user never wrote. The safety check still runs on the expanded IRI — only the *message* changed to name the input.

   ```
   $ pragma graph inspect 'ds:my component'

   # before
   ## Error: INVALID_INPUT
   Invalid uri "https://ds.canonical.com/my component".
   Recovery: URI contains characters not allowed in IRIs.

   # after
   ## Error: INVALID_INPUT
   Invalid uri "ds:my component".
   Recovery: URI contains characters not allowed in IRIs.
   ```

2. **Percent-encoded URIs resolve instead of failing.** A client that percent-encodes the value it puts in a `pragma:{+uri}` resource read got a hard failure. Resolution now retries once with the value decoded; if the decoded form resolves it is used, otherwise the original input is reported per (1). `decodeURIComponent` throws `URIError` on malformed input, so the decode attempt is guarded — a stray `%` stays a reportable input error rather than becoming a crash.

   ```
   $ pragma graph inspect 'ds%3AComponent'      # = MCP resource pragma:ds%3AComponent

   # before
   ## Error: INVALID_INPUT
   Invalid uri "ds%3AComponent".
   Recovery: Use a prefixed URI (e.g. "prefix:name") or a full URI.

   # after
   ## ds:Component — Component
   - **rdfs:label**: Component
   - **rdfs:subClassOf**: ds:UIBlock
   …
   ```

   Malformed encodings still report cleanly, with no `URIError` escaping:

   ```
   $ pragma graph inspect '%E0%A4%A'
   ## Error: INVALID_INPUT
   Invalid uri "%E0%A4%A".
   Recovery: Use a prefixed URI (e.g. "prefix:name") or a full URI.
   ```

3. **Documented examples that 404 against the shipped graph.** `ds:button` is not in the shipped pack index — the real entity is `ds:global.component.button`. A copy-pasteable example that fails is the first thing an agent or a new user tries. Retargeted on `graph inspect` (both examples plus the verb's `doc` line, which named `ds:button` as the addressing example) and, as a drive-by, on `graph query`'s ASK example, which silently answered `false`. `docs/reference/*.md` regenerated with `bun run gen:reference` rather than hand-edited.

   The `graph query` example had to move to the absolute-IRI form — **surprise:** the store's SPARQL parser rejects a prefixed local name containing more than one dot, so `ASK { ds:global.component.button a ds:Component }` is a parse error at the second `.` while `ASK { <https://ds.canonical.com/global.component.button> a ds:Component }` answers `true`. `graph inspect` is unaffected because it expands the prefix itself and embeds an `<iri>` token.

Deliberately **not** touched: `resources/provider.ts`'s `TEMPLATE_DESCRIPTION` still advertises `pragma:ds:button`, and its index reads are still unmemoised. Both are known and deferred to concurrent work on that file.

## QA

```bash
# fix 1 — the message names the input
pragma graph inspect 'ds:my component'

# fix 2 — a percent-encoded URI resolves; a malformed one reports cleanly
pragma graph inspect 'ds%3AComponent'
pragma graph inspect '%E0%A4%A'

# fix 3 — every documented example resolves
pragma graph inspect ds:global.component.button
pragma graph inspect https://ds.canonical.com/global.component.button
pragma graph query "ASK { <https://ds.canonical.com/global.component.button> a ds:Component }" --format json
```

Gate from the repo root: `bun run check` passes across all 62 projects. `bun run test` — `@canonical/pragma-cli` passes 1358 vitest + 22 perf tests (no assertion changes were needed; the new message shape broke nothing). Nx classified `pragma-cli:test` and `react-ds-app-launchpad:test` as flaky under parallel load — both pass solo, failing only with 5000 ms timeouts under contention. `svelte-ds-*` fail on missing Playwright browser binaries in a fresh worktree (environment-only, discountable per AGENTS.md). The known scaffolded-fixture leak into `packages/cli/pragma/src/lib/` reproduced and was cleaned; it is not committed.

### PR readiness check

## Screenshots

n/a — CLI/MCP diagnostics only, no visual surface. The before/after strings are quoted verbatim above.

### DIFF
--- a/packages/cli/pragma/docs/reference/commands.md
+++ b/packages/cli/pragma/docs/reference/commands.md
@@ -351,7 +351,7 @@ pragma doctor --format json  # machine-readable checks
 
 Show every triple where a URI is the subject, grouped by predicate.
 
-Inspect one entity: all predicate/object pairs asserted on the subject. Address it by prefixed name (ds:button) or absolute IRI.
+Inspect one entity: all predicate/object pairs asserted on the subject. Address it by prefixed name (ds:global.component.button) or absolute IRI.
 
 ```
 pragma graph inspect <uri>
@@ -369,8 +369,8 @@ pragma graph inspect <uri>
 **Examples**
 
 ```bash
-pragma graph inspect ds:button
-pragma graph inspect https://ds.canonical.com/button
+pragma graph inspect ds:global.component.button
+pragma graph inspect https://ds.canonical.com/global.component.button
 ```
 
 ### pragma graph query
@@ -396,7 +396,7 @@ pragma graph query <sparql>
 
 ```bash
 pragma graph query "SELECT ?s WHERE { ?s a ds:Component }"  # list every component subject
-pragma graph query "ASK { ds:button a ds:Component }" --format json
+pragma graph query "ASK { <https://ds.canonical.com/global.component.button> a ds:Component }" --format json
 ```
 
 ## info
--- a/packages/cli/pragma/docs/reference/tools.md
+++ b/packages/cli/pragma/docs/reference/tools.md
@@ -178,7 +178,7 @@ _No input parameters._
 
 ### graph_inspect
 
-Inspect one entity: all predicate/object pairs asserted on the subject. Address it by prefixed name (ds:button) or absolute IRI.
+Inspect one entity: all predicate/object pairs asserted on the subject. Address it by prefixed name (ds:global.component.button) or absolute IRI.
 
 Read-only.
 
--- a/packages/cli/pragma/src/capabilities/graph/examples.exec.test.ts
+++ b/packages/cli/pragma/src/capabilities/graph/examples.exec.test.ts
@@ -1,0 +1,107 @@
+/**
+ * The documented examples, EXECUTED against the shipped pack (PROTECTED).
+ *
+ * The existing documentation suite validates example GRAMMAR — that a `cmd`
+ * parses as a command this CLI declares. Grammar is not truth: the previous
+ * `graph inspect ds:button` examples passed that check for as long as they
+ * existed while 404ing for every user who copied them, and the `graph query`
+ * ASK example passed while quietly answering `false`.
+ *
+ * So these run the examples as WRITTEN, against the embedded graph, and assert
+ * they answer. An example is a promise to a stranger; this is the only check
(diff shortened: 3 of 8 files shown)


advl:
All three addressed in `f163ca7`. The point stands, and the gap is exactly how these bugs got in.

**The examples are now executed, not just parsed.** `examples.exec.test.ts` runs every documented example against the shipped pack and asserts it *answers*: the two `graph inspect` forms resolve to the documented entity, the ASK returns `true`, the SELECT returns rows. Your framing was the useful part — the documentation suite validates example *grammar*, and grammar is not truth. `ds:button` passed that check for as long as it existed while 404ing for everyone who copied it, and the ASK example passed while quietly answering `false`.

Arguments and parameter names are read from the verb's own declaration rather than retyped, so editing an example cannot leave the test asserting the old one.

**`resolveUri` gets a unit suite** (`iri.test.ts`). Every read reaches it, so its diagnostic is the sentence a user or agent has to act on — and nothing asserted that sentence, which is how it came to quote a string the caller never typed. Covers: the plain forms; an encoded prefixed name; the message naming the input rather than the expansion (and *not* containing the namespace); a malformed escape refused as `INVALID_INPUT` rather than escaping as a `URIError`; an unknown prefix offering the known ones; and the injection guard.

**The encoded form is also asserted through the server**, since that is where a client actually meets it and a unit test cannot see the template: `pragma:ds%3AComponent` must read byte-identically to `pragma:ds:Component`, and a malformed escape must still come back `InvalidParams`.

Full CLI suite: **1369 passed**, 0 failures. Package `check` green.

One note while adding these: `identity.test.ts` began flaking under the extra parallel load — it mutates `process.env.XDG_CONFIG_HOME` globally, the same shape of bug as the `create` suite's `process.chdir`. It passes solo. I kept my own additions to one store per file rather than per case so they add as little load as possible, but the underlying test-isolation issue is real and wants its own fix.

PR #1025, opened 2026-08-27
feat(pragma-cli): curate the MCP resource listing from declared slices

advl:
## Done

PR2 of the MCP-resources stack. Stacked on #1021 (`feat/mcp-resource-read-terms`).

**The problem, measured.** `resources/list` returned **712 entries / 158,495 bytes (154.8 KB)** on every client connect — the whole enriched pack index, all 604 individuals included. MCP defines a `cursor`/`nextCursor` contract for the listing, but SDK 1.27.1's high-level `McpServer` list handler **ignores `request.params.cursor` and never returns `nextCursor`** (verified on the wire), so there is no paging to fall back on: everything listed is sent, every connect. Agent context is a precious resource — that was ~155 KB of it spent before the agent had asked a question.

**After: 110 entries / 36,261 bytes (35.4 KB)** — a 77% cut, and the individuals lost nothing (they were never readable *from* the listing; they are reached by completing the `pragma:{+uri}` variable, which now offers all of them).

- **A declared listing (`mcpListable`).** A module declares which slices of the index it contributes — `{ type?, box?, as?, weight? }`, deliberately `CompletionSourceRef`'s vocabulary rather than a second one — and the listing is their union.
- **DERIVED, never authored twice.** Every read story ALREADY declares the type set it addresses (`PackLookup.type` / `types`) because that is what its name resolve is constrained by, so `compilePack` reads the listing off that same declaration: **zero new authoring for all six data nouns**. `compileStoryModule` is now the one place a story becomes a module, so the derivation cannot reach the distribution's static stories and miss the config/package ones. Only `ontology` declares by hand (`{ box: "tbox", as: "entities", weight: 0.5 }`) — "the whole TBox" is not a type filter.
- **`ds:Token` contributes nothing.** It is declared by a read story but SPARQL-served and never indexed; an empty collection would advertise a browse with nothing to show.
- **Declared weights.** `CLASS_PRIORITY = 0.9` / `INDIVIDUAL_PRIORITY = 0.3` were the kernel freezing editorial judgement about someone else's graph. They are replaced by `PackLookup.weights` (prefixed type → 0–1, unlisted default 1), which does **two** jobs from one declaration: `annotations.priority` on the listed resource, and a tiebreaker in `rankUriCompletions`. A `weights` key naming a type the lookup does not address is **rejected**, not ignored — a silent no-op is how a weight that never applied survives review.
- **Completion honesty.** `COMPLETION_LIMIT = 50` pre-truncated the ranked list, and the SDK derives `total`/`hasMore` from the array the callback returns — so `ds:` reported `{ values: 50, total: 50, hasMore: false }` while **499** entities matched. The limit is deleted; the full ranked list is returned and the SDK slices it. Completion still offers every entity: it is the discovery path for the individuals the listing no longer enumerates.
- **Human names.** 570 of 712 entries showed a bare URI with no description, and 383 of them had an unused `altNames[0]` ("Button", "FileTree") sitting in the index. `name` is now the stable prefixed URI; MCP's `title` carries the human label.
- **`pragma/type` in `_meta`** — an addition beside the frozen `pragma/box` / `pragma/instanceCount`, so an agent can filter a family without spending a read. `docs/mcp-integration.md` and the `surface.v2.json` `$comment` move with it.
- **`audience: ["user", "assistant"]`** — `["assistant"]` hides every resource from human pickers in clients that honour the field, and browsing a design system in a picker is exactly a human use case.
- **`pragma:sources` is readable.** It is the ONLY resource a cold server advertises, and reading it failed `INVALID_INPUT — Invalid uri "sources"`.
- **One `ENTITY_MIME_TYPE`** read by the registration, the read body and every listed entry — with #1021 switching reads to `text/turtle`, a listing still advertising `application/json` would be the same shape of failure as the `recipes:{+uri}` template over `pragma:` URIs.

**Not touched:** `mcpSurface.resources` keeps its `["pragma:{+uri}"]` template list. Collections are story/graph data, and the covenant's own precedent is explicit ("Prompt NAMES are graph data, guarded by the prompt tests, not frozen here"), so `emitSurface`/`surfaceConformance`/`surface.test.ts` stay out of this PR. The behaviour is guarded by tests instead.

**Drive-by:** `BUDGETS.md` gains the `resources/list` payload ceiling. The documented "MCP p95 (warm) < 100 ms" is not enforced by any test (the perf pass spawns `dist/src/bin.js` and measures CLI latency only), and payload *size* is the property that actually regressed.

### One thing that did not land as specified

The brief asked for a test pinning `ds:global.component.button` **first** for the query `"button"` once `ds:Subcomponent` is weighted `0.6`. That is not reachable with type-level weights, and the measurement says why:

| # | before (alphabetical ties) | after (`ds:Subcomponent: 0.6`) |
|---|---|---|
| 1 | `ds:apps_launchpad.component.button` `[ds:Component]` | `ds:apps_launchpad.component.button` `[ds:Component]` |
| 2 | `…launchpad.subcomponent.modal-…-close_button` | `ds:apps_lxd.component.code_snippet_with_copy_button` |
| 3 | `…launchpad.subcomponent.side_panel-…-close_button` | **`ds:global.component.button`** |
| 4 | `ds:apps_lxd.component.code_snippet_with_copy_button` | `ds:global.component.button_link` |
| 5 | **`ds:global.component.button`** | `ds:global.component.dropdown_button` |

Every subcomponent now sinks below every component, and the answer moves 5th → 3rd. It cannot reach 1st because the two entries above it are **also `ds:Component`** — only the *tier* (`global` vs `apps_launchpad`) separates them, and no type-level weight can express that. Getting to 1st needs a different rule (a tier preference, or scoring the local name segment), which I left out rather than smuggling in an unrequested heuristic tuned to hit a target. The test therefore asserts the properties that are true and derived: every component above every subcomponent, and the weighted rank strictly better than the unweighted one.

## QA

```bash
git fetch origin && git checkout feat/mcp-resource-listing
bun install
cd packages/cli/pragma && bunx vitest run src/capabilities/resources/resources.test.ts src/kernel/packs/compile.test.ts
```

Or over a live server:

```bash
cd packages/cli/pragma && bun run build
# then, from an MCP client, connect and call resources/list — 110 entries,
# 8 collections carrying pragma/instanceCount, no individuals
# and completion: complete the `uri` variable with "ds:" → total 499, hasMore true
```

Tests added in `resources.test.ts`: collections listed with counts and individuals absent; the listing-size budget asserted twice (off `buildResourceList` and off what actually crosses the wire); a declared story type reaching the listing WITHOUT the module authoring it (read off the compiled module, so an agreement that merely holds today cannot satisfy it); a declared type the index does not know contributing nothing; `total`/`hasMore` truthful for a >100-match query; `title` from `altNames`; `pragma/type` present on every entry; `pragma:sources` reading successfully on a cold store; the weight reaching both `annotations.priority` and the completion rank. In `compile.test.ts`: the derivation itself, and both `weights` rejections (out of range, and a key naming an unaddressed type).

**Gate** (from the repo root):

- `bun run check` — **PASS**, all 62 projects.
- `bun run test` — the CLI package is `1181 passed / 3 failed`; all three are the known `create` 5-second timeouts under parallel load, and all three pass when the suites are run solo. `svelte-ds-global` / `svelte-ds-app` / `svelte-ds-app-wpe` fail on missing Playwright browser binaries in a fresh worktree (`Executable doesn't exist at …/ms-playwright/firefox-1532/firefox`) — environment-only, discountable per `AGENTS.md`.

### PR readiness check

## Screenshots

Not applicable — MCP wire surface, no visual change. The measurement is the artefact:

| | entries | `resources/list` bytes |
| --- | ---: | ---: |
| before | 712 | 158,495 (154.8 KB) |
| after | 110 | 36,261 (35.4 KB) |


advl:
Thanks — all four were real. Fixed in 1b53d9a, each pinned by a test.

**1. Absolute-IRI lookup types** (`compile.ts`) — correct, and the weight keys had it too. `PackLookup.type`/`types` are documented "prefixed name or IRI", so a story constrained to `https://…/Widget` compiled clean and then matched no class *and* no weight. Types and weight keys are now compacted against the same map `resolveUri` expands with, so the two directions agree. Test asserts an absolute type and its weight survive together.

**2. Missing count on an empty collection** — correct, and the consequence is the sharper half of your point: without a default, an empty collection is indistinguishable from an entry that is not a collection at all. A declared collection whose class the index knows now always states a count, zero included.

**3. Weight from the primary type only** — correct, including the drift you flagged: ranking could move when an unrelated, lexically earlier type was added. The effective weight is now taken across every declared membership.

On the aggregation you left open, I chose **the lowest declared weight**, with unweighted types not voting. A weight below the default is a demotion, and a demotion one membership asks for should not be cancelled by another that never asked for anything. The test separates two entities identical but for a secondary type's weight, so it fails if the primary-only read comes back.

**4. `String.length` vs UTF-8 bytes** — correct. Measured: today's listing undercounts by **4 bytes** (36,261 vs 36,265), so nothing is currently over the line, but the assertion claims to budget bytes and a graph with non-ASCII labels could pass while exceeding the 60 KB ceiling. Both sites now use `Buffer.byteLength`.

Full CLI suite: **1187 passed**, 0 failures.

PR #1030, opened 2026-08-27
refactor(svelte-ds-app-wpe): conflate the dso: prefix into ds:

advl:
## Done

Split out of #1029 so it can get your review on its own rather than dragging
`@canonical/workplace-engineering-ui` onto a 100-file integration PR.

The design system declared its ontology terms under two prefixes for the same
namespace — `ds:` and `dso:`. `dso:` is retired. It matters here because the
implementation collector only recognises `ds:`: a `dso:` annotation is silently
invisible to the knowledge graph, so whole component families were being dropped
without any error.

- Five `@implements` annotations renamed `dso:` → `ds:` (three `.svelte`, two `.css`).
- **Comment-only.** No behaviour change, no export change, no rendered output change.
- `ds-app-wpe` declares no `design-system.json`, so it is not collected today —
  this is preparation, not a graph change. Adding the package to the collector
  would be a separate, opt-in change for you to decide on.

Fixes nothing; unblocks #1029.

## QA

```bash
git diff main...refactor/dso-prefix-ds-app-wpe   # 5 files, 5 insertions, 5 deletions
grep -rn "@implements dso:" packages/svelte/ds-app-wpe/   # no matches
cd packages/svelte/ds-app-wpe && bun run check && bun run test
```

### PR readiness check

- [x] PR should have one of the following labels
- [x] No new package in this PR
- [x] If this PR does not require visual testing, add the `no visual change` label

🤖 Posted by Claude Code

### DIFF
--- a/packages/svelte/ds-app-wpe/src/lib/components/KeyboardKey/KeyboardKey.svelte
+++ b/packages/svelte/ds-app-wpe/src/lib/components/KeyboardKey/KeyboardKey.svelte
@@ -37,5 +37,5 @@ non-interactive and purely informational.
 <KeyboardKey keyValue="enter" />
 ```
 
-@implements dso:global.component.keyboard_key
+@implements ds:global.component.keyboard_key
 -->
--- a/packages/svelte/ds-app-wpe/src/lib/group/Cards/Cards.svelte
+++ b/packages/svelte/ds-app-wpe/src/lib/group/Cards/Cards.svelte
@@ -43,5 +43,5 @@ master columns. Outside a grid it degrades to a plain block.
 </div>
 ```
 
-@implements dso:global.group.cards
+@implements ds:global.group.cards
 -->
--- a/packages/svelte/ds-app-wpe/src/lib/group/Cards/styles.css
+++ b/packages/svelte/ds-app-wpe/src/lib/group/Cards/styles.css
@@ -1,6 +1,6 @@
 /**
  * Cards styles
- * @implements dso:global.group.cards
+ * @implements ds:global.group.cards
  *
  * A two-level subgrid: the parent `.grid` supplies columns; `Cards` inherits
  * them and defines four shared section row tracks (image / header / content /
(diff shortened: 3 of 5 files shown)


jmuzina (approved):
Should we also apply this to the 4 instances of `dso:` in the global svelte package?

Image

advl:
Checked — those four are already done, by #1029 (`6e865b981`), which is an ancestor of this PR's merge commit.

At `6e865b981^` there were exactly **4** `dso:` instances in `packages/svelte/ds-global`, in precisely the files you counted:

```
Breadcrumbs/styles.css:3              @implements dso:global.pattern.breadcrumbs
Breadcrumbs/types.ts:35               @implements dso:global.pattern.breadcrumbs
Breadcrumbs/common/Item/styles.css:3  @implements dso:global.subcomponent.breadcrumbs-item
Breadcrumbs/common/Item/types.ts:21   @implements dso:global.subcomponent.breadcrumbs-item
```

#1029 renamed all four **and** corrected the parent from `pattern.breadcrumbs` to `component.breadcrumbs` — a classification fix a pure prefix rename would have missed.

Timing explains the near-miss: #1029 merged at 22:04Z and your review landed at 23:20Z, so it was accurate against this branch (cut before #1029) but already resolved on `main`.

Two things worth recording while we were in here:

- **Why `ds-global` was normalised earlier than `ds-app-wpe`:** `scripts/collect-implementations.ts` only harvests packages carrying a `design-system.json`. `svelte/ds-global` has one; `svelte/ds-app-wpe` does not — so this PR's annotations are never collected into `data/`, and no regeneration was needed here. `data/svelte-ds-global/implementationObjects.ttl` is in sync.
- **Repo-wide there is now exactly one `dso:` left**, in `packages/cli/pragma/src/kernel/runtime/graphpack/embedded/pack.generated.ts` — a harvested upstream prefix declaration for `https://ds.canonical.com/ontology#` inside a generated artifact. That is a genuinely different namespace, not a retired alias, so it is correct as-is and will refresh with the next pack regeneration.

No follow-up PR needed.

PR #1037, opened 2026-08-27
refactor(pragma-cli): route kernel cross-domain imports through domain barrels

advl:
## Done

Routes the cross-domain imports **originating in `packages/cli/pragma/src/kernel/**`** through the sibling domain's barrel, instead of naming a file inside it.

`cs:packaging.import.cross_domain` and `cs:packaging.export.barrel` both say the same thing: a cross-domain import goes through the domain's `index.ts`, never `../<domain>/<file>.js`. PR #1026 added the missing barrels; this is the rerouting half that was deliberately deferred with them.

- **84 import statements in 44 files** rewritten from `../<domain>/<file>.js` to `../<domain>/index.js`.
- **No barrel's export surface is widened.** A reach is rerouted only when the target barrel already exports every symbol the site imports; everything else is left exactly as it was and reported below.
- Mechanical only — no renames, no export-shape changes, no constants/types moves, no new lint rule or gate. `src/testing/**` is untouched.
- Biome's `organizeImports` merged the import lines that became duplicates of one another; that is the whole of the non-specifier diff.

### Laziness is preserved, and it constrained the change

Three barrels carry an explicit warning that the module behind them is reached by a **dynamic** import so Ink/React, the config reader and the MCP SDK stay off the `--help`/`__complete` fast path (`buildSetupRun`, `loadEffectiveModules`, the MCP prompt source). Those are dynamic `import(...)` calls with no `from`, so no reroute could touch them — but the wider hazard is real: a barrel re-exports its whole domain, so routing a cheap leaf import through it can drag a heavy sibling onto the static graph.

Every candidate reroute was therefore applied **one at a time against the repo's own guards** — `src/capabilities/lazy.test.ts` and `src/kernel/completion/safety.test.ts` — and reverted when it made zod, `kernel/config/readConfig.ts`, `commander`, `@canonical/ke` or `kernel/runtime/boot.ts` statically reachable from `capabilities/index.ts`, `buildProgram.ts` or the completion entries. **23 of 247 reaches were rejected that way** (13 of them in this slice) and are left as direct file imports.

18 of the 23 are `import type` — erased at compile time, so free at runtime — but the guards walk `from "…"` textually and, as their own comments state, cannot tell an `import type` from a value import. They are consequently un-reroutable while the guard reads imports textually. The remaining 5 are genuine value-import hazards (`suggestNames` pulling `commander` via `buildProgram`, `distributionSource`/`resolvePackDetail` pulling `packs/schema.ts`'s zod, `resolveUri` pulling the config reader). Both sets are recorded here rather than fixed by widening a barrel or relaxing a guard.

### Slicing

This is one of two file-disjoint slices, split by importing domain so each lands independently and neither conflicts with the other:

- this PR — importers under `src/kernel/**` (44 files)
- `refactor/cross-domain-barrel-imports-capabilities` — importers under `src/capabilities/**` (83 files)

## QA

```bash
bun install
# package loop, cache explicitly skipped so nx cannot serve a stale pass
cd packages/cli/pragma && bun run check && NX_SKIP_NX_CACHE=true bun run test
# gate of record
cd ../../.. && NX_SKIP_NX_CACHE=true bun run check && NX_SKIP_NX_CACHE=true bun run test -- --no-bail
```

- `bun run check` (biome + `tsc --noEmit` + webarchitect): green, root and package.
- `NX_SKIP_NX_CACHE=true bun run test` in `packages/cli/pragma`: **1463 passed / 13 skipped / 5 todo** across 130 files, plus the perf suite **22/22**.
- **Perf budgets pass unchanged** — `src/testing/perf/budgets.test.ts`: `--help`, `__complete`, `__complete skill lookup`, warm store-backed verb, warm in-process MCP call and warm project-config load all under budget. No budget was touched.
- The root `bun run test` also reports failures in `@canonical/svelte-ds-global`, `-app` and `-app-wpe` (Playwright Firefox/WebKit are not installed in this environment) and intermittent 5000 ms timeouts in pragma's behavioural/MCP suites under Nx parallel load — a **different set on each run**, and all green when `packages/cli/pragma` runs on its own. Environmental, not from this diff.

### PR readiness check

- [x] No new package
- [x] `no visual change` applied — this PR changes import specifiers only

## Screenshots

Not applicable — no runtime behaviour or output changes.

---


advl:
Sibling slice: #1038 routes the same reaches for importers under `src/capabilities/**`. The two are file-disjoint by construction, so they land independently and in either order.

---

PR #1061, opened 2026-08-29
test(pragma-cli): make the interactive cancel + no-freeze guarantees executable under a real pty

advl:
## Done

The interactive-cancel and no-freeze guarantees (E4 / AV-231) are now **executable**: a pseudo-terminal test layer replaces the five scaffolded `it.todo`s in `journeys.interactiveTty.e2e.test.ts` with real PTY-driven journeys that spawn the shipped `dist/src/bin.js` under a terminal, type raw keystrokes, and assert stderr content + exit codes against `EXIT` (`kernel/project/cli/exitCodes.ts`).

**The pty layer** (`src/testing/behavioral/ptyDriver.ts`) drives the binary under util-linux **`script(1)`**, not `node-pty`:

- `node-pty` has no Linux prebuilds — `bun add node-pty` runs `node-gyp rebuild` at install (needs python3 + a C++ toolchain on every dev machine, plus a root-level `trustedDependencies` entry for bun to run the install script at all). It fails outright on hosts without a compiler profile.
- The prebuilt fork (`@homebridge/node-pty-prebuilt-multiarch`) does install and load, but is a third-party fork that fetches a platform binary from GitHub releases at install time and still needs the root `trustedDependencies` edit — new failure modes in shared install infrastructure for a suite that only runs on Linux CI.
- `script -qec … /dev/null` allocates the same kernel pty with **zero dependency-graph changes** (nothing new near the runtime graph, the lazy-graph guard, or the perf fast paths), is preinstalled on ubuntu-latest (`bsdutils`, an Essential package built from util-linux), and `-e` propagates the child's exit code including 130.

What the five journeys prove (each shown red against a deliberately broken variant — see QA):

- **H1** — Ctrl-C at a wizard prompt: `Cancelled.` on stderr, exit 0, nothing written, and the transcript proves execution was never entered.
- **H2** — Ctrl-C during execution: `Cancelled.`, exit **130**, transcript proves the run entered `Generating` and never completed. The interrupt window is made deterministic with an injected `fs.promises.writeFile` latency shim (`NODE_OPTIONS --require`) — the subject's control flow is untouched; it just runs on a slow disk.
- **H3** — `create 2>stderr.txt` with stdin still a TTY: the refusal (usage error, exit 2) instead of an invisible wizard blocked on stdin.
- **C3** — EOF (`\x04`) at a prompt resolves to a usage error naming the unanswered prompt, exit 2 — **this required a product fix**, see below.
- **C4** — Enter at the shipped select (`create package` → `type`) submits the highlighted row and the wizard advances to the next prompt; clean Ctrl-C teardown asserted after.

**Drive-by that is really the headline: C3 was not true before this PR.** The scaffold recorded EOF-at-prompt as a guarantee "the CLI already promises", but the wizard had **no `\x04` handling at all**: Ink handed the byte to the focused text input as a literal character and the run hung until killed (reproduced under a real pty before writing the fix). Fixed in summon-core: `SessionController.eof()` rejects the pending prompt with `MISSING_REQUIRED_ANSWER` (the same code the non-interactive strategy raises for an absent answer, so the CLI boundary maps it to a usage error, exit 2) and aborts the run; the Wizard binds Ctrl-D to it. Covered in-process in `session.test.ts` and `wizard.test.tsx`, and end-to-end by the C3 journey. An EOF is deliberately **not** a decline (H1's exit 0) and not an interrupt (H2's 130).

Two deliberate rulings, written into the files:

- **A missing pty driver FAILS, not skips.** On Linux (every CI lane) `script(1)`'s absence or a failing smoke probe throws from `beforeAll`, failing all five journeys with the reason — the scaffold's `describe.skip` gate could let the guarantee silently stop running. Non-Linux platforms skip with the reason printed (BSD `script` is flag-incompatible; CI exercises the suite on Linux regardless).
- **No new CI job.** The journeys run as their own serial pass (`vitest.tty.config.ts`, `test:tty`) chained into the package's `test` target, exactly the `test:perf` precedent — excluded from the parallel coverage pass, where pty deadline waits are starved into flakes.

Stability (the suite must not inherit `crossCli.subprocess.test.ts`'s failure modes): content is asserted **before** exit codes; every wait shares one generous deadline whose expiry rejects **with the captured transcript** (never a bare assertion mismatch); the driver kills and reaps its child on every path and `afterEach` sweeps stragglers; the harness survives its subject hanging (that is exactly what H3/C3/C4 test — verified by red runs).

Noted, not fixed here (its own ledger item): `crossCli.subprocess.test.ts` fails opaquely because it asserts `.status` before stderr (`expect(pragma.status).toBe(2)` at e.g. lines 388/435/490 → "expected null to be 2" on a spawn timeout) and its `helpOf`/producer helpers drop `spawnSync`'s `.error` (ETIMEDOUT) entirely. The cheap diagnosability fix is to assert the captured stderr/text first (or fold the buffers + `.error` into the status assertion's message).

Also updated the now-stale "no suite can drive a real TTY through a subprocess" comment in `create/mount.ts`.

## QA

- `bun run check && bun run test:vitest && bun run test:perf && bun run test:tty` in `packages/cli/pragma`; `bun run check && bun run test` in `packages/summon/core`.
- Red-proofs (run locally, then reverted — each journey fails diagnosably against a broken subject):
  - interrupt exit collapsed to 0 in `dispatch.ts` → H2 fails `expected +0 to be 130`;
  - `canPrompt` gating on stdin only → H3's invisible wizard hangs; the driver's deadline kills it and fails with the transcript;
  - Ctrl-D binding stripped from the built wizard → C3 hangs; same diagnosed failure.
- `bun run gen:reference` produces no diff.
- Interactive wizard still behaves normally on a real terminal (H1/H2/C3/C4 transcripts in the test output are the wizard's actual frames).

### PR readiness check

  - [x] Packages with build steps: `build` and `build:all`.

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_011B8Z3Lq1eARN1wLfKGvzqC

### DIFF
--- a/packages/cli/pragma/package.json
+++ b/packages/cli/pragma/package.json
@@ -41,9 +41,10 @@
     "check:biome": "biome check",
     "check:biome:fix": "biome check --write",
     "check:ts": "tsc --noEmit",
-    "test": "bun run test:vitest && bun run test:perf",
+    "test": "bun run test:vitest && bun run test:perf && bun run test:tty",
     "test:vitest": "vitest run --coverage",
     "test:perf": "vitest run --config vitest.perf.config.ts",
+    "test:tty": "vitest run --config vitest.tty.config.ts",
     "test:watch": "vitest",
     "test:coverage": "vitest run --coverage",
     "eval": "bun run src/testing/eval/report.ts"
--- a/packages/cli/pragma/src/capabilities/create/mount.ts
+++ b/packages/cli/pragma/src/capabilities/create/mount.ts
@@ -175,8 +175,9 @@ export function explicitLeafAnswers(
 
 /**
  * The mount's mode resolution — `decideInteraction` over the leaf's five
- * inputs, with the TTY fact INJECTED so the resolution is testable with
- * `tty: true` (no suite can drive a real TTY through a subprocess). The
+ * inputs, with the TTY fact INJECTED so the resolution is testable in-process
+ * with `tty: true` (the real-terminal half is exercised end to end by the
+ * pty journeys in `testing/behavioral/journeys.interactiveTty.e2e.test.ts`). The
  * action calls it with {@link canPrompt} — the same exported H3 gate the
  * kernel's interaction context reads — so the mount and `runCreate` can
  * never disagree about interactivity.
--- a/packages/cli/pragma/src/testing/behavioral/journeys.interactiveTty.e2e.test.ts
+++ b/packages/cli/pragma/src/testing/behavioral/journeys.interactiveTty.e2e.test.ts
@@ -1,66 +1,269 @@
 /**
- * E4 (AV-231, Backlog E) — the interactive TTY journey. SCAFFOLD.
+ * E4 (AV-231, Backlog E) — the interactive TTY journeys: cancel + no-freeze,
+ * driven end to end through a REAL pseudo-terminal.
  *
- * No test in the suite ever drove a REAL terminal, so the interactive-cancel and
- * no-freeze guarantees (H1–H3, C3/C4) were never exercised end to end. Confirming
- * them needs a pseudo-terminal: spawn the compiled `pragma` under a PTY, send raw
- * keystrokes (Ctrl-C = `\x03`, EOF = `\x04`), and observe the exit code + output.
+ * Until this suite, no test ever drove the shipped binary under a terminal:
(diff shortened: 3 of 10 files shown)


advl:
Closing — superseded; the interactive-journey work will be reconsidered separately.

PR #1067, opened 2026-08-30
chore: reconcile feat/prism-docsite with main — main's side for everything but the docsite

advl:
Implements rulings **REC-1** and **REC-2** of 2026-08-30: take main's `packages/cli/pragma` 100%, take main's side for the rest, keep the docsite branch as pure as possible.

This is the reconciliation the program has been deferring. The branch was **101 ahead / 111 behind** `main`, and the gap was growing daily.

## The census

292 conflicts. **276 resolved to main's side; 16 were files main had deleted, and are deleted here.**

| Conflicts | Area | Resolution |
|---|---|---|
| 141 | `packages/cli/pragma` | main's, wholesale — REC-1 |
| 127 | CHANGELOG / package.json / lock / lerna | main's — release churn |
| 26 | workflows, docs, router, boilerplate, svelte | main's — REC-2 |
| 2 | `packages/runtime/ke-graphql` | CHANGELOG and package.json **only** |
| **0** | `packages/docsite/**` | — |
| **0** | `apps/react/pragma-docs/**` | — |

**The last two rows are the finding.** The two lineages collided almost entirely in the CLI — two independent rewrites of one package, which is exactly what B.05 and B.01 each produced on their own side. The despecialization's own surface never diverged, because main never touched it. ke-graphql's *source* has zero conflicts; only its release metadata moved.

That also means the risk in this merge is not textual but semantic: the docsite program now sits on top of main's CLI rather than its own. CI is the judge of that, and it is why this PR exists rather than a direct push.

## Two deliberate exceptions to "take main's side"

1. **`.github/workflows/pr.yml`** — main has no knowledge of the Gates step added in #1064, so taking main's file verbatim silently deleted it. Re-applied on top of main's version, so both despecialization gates still run on every PR. Without this, G-2 goes back to sleep and nobody notices.
2. **Nothing else.** In particular the CLI's `test` script is main's verbatim, which already carries the perf ruling — #1066 landed on main while this merge was being resolved, so the two agree without special handling.

## What this unblocks

`feat/prism-docsite` → `main` becomes a reviewable step rather than an archaeological dig. That final merge is deliberately **not** taken here: it is the one handoff reserved for the owner.


advl:
## Update — the rule applied properly, and one carve-out

Two rounds of CI found two things, one a flaw in how I applied REC-2 and one a genuine second collision.

**1. Resolving conflicts to main's side was too weak a reading.** Where main changed one file and the branch changed a different file in the same package, git auto-merged and produced a hybrid — main's router *types* with the branch's router *source*, which cannot typecheck. Nothing conflicted; the merge was clean and wrong. The tree is now **main's, plus the docsite paths** (`packages/docsite/**`, `apps/react/pragma-docs/**`, `packages/runtime/ke-graphql`), which is REC-2 as stated rather than as conflicts happened to allow.

**2. `bun.lock` has no "main's side" to take.** The merged tree contains three workspace packages main has never seen, so main's lockfile describes a workspace that does not exist. A lockfile is derived, not authored — regenerated. That single mistake accounted for three of the four failures in the first round, all of them jobs installing with `--frozen-lockfile`.

**3. The router is a second collision, and this PR does not resolve it.** Both lineages rewrote the router:

| | branch | main |
|---|---|---|
| route declaration | `component:` on the route input | reshaped by #880; `component` is not a member of `DataRouteInput` |
| construction | `createHashRouter`, `createStaticRouter` | neither exported; `createRouter` |

The docsite is written against the branch's API — **98 files** use a `component:` route field, **22** use `createStaticRouter`. Taking main's router breaks the app; keeping the branch's leaves a rival implementation of a shared package alive here, which is the same shape as the CLI collision REC-1 was written to end.

**That is a ruling, not a merge decision, so this PR does not make it.** `packages/runtime/router`, `packages/react/router` and `packages/react/ssr` keep the branch's version; everything else outside the docsite is main's. With the router restored and rebuilt, the docsite typechecks clean locally.

The three ways out — port the docsite to main's router, carve the router out for good, or decide which router survives on merit — are recorded with their costs in `canonical/pragma-adrs` `session/B/B.04` §9.7. If the choice is not made before then, it moves to the `feat/prism-docsite` → `main` merge, with more commits on top.

advl:
## The three failing Chromatic UI Tests are the intended effect of this PR, and they are safe to approve

Posting the evidence rather than asking for a rubber stamp.

### What is failing

| Check | Chromatic build |
|---|---|
| `UI Tests: packages/react/ds-global` | [890](https://www.chromatic.com/build?appId=67be30a0b85c4e44b6744158&number=890) |
| `UI Tests: packages/svelte/ds-app` | [126](https://www.chromatic.com/build?appId=6a3ca6a1c2cb863811b05c34&number=126) |
| `UI Tests: packages/svelte/ds-global` | [209](https://www.chromatic.com/build?appId=6a3ca65a785e76b915a40851&number=209) |

Ten `UI Review:` statuses are `pending` behind them and clear once these are decided. Every other Chromatic check — all 16 `Storybook Publish` jobs and the remaining `UI Tests` — is already green.

### Why they changed, and why that is correct

These are **design-system** packages. This PR does not edit them:

```
$ git diff --stat origin/main origin/claude/reconcile-main -- \
    packages/react/ds-global packages/svelte/ds-app packages/svelte/ds-global
(empty)
```

**All three are byte-identical to `main`.**

Chromatic compares against this PR's *merge target*, `feat/prism-docsite`, not against `main` — and that branch was **10 commits behind main** on these packages:

```
$ git log --oneline origin/feat/prism-docsite..origin/main -- \
    packages/react/ds-global packages/svelte/ds-global
78eb47ac1 chore: version bump to v0.36.0
5198c5399 chore: version bump to v0.35.0
acfa3ca24 refactor(storybook): consolidate hand-rolled router decorators onto withHashRouter (#996)
416d59636 feat(router)!: pre-1.0 API consolidation (re-land of #973) (#981)
b1f6b4fdb feat(svelte-ds-global): Upstream `SkipLink` from WPE tier to Global tier (#859)
… 10 total
```

So the diffs are precisely *"what landed on main since this branch forked"*, which is what a reconciliation is for.

### What you would be approving

| Package | Components with visual changes |
|---|---|
| `react/ds-global` | Accordion, Breadcrumbs, Cards, ContextualMenu, Icon, InlineCode, KeyboardKey, KeyboardKeys, Popover, Spinner, Tabs, Timeline, Tooltip |
| `svelte/ds-global` | Breadcrumbs |
| `svelte/ds-app` | none — version and CHANGELOG only |

**Approving these accepts exactly the state that is already on `main`.** No visual state originates in this PR; the snapshots are the ones `main` has been serving since v0.36.0. If any single diff looks wrong, it is wrong *on main too* and should be raised there rather than blocked here.

### Everything else on this PR

`red = 0`. The two `build-matrix` jobs are the last non-Chromatic checks and are running. The earlier failures are fixed and recorded:

- **`cda5e39ce`** — the committed graphpack was artifact **v1** beside this branch's **v2** compiler (27 failures). Regenerating also required anatomy-dsl **0.3.0**, because 0.2.2 cannot compile against a v2 compiler at all (`M005`: a bare `uri` colliding with the structural `uri: ID!`).
- **`cc08f4990`** — main's `Outlet` runs route components' hooks in its own hook list, so navigating between routes with differing hook counts threw *"Rendered fewer hooks than expected"*. Fixed and cherry-picked here; it targets `main` as #1073.

The app suite on this branch is **98 files / 461 tests, all green** — it was two failing.

Ledger: `canonical/pragma-adrs` `session/D`.

advl:
## Correction to the Chromatic analysis above — `main` has moved

My earlier comment claimed the three packages with failing UI Tests are **byte-identical to `main`**. That was true when written, against `main` at `40d2ad7bd`. It is no longer literally true, and since the comment exists to let someone approve quickly, the stale half should not stand.

**What changed.** `main` advanced to `5bad77e66` with #961, *"feat(svelte-ds-global): Upstream Svelte `Announcement` from WPE tier to Global"*, which touched two of the three packages:

| Package | vs `main` now |
|---|---|
| `packages/svelte/ds-global` | **8 files** — the branch lacks the upstreamed `Announcement` (7 files + the `_work_in_progress` barrel entry) |
| `packages/react/ds-global` | **1 file** — `Announcement/styles.css`: `gap: 0` → `var(--dimension-050)`, and the heading font-weight token |
| `packages/svelte/ds-app` | still identical |

So the accurate statement is: **the three packages are identical to `main` as of `40d2ad7bd`, and are now one commit behind it.**

**The conclusion is unchanged, and here is why.** Chromatic compares this PR against its *merge target*, `feat/prism-docsite` — not against `main`. #961 landing on `main` does not alter this branch's snapshots or its baseline. The three failing UI Tests are still exactly *"what landed on `main` since this branch forked"*; that set has simply grown by one commit, which this branch does not yet carry.

**What this does mean:** the branch is now one commit behind `main`, and #961 will arrive either through a refresh of this branch or at the final merge (rung D-5). It is ordinary drift of the kind the reconciliation exists to absorb, and one commit is not worth re-running the whole Chromatic suite for right now — but it should be picked up before the final merge rather than discovered there.

The rest of the earlier comment stands: every other Chromatic check is green, ten `UI Review` statuses clear behind these three, and `red = 0` on everything non-Chromatic.

PR #1088, opened 2026-09-01
chore(deps): update @biomejs/biome to 2.5.11

advl:
## Done

Updates `@biomejs/biome` from 2.4.9 to **2.5.11** (the newest published release) across every workspace, and adapts the repo to what 2.5 changed. This **supersedes #786**, the Renovate PR for the same bump, which failed `build-matrix` on both Node 22 and 24.

### Why #786 failed

Two independent problems, both visible in [its build-matrix log](https://github.com/canonical/pragma/actions/runs/33534802932):

1. **The `$schema` string was left at 2.4.9.** `@canonical/biome-config:check` reported `biome.json:2:14 deserialize — The configuration schema version does not match the CLI version 2.5.11 … Expected: 2.5.11, Found: 2.4.9`. Renovate bumps the dependency but not the schema URL inside the config it ships.
2. **2.5 changed rule coverage and formatter output**, which #786 did not account for. Five projects failed `check` outright (`ds-assets`, `storybook-addon-shell-theme`, `task`, `storybook-helpers`, `ke-graphql`), and because Nx bails on first failure, another 40 never ran — so the log understates the work. Running the full gate locally surfaced four more (`styles`, `summon-core`, `pragma-cli`, `react-ds-global`, `lit-ds-prototype`, `svelte-ds-app-launchpad`).

### Toolchain changes

- `@biomejs/biome` `2.4.9` → `2.5.11` in all 61 workspaces, in the `@canonical/biome-config` peer range, and in the summon React application template.
- `$schema` in `configs/biome/biome.json` moved to `2.5.11`, in lockstep.
- `linter.rules.recommended: true` → `linter.rules.preset: "recommended"`, produced by `biome migrate`. The old field is deprecated in 2.5 and removed in the next major. Identical rule set.
- `configs/biome/README.md` still advertised a `1.9.4` schema URL in its usage example; refreshed to match (drive-by, one line).

**`@canonical/biome-config` did not need a separate bump.** It is a workspace package here (`configs/biome`), not an external dependency — Renovate groups the two because the config is *published* from this repo. Its peer range moves with the same commit, and Lerna will version and release it as part of the normal cycle.

### Rule and formatter fallout from 2.5

- **`correctness/noUnsafeOptionalChaining` now sees through a TypeScript `as` cast.** `(result.data?.field as T).prop` was invisible to 2.4 and is flagged by 2.5 — same config, same files, 0 → 61 diagnostics. Every one is a GraphQL test assertion against `ExecutionResult["data"]`, which is typed `| null | undefined`. **Fixed, not suppressed** — see the drive-by below.
- **`correctness/noUnsafeOptionalChaining` now sees through a TypeScript `as` cast.** `(result.data?.field as T).prop` was invisible to 2.4 and is flagged by 2.5 — same config, same files, 0 → 61 diagnostics. Every one is a GraphQL test assertion against `ExecutionResult["data"]`, which is typed `| null | undefined`. Suppressed at file scope in the six `ke-graphql` test files and at one site in the CLI setup test, each with the reason written inline. The rule is right that these can throw; in a test an absent result fails the run either way, so the suppression is deliberate debt rather than a fix, and rewriting the assertions is worth a follow-up outside a toolchain bump.
- **`a11y/noAmbiguousAnchorText` is new** and flags `<a href="#example">Learn more</a>` in an Accordion story. **Fixed, not suppressed** — see the drive-by below.
- **Nine `biome-ignore lint/style/useExportType` comments are now dead.** They were written against a false positive on Svelte "cross-language" exports that 2.5 fixed, so they now fail as `suppressions/unused` errors. Removed.
- **Formatter output changed** for curried calls (`it.each(X)(desc, fn)` no longer breaks the first argument list), CSS selector-list indentation, and blank lines after import blocks. Applied with `bun run check:fix` and reviewed: whitespace only, no semantic edits. `git diff -w` over the CSS and component files is empty.

### Drive-by: the two new-rule suppressions, fixed properly

Bundled deliberately rather than deferred: leaving a rule off is cheap to write and expensive to reverse, and both of these were one commit's work. Called out here per `AGENTS.md`.

**`correctness/noUnsafeOptionalChaining` — all 61 sites fixed, all 7 suppression comments deleted.** The pattern was `(result.data?.field as T).prop`: the optional chain says `data` may be nullish, the cast then asserts it is not, and a property is read straight off it. When `data` really is absent that surfaces as `TypeError: Cannot read properties of undefined`, several reads away from the cause.

- Adds `assertPresent` / `assertData` to `ke-graphql`'s internal `src/testing/` surface — narrowing assertions (`asserts value is NonNullable<T>`) that check first and let the read proceed without the chain. `assertData` also prints the result's `errors`, which is what actually explains an absent `data`. Test-support only: `src/testing/**` is excluded from both the build and the coverage report, so nothing ships and no threshold moves.
- The six `// biome-ignore-all` file headers in `ke-graphql` are gone; each affected block now asserts, then reads. Within a block that gained an assertion, the dead `?.` on that same value is dropped too, so the block reads consistently.
- No non-null assertions were used — `noNonNullAssertion` is enforced here, and trading one violation for another is a known trap in this repo.
- The one inline suppression, in `packages/cli/pragma/src/capabilities/setup/setup.test.ts`, needed no helper at all: `PromptDefinition["choices"]` is already `ReadonlyArray<{ label: string; value: string }>`, so the `as { label: string }[]` cast was never buying anything. `targets?.choices?.every(…)` is a safe optional call — absent choices yield `undefined`, and `expect(undefined).toBe(true)` fails cleanly instead of throwing.
- No assertion was weakened to satisfy the rule. Every test still fails for exactly what it failed for before, with a better message when the cause is an absent result.

**`a11y/noAmbiguousAnchorText` — fixed, suppression deleted.** The Accordion `WithRichContent` story said `Learn more`, which tells a screen-reader user navigating by links nothing about where the link goes. Reworded to `Usage examples`, which describes the `#example` destination and is close enough in length to keep the visual baseline honest.

**Not touched, on purpose: the `noSvgWithoutTitle` override for `**/*.svg`.** That one is a correct decision, not debt — an SVG `<title>` renders as a hover tooltip, and adding one to 165 decorative icons and brand marks would make the UX worse. The rule stays on for inline JSX `<svg>`, where the accessible name genuinely belongs.

## QA

Run from the repo root on this branch:

```bash
bun install
bun run check   # 62 projects
bun run test
```

- `bun run check` passes for all **62 projects** (exit 0), from a clean install on a fresh worktree.
- `bun run test` is **not clean on my machine**, and I want to be straight about that rather than claim a green gate. Three full root runs each exited non-zero naming a *different* set of projects, and every single failure is `Error: Test timed out in 5000ms` — 54 timeouts and **0 assertion errors** in the last run, on tests whose baselines are milliseconds. Every package the diff touches (`ke-graphql`, `react-ds-global`, `svelte-ds-app-launchpad`, `task`, `summon-core`, `storybook-helpers`, `react-ds-global-form`) passes in isolation. This looks like local machine contention, not a real regression, but **`build-matrix` on this PR is the verdict of record** — please read it rather than my word.
- Behaviour worth a second pair of eyes: the two `a11y` suppressions (SVG assets, story anchor text) are judgement calls about where an accessible name belongs. If the team would rather add `<title>` to the assets, that is a different and much larger PR.

## Follow-ups (not done here, deliberately)

- ~~Rewrite the 62 `(data?.field as T).prop` assertions in `ke-graphql` and the CLI so `noUnsafeOptionalChaining` can be re-enabled there.~~ Done in this PR — see the drive-by above.
- 2.5 also emits several new **warnings** (`complexity/useOptionalChain`, `suspicious/noTemplateCurlyInString`, `correctness/noUnusedImports`) in `ke-graphql`, `summon-core` and `pragma-cli`. Warnings do not fail `biome check`, and fixing them means touching production narrowing logic, so they are out of scope for a dependency bump.
- `packages/storybook/helpers/vitest.setup.js` and `vitest.setup.d.ts` are committed build artifacts sitting next to their own `vitest.setup.ts`. 2.5's assist rules lint them; this PR just adds the blank line the formatter wants. They should probably be deleted and gitignored.

## Notes for the reviewer

- **#786 should be closed** in favour of this PR. I have deliberately not touched the `renovate/biome` branch or closed it — Renovate owns that branch and would rebase over any edit.
- The `bun.lock` diff is 70 lines, all `@biomejs/biome` version strings, and nothing else.

## Screenshots

Not applicable in substance — Biome is a lint/format toolchain and cannot change rendered output, and every other source edit here is whitespace, comments, or suppressions. **One exception:** the drive-by above rewords one link inside the Accordion `WithRichContent` story, `Learn more` → `Usage examples`. That is four characters wider in a single story and the only rendered-output change in the diff. `Chromatic: skip` is still applied — remove the label if you would rather have that one story re-baselined.

### PR readiness check

  - [x] Packages with build steps: `build` to build the package for development or distribution, `build:all` to build **all** artifacts.
- [x] If this PR introduces a **new package**: n/a — no new packages.


MichaelKapustey (approved):
I would prefer reviewing PR like this as a stack:
1. package.json updates
2. prettier changes
3. Actual changes like addition of a file "assertPresent.ts" or hardening the type,etc

jmuzina (approved):

advl:
Thanks @MichaelKapustey @jmuzina for the reviews. 


A quick note on : 

> I would prefer reviewing PR like this as a stack:

>    package.json updates
>   prettier changes
>   Actual changes like addition of a file "assertPresent.ts" or hardening the type,etc

Well noted ! Will keep the drive-bys separate as we go. However, the formatting (I imagine that's what you mean with prettier) changes and the package.json updates should ideally ship together since the CI wouldn't pass with only one of the two.

jmuzina:
> However, the formatting (I imagine that's what you mean with prettier) changes and the package.json updates should ideally ship together since the CI wouldn't pass with only one of the two.

Github now supports [stacked PRs](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests), so you could chain multiple PRs and keep semantic links between them:

1. package.json updates / biome schema updates (will fail CI at first)
2. Application of formatting changes from the previous step
3. Other changes

Then you could merge them as a stack or merge them individually up into the first one before merging them as one PR. The benefit here is it compartmentalizes review load

advl:
Thanks for the hint @jmuzina  ! Will do :)

PR #1039, opened 2026-08-27
feat(ds-global): add Modal pattern

stavropouloskonstantinoss:
Adds the Modal pattern to react-ds-global.

## Done

New Modal pattern in react-ds-global. ( [documentation](https://docs.superhuman.com/d/Pragma-Documentation-Database_dNyzE_TLZDh/per-Batches_suJk3KZ6?loginId=ZGlhbmEuc3RhbmNpdWxlc2N1QGNhbm9uaWNhbC5jb20#UI-Blocks-per-Batches_tubiUNey/r60&view=center)

## QA

QA: concrete steps a reviewer follows — e.g. run Storybook for react-ds-global, open the Modal story, verify it opens/closes via the close button, Escape, and backdrop click, and that it renders correctly on light and dark surfaces.

### PR readiness check

- [x] PR should have one of the following labels:
  - `Feature 🎁`, `Breaking Change 💣`, `Bug 🐛`, `Documentation 📝`, `Maintenance 🔨`.
- [x] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [x] The code follows the appropriate [code standards](https://github.com/canonical/web-code-standards)
- [x] All packages define the required scripts in `package.json`:
- [x] No new package

## Screenshots

image

PR #1103, opened 2026-09-03
refactor(ds-global): control the Modal pattern through a ref

potentialwook:
Stacked on #1039 — **base is `feat/modal-pattern`**, not `main`.

## Done

Reshapes the `Modal` pattern from a controlled component into a self-contained one, with optional external control through a ref.

- **`open` and `onOpenChange` are gone.** The open state lives in the native `<dialog>`, so the header's close button, Escape and (where `closeOnBackdropClick` opts in) a backdrop click all close the modal with nothing wired up. This removes the desync trap the old contract had to warn about: a consumer who did pre-close work but forgot to set `open` to `false` left the prop and the DOM disagreeing, and the modal could never be reopened.
- **`ref?: Ref<HTMLDialogElement>`** — the escape hatch: `ref.current?.showModal()` opens, `ref.current?.close()` closes. Merged into the component's own ref via a callback ref, so the internal reference stays stable, callback refs work, and a React 19 ref cleanup function is honoured.
- **`defaultOpen?: boolean`** — uncontrolled, acted on once by a mount effect with a spent latch, so no later render can reopen a modal the user has closed. Renders closed on the server, because `showModal()` is what makes a dialog modal.
- **`onClose` is no longer omitted** from the native attributes: the native `close` event is how a consumer hears about a dismissal now, and `close(value)` + `returnValue` is how an action tells itself apart from one.
- **No `onCancel`/`onClose` handlers of our own.** Escape's `cancel` event closes the dialog as its default action — the platform's close watchers do it better than a handler could, and a consumer that must intervene still gets the event.
- **`withModal`** drops `useState` and drives the dialog through a ref. HOC-owned props (`ref`, `defaultOpen`) are now applied after the `modalProps` spread, because `Omit` is a type-level barrier only and a stray `ref` would silently unwire the trigger.

Drive-bys, both found while reviewing the above and both in the lines this PR already rewrites:

- **a11y:** `aria-labelledby={titleId}` was set unconditionally. `aria-labelledby` beats `aria-label` in the accessible-name computation, so a header-less modal's own `aria-label` was being silenced — the exact case the old code comment claimed was handled. A consumer-supplied name now wins.
- **bug:** `{...props}` spread after `onClick` meant any consumer `onClick` silently replaced the backdrop handler and disabled `closeOnBackdropClick`. The two are now composed, consumer first.

Storybook:

- Every story pairs its modal with a trigger button and wires the footer actions to `close`, so dismissing a preview no longer leaves a dead frame with no way back — `NotDismissible` was the real trap, having neither a close icon nor backdrop dismissal.
- Snippets show the wiring a consumer actually writes (ref, trigger, `onClick={close}`) instead of the old `open`/`onOpenChange` pair.

## QA

1. `bun install && cd packages/react/ds-global && bun run check && bun run test`
2. `bun run storybook` → **patterns/Modal**. For each story: dismiss the modal (X, Escape, or backdrop where the story opts in), confirm the trigger button is revealed, and reopen it. Footer `Cancel`/`Confirm`/`Discard`/`Accept` should all close.
3. **patterns/Modal/withModal** → unchanged behaviour: trigger opens, X/Escape close, `FooterAction`'s "Got it" closes.
4. Reopen a modal several times in a row — the state that used to desync now lives in the element, so this must never wedge.

## Notes for the reviewer

- 51 tests over the pattern (was 40). Each new guard was mutation-tested: the fix was reverted one at a time and a named test confirmed to fail. That caught a weak test of my own, since split in two.
- The jsdom `close` stub was made faithful to the platform — it only fires `close` when the dialog was open, and it records `returnValue`.
- **Left for a separate PR:** a header-less modal with no `aria-label` still emits `aria-labelledby` pointing at an unused id, so it has no accessible name (`aria-valid-attr-value`). The fix is the Header registering its title id through context so the Modal only points at an id that exists — a Header/context change, unrelated to ref control.
- No unmount cleanup calling `close()`: the HTML spec's dialog removing steps already drop the element from the top layer, and an unmount is not a dismissal.
- **CI on this branch will show failures inherited from the base.** The base is 33 commits behind `main`, which bumped biome and its shared config: `ds-assets`, `storybook-addon-shell-theme`, `task`, `storybook-helpers`, `ke-graphql` and one Accordion story all fail `check` on files this PR does not touch (`main` has already fixed the Accordion one). It clears when #1039 rebases onto `main`; doing it here would rewrite #1039's commits.

### PR readiness check

- [x] PR should have one of the following labels: `refactor`.
- [x] All packages define the required scripts in `package.json` — no package added or changed.
- [x] No new package, so no manual first publish or trusted-publisher setup.
- [ ] If this PR does not require visual testing, add the `no visual change` label to skip Chromatic. — **it does**: every Modal story gains a trigger button.

## Screenshots

Chromatic will diff the Modal stories: each preview now renders an "Open modal" button behind the backdrop.

### DIFF
--- a/packages/react/ds-global/src/lib/pattern/Modal/Modal.ssr.tests.tsx
+++ b/packages/react/ds-global/src/lib/pattern/Modal/Modal.ssr.tests.tsx
@@ -6,7 +6,7 @@ describe("Modal SSR", () => {
   it("doesn't throw", () => {
     expect(() =>
       renderToString(
-        <Component open>
+        <Component defaultOpen>
           <Component.Content>Placeholder content</Component.Content>
         </Component>,
       ),
@@ -15,7 +15,7 @@ describe("Modal SSR", () => {
 
   it("renders the dialog and its composed parts", () => {
     const html = renderToString(
-      <Component open>
+      <Component defaultOpen>
         <Component.Header>Title</Component.Header>
         <Component.Content>Placeholder content</Component.Content>
         <Component.Footer>
@@ -27,5 +27,17 @@ describe("Modal SSR", () => {
     expect(html).toContain('class="ds modal-header"');
     expect(html).toContain('class="ds modal-content"');
     expect(html).toContain('class="ds modal-footer"');
+  });
+
+  it("renders closed even with defaultOpen", () => {
+    // showModal() is what makes a dialog modal, and it only runs in the mount
+    // effect: a server-rendered `open` attribute would paint a non-modal
+    // dialog with no backdrop and no focus trap.
+    const html = renderToString(
+      <Component defaultOpen>
+        <Component.Content>Placeholder content</Component.Content>
+      </Component>,
+    );
+    expect(html).not.toMatch(/<dialog[^>]*\sopen/);
   });
 });
--- a/packages/react/ds-global/src/lib/pattern/Modal/Modal.stories.tsx
+++ b/packages/react/ds-global/src/lib/pattern/Modal/Modal.stories.tsx
@@ -1,9 +1,40 @@
 import type { Meta, StoryObj } from "@storybook/react-vite";
+import type { ReactElement, ReactNode } from "react";
+import { useRef } from "react";
 import { Button } from "../../component/Button/index.js";
 import { Chip } from "../../component/Chip/index.js";
 import { InlineCode } from "../../component/InlineCode/index.js";
 import { KeyboardKey } from "../../component/KeyboardKey/index.js";
 import Modal from "./Modal.js";
+import type { ModalProps } from "./types.js";
+
+/**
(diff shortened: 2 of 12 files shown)


stavropouloskonstantinoss:
Thanks for putting this together @potentialwook , and for exploring ref-based control as an alternative to React state. It's a valid approach and I don't have a strong argument for one implementation over the other.

My one hesitation is about what the API asks of the consumer. With the ref, they have to know that the modal is built on a native <dialog> and drive it imperatively (showModal() / close()), so an implementation detail becomes part of the public contract. React is a declarative library: consumers generally expect to describe what the UI should look like for a given state and let React work out the how, rather than issuing step-by-step instructions to the DOM. @advl What are your thoughts?

PR #1119, opened 2026-09-04
feat(Button)!: migrate to design tokens. Adopt new density across the design system

MichaelKapustey:
## Done

Migrated Button from the legacy `severity` API to the DS25 modifier families:

- `importance` (`primary | secondary | tertiary`, default `secondary` = the legacy default outlined look), `anticipation` (`constructive | destructive | caution`), `emphasis` (narrowed to `branded`).
- `information` was dropped because it is not used anywhere
-  `emphasis=branded` is a launchpad extra: the global Button's modifier families are anticipation + importance only. 
- CloseButton: `severity="base"` becomes `importance="tertiary"`. Modal/SidePanel stories moved to the new props.

Adopted the shared density system:

- Density family renamed to the global vocabulary: `comfortable | dense | compact` 
- Button consumes the `@canonical/styles` density channels (`--density-line-height-effective`, `--density-padding-inline`) via plain var().
- Heights: comfortable/default 32px, dense 26px, compact 22px — dense and compact are pixel-identical to the shipped Chromatic baseline; the default normalizes 34→32 (and inline padding 12→16) onto the design-system size, folded into this migration's visual review.
- `compact` is a launchpad-only extra 
- 

Stories reworked

! Breaking: the `severity` prop is removed and density values are renamed.

## QA
Visual review

### PR readiness check

  - [x] Packages with build steps: `build` to build the package for development or distribution, `build:all` to build **all** artifacts.
- [ ] If this PR introduces a **new package**: not applicable — no new package.
- [ ] Chromatic must run — no skip label; this is a visual migration.

### Color modifiers
Information is not used anywhere, therefore dropped
Now
image
Was
image

Now
I'm to lazy to screenshot a light theme without information. Please imagine its not there
image
Was
image

### Densities

Now
image
Was
image

### DIFF
--- a/packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
+++ b/packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
@@ -2,15 +2,31 @@
   import { ArchiveIcon } from "@canonical/svelte-icons";
   import { defineMeta } from "@storybook/addon-svelte-csf";
   import { fn } from "storybook/test";
+  import type { ModifierFamilyValues } from "../../modifier-families/index.js";
   import { MODIFIER_FAMILIES } from "../../modifier-families/index.js";
   import Button from "./Button.svelte";
-  import type { ButtonProps } from "./types.js";
 
-  const BUTTON_SEVERITIES = [
-    "brand",
-    "base",
-    ...MODIFIER_FAMILIES.severity,
-  ] as const satisfies readonly NonNullable<ButtonProps["severity"]>[];
+  type ButtonVariant = {
+    importance?: ModifierFamilyValues["importance"];
+    anticipation?: ModifierFamilyValues["anticipation"];
+    emphasis?: Extract<ModifierFamilyValues["emphasis"], "branded">;
+    criticality?: Extract<ModifierFamilyValues["criticality"], "information">;
+  };
+
+  type MatrixColumn = {
+    label: string;
+    variant: ButtonVariant;
+  };
+
+  const MATRIX_COLUMNS: MatrixColumn[] = [
+    { label: "default", variant: {} },
+    ...MODIFIER_FAMILIES.anticipation.map((anticipation) => ({
+      label: anticipation,
+      variant: { anticipation },
+    })),
+    { label: "information", variant: { criticality: "information" } },
+    { label: "branded", variant: { emphasis: "branded" } },
+  ];
 
   const { Story } = defineMeta({
     title: "Components/Button",
@@ -27,6 +43,21 @@
   };
 </script>
 
+{#snippet variantMatrix(props: { disabled?: boolean; loading?: boolean })}
+  <div class="matrix">
+    <span></span>
+    {#each MATRIX_COLUMNS as column (column.label)}
+      <span class="matrix-label">{column.label}</span>
+    {/each}
+    {#each MODIFIER_FAMILIES.importance as importance (importance)}
+      <span class="matrix-label">{importance}</span>
+      {#each MATRIX_COLUMNS as column (column.label)}
+        <Button {importance} {...column.variant} {...props}>Button</Button>
+      {/each}
+    {/each}
+  </div>
+{/snippet}
+
 <Story
   name="Default"
   args={{
@@ -38,19 +69,51 @@
   {/snippet}
 </Story>
 
-<Story name="Severities">
+<Story name="Importance">
   {#snippet template(args)}
     <div class="row">
-      {#each BUTTON_SEVERITIES as severity (severity)}
-        <Button {...args} {severity} onclick={fn()}>
-          {severity}
+      {#each MODIFIER_FAMILIES.importance as importance (importance)}
+        <Button {...args} {importance} onclick={fn()}>
+          {importance}
         </Button>
       {/each}
     </div>
   {/snippet}
 </Story>
 
-<Story name="Densities">
+<Story name="Anticipation">
+  {#snippet template(args)}
+    <div class="row">
+      {#each MODIFIER_FAMILIES.anticipation as anticipation (anticipation)}
+        <Button {...args} {anticipation} onclick={fn()}>
+          {anticipation}
+        </Button>
+      {/each}
+    </div>
+  {/snippet}
+</Story>
+
+<Story name="Criticality">
+  {#snippet template(args)}
+    <Button {...args} criticality="information" onclick={fn()}>
+      information
+    </Button>
+  {/snippet}
+</Story>
+
+<Story name="Emphasis">
+  {#snippet template(args)}
+    <Button {...args} emphasis="branded" onclick={fn()}>branded</Button>
+  {/snippet}
+</Story>
+
+<Story name="Matrix">
+  {#snippet template()}
+    {@render variantMatrix({})}
+  {/snippet}
+</Story>
+
+<Story name="Density">
   {#snippet template(args)}
     <div class="row">
       {#each MODIFIER_FAMILIES.density as density (density)}
@@ -107,7 +170,7 @@
       <br />
       <br />
     </div>
-    <p style="font-size: 12px; color: var(--lp-color-text-muted);">
+    <p style="font-size: 12px; color: var(--color-text-muted);">
       Click the button to toggle the loading state.
     </p>
   {/snippet}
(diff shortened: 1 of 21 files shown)


advl (approved):
I wonder if we could reuse the matrix utility somehow

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -131,13 +205,30 @@
...
-  {#snippet template(args)}
-    <Button {...args} severity="brand" onclick={fn()}>
-      {#snippet iconLeft()}
-        <ArchiveIcon />
-      {/snippet}
-      Brand action
-    </Button>
-  {/snippet}
-</Story>
+<style>
```
steciuk, line 222: I clearly remember in-CSF story styles not being properly scoped and instead becoming global for the entire storybook not so long ago. Checked, and seems working as expected now! Amazing!

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -27,6 +41,25 @@
   };
 </script>
 
+{#snippet variantMatrix(props: { disabled?: boolean; loading?: boolean })}
```
steciuk: 1.
```suggestion
{#snippet variantMatrix(props: ButtonProps)}
```

2. Could we move the snippet definition to just above the styles for it (to the bottom)

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -27,6 +41,25 @@
...
+  <div class="matrix">
```
steciuk: Could we please use a table instead? Seems like that's exactly what we need here.

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -38,19 +71,43 @@
   {/snippet}
 </Story>
 
-<Story name="Severities">
+<Story name="Importance">
+  {#snippet template(args)}
+    <div class="row">
+      {#each MODIFIER_FAMILIES.importance as importance (importance)}
+        <Button {...args} {importance} onclick={fn()}>
```
steciuk, line 62: If we have a story that hardcodes `importance` and overrides whatever was set by the user in the story's controls (`{importance}` after `{...args}`), that particular control should be disabled.

Otherwise it's modifyable but without effect. This applies to other stories below as well.

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -122,6 +179,23 @@
   Disabled button
 </Story>
 
+<Story name="Disabled matrix">
```
steciuk: 1. The disabled state here should come from a storybook arg to reflect and be able to be changed by the story's control.
2. I wonder if the docs page isn't becoming a touch too crowded with all the different preloaded variants/combinations. Maybe we could `tags={["!autodocs"]}` to keep the main page a bit leaner while still keeping the stories for the automatic visual testing?

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -122,6 +179,23 @@
...
+  {#snippet template()}
+    {@render variantMatrix({ disabled: true })}
+  {/snippet}
+</Story>
+
+<Story name="Loading matrix">
```
steciuk: Same as above. The `loading` could be set as an `arg`.

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -38,19 +71,43 @@
...
+        <Button {...args} {anticipation} onclick={fn()}>
+          {anticipation}
         </Button>
       {/each}
     </div>
   {/snippet}
 </Story>
 
-<Story name="Densities">
+<Story name="Emphasis">
```
steciuk: Define `emphasis` as Story's `arg` please.

packages/svelte/ds-app-launchpad/src/lib/components/Button/styles.css
```diff
@@ -1,35 +1,42 @@
...
+    calc(
+      (
+        var(--dimension-block-size-button) -
+        1lh -
+        2 *
+        var(--dimension-border-width-button, 0px)
+      ) /
+      2
+    )
   );
```
steciuk, line 36: Would you mind explaining (and maybe leaving a comment) why these values are calculated the way they are?

steciuk (changes requested):

packages/svelte/ds-app-launchpad/src/lib/components/Button/styles.css
MichaelKapustey: The fallback calculation comes from here:
https://github.com/canonical/pragma/blob/47d4182a8182f6766925c94a303032bd953c946c/packages/styles/main/src/modifiers.density.css#L53-L57

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
```diff
@@ -205,30 +177,58 @@
   Link button
 </Story>
 
+{#snippet variantMatrix(props: ButtonProps)}
+  <table class="matrix" aria-label="Button modifier combinations">
+    <colgroup span="2"></colgroup>
+    <colgroup span={MODIFIER_FAMILIES.anticipation.length}></colgroup>
+    <colgroup></colgroup>
```
steciuk, line 184: Nit:  These `colgroup`s don't have any styling or width control applied to them. Hence, I don't think they serve any purposue here. Correct me if I'm wrong. Otherwise:

```suggestion
```

packages/svelte/ds-app-launchpad/src/lib/components/Button/styles.css
steciuk: So if I read [this](https://github.com/canonical/pragma/blob/main/packages/styles/main/src/modifiers.density.css) right, it should rather be:

```css
  --dimension-padding-block-button: max(
    0px,
    calc(
      (
        var(--density-line-height-effective) -
        1lh -
        2 *
        var(--dimension-border-width-button, 0px)
      ) /
      2
    )
);
```

And our root should gain ` .app` class, which guarantees `--density-line-height-effective` to contain the correct fallback chain?

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
MichaelKapustey: They have semantical meaning. Columns  in a group do belong to the same group

packages/svelte/ds-app-launchpad/src/lib/components/Button/styles.css
MichaelKapustey: I considered requirement for a class a big change, that is why there is
```
  --dimension-block-size-button: var(
    --density-line-height-effective,
    calc(var(--baseline-height, 0.25rem) * 8)  <--- the fallback
  );
```

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
steciuk: Just a `th` with defined `span` and a `scope="colgroup"` seems to be exposed to AT and announced correctly, and I cannot find any resources that would outline the need for additional matching `clogroup`. Have you?

steciuk (approved):

packages/svelte/ds-app-launchpad/src/lib/components/Button/Button.stories.svelte
MichaelKapustey: W3C’s irregular-table guidance explicitly says that the groups must first be defined in the markup using `<colgroup>` before making these header associations: https://www.w3.org/WAI/tutorials/tables/irregular/ .

PR #1128, opened 2026-09-05
fix(react-ds-app)!: the app-tier Button sheets read tokens that exist

advl:
## Done

Four packages — `@canonical/react-ds-app-lxd`, `-anbox`, `-landscape` and
`-portal` — ship one Button stylesheet each, and the four files are byte for
byte identical. They are a carry-forward of the shared `ds-react-core` Button
stylesheet of 2025, copied into the four tiers when they were split out (#409,
2026-02-06) and mechanically renamed `--intent-*` to `--modifier-*` — verified
by diff against `packages/ds-react-core/src/ui/Button/styles.css` at
`cb5636915`, where that rename is the *only* difference. They were never a copy
of the global tier's Button, which already declared `display: inline-flex` and
`text-align: center` at the commit that created these files, and never an
application-tier design decision. The global tier moved on; the carry-forward
did not.

Five of the copy's declarations read custom properties that
`@canonical/react-ds-global` has since renamed or dropped
(`ds-app-lxd/src/lib/Button/styles.css:32,39-42` before this change, and its
three twins):

```css
margin-block-end:  var(--button-margin-bottom);
margin-inline-end: var(--button-margin-left);
padding-block:     var(--button-padding-vertical);
padding-inline:    var(--button-padding-horizontal);
background-color:  var(--modifier-color, var(--button-color-background));
```

Nothing these packages load declares any of those five names. Two were renamed
— the paddings, to `--button-padding-inline-start` and `-end` — and three were
dropped: the global sheet declares no margin on the button at all, and
`--modifier-color` went with the anticipation family's rework, which kept
`--modifier-color-text`, `-border`, `-icon` and three `-foreground-*` names
(`-ghost`, `-primary`, `-secondary`). The surface channel the importance
modifiers write is `--modifier-surface`, a different one.
(The only place in the repository that still declares the four `--button-*`
names is the separate Lit prototype package, on `:host`
— `packages/lit/ds-prototype/src/lib/Button/styles.css:12-15`; no React
application tier depends on it.)

A declaration whose `var()` cannot be resolved is invalid at computed-value
time, which means the initial value, or the fallback where one is written. Two
things have therefore been quietly wrong. The horizontal padding of an
application tier's button is gone wherever the app sheet is emitted second — the
two sheets write `.ds.button` at the same specificity, so bundle order decides,
which is defect D10 itself and the reason both tiers are being layered. #1125
measured the global tier winning in the other order. And every
button carrying a modifier has been painted with the neutral *primary* fill
rather than the surface its modifiers ask for. `--modifier-color` was not
renamed to `--modifier-surface` — it was dropped outright; `--modifier-surface`
is a different, importance channel declared by
`packages/styles/main/src/modifiers.importance.shim.css:46`. The
dead name therefore fell through to `--button-color-background`, which is
declared on `:root`, so *that* fallback's own `var()` substituted at `:root`,
where no modifier class exists — the neutral fill, every time, beating the
global sheet's `--modifier-surface`, which the shim declares on the button
itself where the classes are in scope. The block padding survived only
because the global tier wins it back with a more specific selector, the density
seat `:is(.comfortable, .dense, .app, .site, .docs) .ds.button` at (0,3,0).

That accident ends with the layering. #1123 put the global tier in
`ds.components.global` and #1125 put each of these tiers in a layer named
after the tier it implements — `ds.components.apps-lxd`, `-anbox`,
`-landscape`, `-portal` (VC.31, owner ruling 2026-09-06; the single shared
`ds.components.app` is retired). A
layer beats a selector: the app tier now wins every property it names however
specific the global tier's rule, so all four dead declarations take effect and
the button computes `padding: 0` on every side, 6px shorter than the density
grid it is meant to sit on. #1125 measured this and left it here deliberately,
because its own contract was that no declaration is added or removed.

**Each sheet is now the delta it was always meant to be** — only the properties
on which the application tier disagrees with the global one, reading only tokens
that exist. Which those are was measured, not guessed: every Button story
rendered under the global sheet alone and under the global sheet plus the app
sheet, both unlayered as on `main`, with every longhand and every custom
property of every element compared in Chromium 151, at rest and with `:hover`
and `:active` forced through the debugger — then every importance, anticipation,
variant and state the global sheet has a rule for swept across three density and
theme scopes, because the stories only reach the resting default. The whole
sheet is now:

```css
/* ds-app-lxd; the other three are identical but for the layer name */
@layer ds.components.apps-lxd {
  .ds.button {
    font-size: var(--button-font-size);
    --modifier-outline: 1;
  }
}
```

plus the comment that says why each line is there and why nothing else is.

| Declaration | Global tier | App tier | Kept? |
|---|---|---|---|
| `display` | `inline-flex` | `inline-block` | **no** — a 2025 carry-forward with no design source at any application tier, and on an inline-block box the global sheet's `gap` and its `:has(> .icon)` start-padding tighten compute but do not apply, so a leading-icon button silently lost its 8px icon/label gap: 50.27px wide on `main`, against 89.16px for the same markup under the global tier alone. Dropped |
| `text-align` | `center` | `left` | **no** — same provenance, no design source. Dropped |
| `font-size` | *(not declared; 16px inherited)* | 14px | **yes** — real, and kept pending an owner's confirmation on PRA-153 |
| the border | painted only when an importance raises `--modifier-outline` | always painted | **yes** — real, now written as `--modifier-outline: 1` rather than a second `border-color`; kept pending the same confirmation on PRA-153 |
| `cursor`, `color`, `border-style`, `border-width`, `font-weight`, `text-decoration` | — | — | no — measured identical with the app sheet present and absent; pure restatement |
| `background-color` | `var(--modifier-surface, …)` | `var(--modifier-color, …)` — dead name | no — the delta is none for a bare button, which is all the stories carry. For every button that carries a modifier it forced the neutral primary fill: over the ghost surface on `secondary`/`tertiary`, and over the anticipation colour on `primary constructive`/`destructive`/`caution`. **The global value stands, and this is the second and third visible change** |
| `line-height` | density seat, (0,3,0) | `var(--button-line-height)`, (0,2,0) | no — never applied; under layers it would have applied and unseated the button |
| `margin-block-end`, `margin-inline-end` | not declared → `0` | dead token → `0` | no — the delta is none |
| `padding-block` | density seat | dead token → `0` | no — the delta is none; the seat's value stands |
| `padding-inline` | `16px` each side | dead token → `0` | no — the delta is none; **the global value stands, and this is the visible change** |
| `&:hover`, `&:active` | `:hover:not(:disabled)`, (0,4,0) | `:hover`, (0,3,0) | no — outranked on an enabled button; their only measured effect was recolouring a *disabled* button under the pointer |

The `font-size` line deserves its own note, because it looks like a restatement
and is not. The global sheet computes `--button-font-size` — the 14px
text-secondary tier — and publishes it as `--font-size` for the typography engine
to apply, but the engine only reads that off elements carrying `.p`, and the
button carries none. So the global sheet's own label size never reaches the
label: measured, a global-tier button's text computes at the inherited 16px. The
app tier applies it. If F-6's global-tier lane makes the global sheet apply its
own label size, this line becomes a restatement and should go with the rest.

The four READMEs now carry the two facts a consumer of these packages needs and
could previously only find in this description: under Prerequisites, that the
stylesheet is a delta and a button rendered without the global tier's Button
sheet falls back to the browser's own styling; under Storybook, that these
previews load `@canonical/styles` alone, so they show a near-native `<button>`
until the global tier's sheet can be imported there.

Nothing is installed, nothing is generated, and there is no `!important` in any
of the four files.

## QA

Run in each of the four package directories.

| Package | `check` | `build` | `test` | `build:storybook` |
|---|---|---|---|---|
| `ds-app-lxd` | pass | pass | 12 pass | pass |
| `ds-app-anbox` | pass | pass | 12 pass | pass |
| `ds-app-landscape` | pass | pass | 12 pass | pass |
| `ds-app-portal` | pass | pass | 12 pass | pass |

Twelve, not three: this branch has merged `refactor/components-layer-app`
forward three times, and its review rounds added `src/lib/styles.layer.tests.ts`
to each package and then grew it — for the tail assertion, and again for the
per-tier layer name and the `src/lib/index.css` entry. That guard passes on the
rewritten sheet.

**The layers, read from the CSSOM.** Each package's sheet parsed in Chromium
151: one top-level rule, one layer name — its own package's
`ds.components.apps-<name>` — and zero rules outside a layer. Each package's
`src/lib/index.css`, which #1125 added, declares that layer before importing
this sheet, so the layer takes its place in the order whichever sheet a bundler
emits first.

**Computed styles.** A page carrying `class="ds app comfortable light"` on
`<html>`, `@canonical/styles` from this stack, the global tier's stylesheets, and
the package's own sheet. Every element of every Button story, every longhand and
every custom property, animations and transitions wound to a fixed point, in
three states. Two baselines are compared, and both give the same answer for all
four packages:

| Comparison | Differing elements |
|---|---|
| `main` (both unlayered, global first) → this branch (both layered, global then app) | 7 buttons, 3 properties |
| `main` → this branch (both layered, **app then global**) | the same 7, the same 3 |
| this branch, global then app → this branch, app then global | **0** — the order no longer matters |

The properties that move, and why each does:

1. **`padding-inline: 0 → 16px`** on every button, so each is 32px wider. This
   is the fix: the dead token no longer erases the global tier's inline padding.
   `padding-block` stays at the density seat's `0.5px / 5.5px` and the height
   stays 32px — the seat governs the box again instead of being outranked from a
   higher layer. That holds on a root carrying a density or context class, which
   the root contract requires and the Storybook addon supplies
   (`packages/storybook/addon-utils/src/withUtilStyles.ts:213-219`); on a bare
   `ds` root the seat's `:is(.comfortable, .dense, .app, .site, .docs)` never
   matches, and with the app sheet's `line-height` gone the line box falls from
   24px to 16.1px and the button from 26px to 20.09px.
2. **`border-color` serialisation**: `oklch(0 0 0)` → `color(srgb 0 0 0)`, and
   `rgb(25, 25, 112)` → `color(srgb 0.0980392 0.0980392 0.439216)`. Same colour,
   different notation, because the border now comes from the global sheet's
   `color-mix(in srgb, … 100%, transparent)` rather than a flat declaration.
   Checked by screenshot rather than by arithmetic: a flat border and the
   `color-mix` form paint pixel-identical borders for `midnightblue`, for the
   two neutral ramp tokens the stories use, for a `light-dark()` token and for an
   out-of-gamut `oklch(0.6 0.25 25)`.
3. **`background-color` on the *disabled* button under `:hover` and `:active`**:
   for the plain disabled case, `oklch(0.19 0 0)` / `oklch(0.22 0 0)` →
   `oklch(0.69 0 0)`, the global tier's disabled surface (a disabled button
   carrying an importance or anticipation lands on that modifier's disabled
   surface instead). The app sheet's state rules had no `:not(:disabled)` guard,
   so a disabled button recoloured under the pointer. It no longer does.
4. **`display` and `text-align`** return to the global tier's `inline-flex` and
   `center`, with `min-width`/`min-inline-size` on the label and icon spans
   moving `0px → auto` as flex items. That restores the icon/label gap, measured
   on the same leading-icon markup in `app comfortable light`:

   | | width | icon/label gap |
   |---|---|---|
   | `main` (the old sheet) | 50.27px | 0px |
   | this branch | 82.27px | 8px |
   | the global tier alone | 89.16px | 8px |

   `display`, `gap`, `padding-inline` and height now equal the global tier's
   exactly; the remaining 6.89px of width is the 14px label this tier keeps.

**The sweep beyond the stories.** The stories only exercise a bare resting
button, so the sweep crosses **importance × anticipation × disabled** — 16
combinations — and adds `link`, `link+disabled`, `loading`, a leading icon,
`secondary`+icon and inline custom tokens, 38 buttons in all, across
`app comfortable light`, `app comfortable dark`, `app dense light`,
`site comfortable light` and a bare `ds light` root, in four states (rest,
`:hover`, `:active`, `:focus`+`:focus-visible` forced through CDP). 900 element
comparisons. Every mover falls in one of the groups above; the surface changes
in full:

| button | `background-color`, main | on this branch |
|---|---|---|
| `secondary`, `tertiary` (any anticipation) | `oklch(0 0 0)` | `oklch(1 0 0)` — the ghost surface |
| `primary constructive` | `oklch(0 0 0)` | `oklch(0.551 0.1575 144.01)` |
| `primary destructive` | `oklch(0 0 0)` | `oklch(0.5488 0.2107 23.83)` |
| `primary caution` | `oklch(0 0 0)` | `oklch(0.5492 0.1199 66.86)` |
| `secondary caution`, `tertiary caution` | `oklch(0 0 0)` | `oklch(1 0 0)` |
| in `dark`: `secondary`/`tertiary`, plain or `constructive`/`destructive` | `oklch(1 0 0)` | `oklch(0.2308 0 0)` |
| in `dark`: `secondary caution`, `tertiary caution` | `oklch(1 0 0)` | `oklch(0.1221 0 0)` — the caution ghost, not the neutral one |
| in `dark`: `primary` + `constructive` / `destructive` / `caution` | `oklch(1 0 0)` | `oklch(0.6405 0.1743 144.01)` / `oklch(0.6418 0.2184 24.11)` / `oklch(0.6406 0.1384 67.1)` |

An earlier revision of this PR listed importance and anticipation as separate
variants and never crossed them, so it reported only the `secondary`/`tertiary`
case and called it the only one. The `primary` + anticipation row is the fourth
rendering change.

**Which `BREAKING CHANGE` footer applies.** The branch's third commit,
`fix(react-ds-app): the verification pass`, carries the footer that matches the
shipped sheet: four rendering changes, the density/context qualifier on the
unchanged height and label, and today's zero padding stated as
load-order-dependent. It **supersedes** the footer on the first commit, which
described three changes and stated the padding as a fact.

The importance modifiers were checked separately, since `--modifier-outline` is
the one new declaration. A bare button and one carrying `primary`, `secondary`
or `tertiary` all keep the opaque border they have on `main`; `link` keeps the
global tier's transparent border and zero inline padding, and its height is
unchanged at 30px.

**Two things a reviewer should know, neither of them introduced here.**

- These packages never load the global tier's stylesheets. `@canonical/react-ds-global`
  is a dependency, but no module in any of the four imports it, and each
  Storybook's `.storybook/styles.css` imports `@canonical/styles` and nothing
  else. So in the packages' own Storybooks the delta has no base to modify, and
  the preview shows a native `<button>` where before it showed a stale copy of
  the February component with its tokens missing — transparent background, a 3px
  `medium` border, no padding, 25px tall. Neither preview was the component. The
  fix is for the app tier's Button to load the global tier's Button stylesheet,
  which cannot be written on this branch: the global sheet is only in
  `ds.components.global` once #1123 merges, and importing today's unlayered copy
  would outrank everything. Worth its own change once #1123 lands.
- **Two follow-ons are recorded on PRA-153, F-6's carrier, rather than done here.** A test
  that every `var(--x)` this sheet reads is declared by a stylesheet the package
  loads with — cheap, no new dependency — but it must resolve
  `@canonical/react-ds-global`'s defining `:root`, which needs a built-artefact
  path until #1122 adds the `./index.css` subpath, and its union must include
  the token packages or it false-positives. And the app-tier Storybook import of
  the global Button stylesheet, which needs #1123 for the same reason the
  README now gives.
- The `appearance` prop's values are dead classes. `neutral`, `positive` and
  `negative` set no `--modifier-*` custom property in the current token set — the
  anticipation modifiers are `constructive`, `destructive` and `caution` — so the
  three intent stories render identically to the default one. That is a component
  API question, not a stylesheet one, and it reads the same before and after this
  change.

## Screenshots

None. The visible changes are 32px of horizontal padding per button and the
ghost surface on `secondary`/`tertiary`, both stated above with their
measurement; Chromatic covers the four packages.

Stacked on #1125 (`refactor/components-layer-app`), which is stacked on #1120,
and merged forward from it three times — after its review round (the one-line
layer header, the `## Styling` README section), after it answered its review
threads, and after VC.31 renamed the tier layers (`2b1f60e17`). Those are
#1125's; this branch takes them, and `git diff --name-only
origin/refactor/components-layer-app...HEAD` lists only this PR's eight files.
Its sibling #1123 is what puts the global tier in the layer this delta sits
above.

https://claude.ai/code/session_01MS2nYHcgSsxHVNGsB4gYBp


advl:
This PR should not have been created. its contents are not useful as it targets shims.

PR #1160, opened 2026-09-07
refactor(ds-app)!: render SideNavigation as a self-contained nav landmark

potentialwook:
## Done

**BREAKING CHANGE:** `SideNavigation`'s root element changes from `<div>` to `<nav>`, with `aria-label` defaulting to `"Main navigation"` (overridable). Consumers querying by role or asserting on the root tag (e.g. `getByRole("generic")`) need to update to `getByRole("navigation")`. No other package in this repo consumes `SideNavigation` yet, so the blast radius is contained to this PR.

Per `SPEC.md` §1 and §6 (AC1: "The Side Navigation must have a self-contained `<nav>` tag") — PR 2 of the sequence tracked in `SPEC.md` §8:

- `SideNavigation.tsx` now renders `<nav>`, not `<div>`.
- Every `SideNavigation` prop-types file converts from `interface extends HTMLAttributes<HTMLDivElement>`/`ButtonHTMLAttributes<HTMLButtonElement>` to the house style — a `type` alias intersecting an `OwnProps` object with `Omit<ComponentProps<"tag">, keyof OwnProps>`, per this repo's React component props convention. This was more than a mechanical rename: none of these types previously extended their root element's native props at all (`HeaderProps`/`ContentProps`/`FooterProps` used `HTMLDivElement` while the components never spread the rest onto a typed native surface; `ItemProps` and `NavTreeProps` had no native props at all), so native attributes like `id`, `data-*`, or `aria-*` were previously accepted at runtime (via an untyped `...props` spread) but rejected by the type checker. They are now both accepted and typed.
- `CollapseToggleProps` additionally excludes `type` from its native surface (documented inline) since this is always a disclosure trigger, never a submit/reset control; `CollapseToggle.tsx` moves `type="button"` after the props spread so a consumer cannot override it via a wider `Omit` in the future.
- `NavTree` (private, not part of the public barrel) gains the same `className`/native-prop passthrough as its siblings, for consistency within the component family.
- Adds two tests asserting the nav landmark and the `aria-label` override.

## QA

- `bun run test` — new tests assert `getByRole("navigation", { name: "Main navigation" })` and the `aria-label` override.
- `bun run check` passes (biome + tsc + webarchitect).

### PR readiness check

- [x] PR title follows Conventional Commits, marked `!` for the breaking root-element change.
- [ ] Chromatic: skip — root tag change may affect snapshots; leaving default Chromatic run enabled.

### DIFF
--- a/packages/react/ds-app/src/lib/SideNavigation/SideNavigation.tests.tsx
+++ b/packages/react/ds-app/src/lib/SideNavigation/SideNavigation.tests.tsx
@@ -26,6 +26,20 @@ const footerRoot: NavItem = {
 };
 
 describe("SideNavigation", () => {
+  it("renders as a self-contained nav landmark", () => {
+    render(<SideNavigation root={root} />);
+    expect(
+      screen.getByRole("navigation", { name: "Main navigation" }),
+    ).toBeInTheDocument();
+  });
+
+  it("allows overriding the default aria-label", () => {
+    render(<SideNavigation root={root} aria-label="Product navigation" />);
+    expect(
+      screen.getByRole("navigation", { name: "Product navigation" }),
+    ).toBeInTheDocument();
+  });
+
   it("renders the root items in the content region", () => {
     render(<SideNavigation root={root} />);
     expect(screen.getByText("Machines")).toBeInTheDocument();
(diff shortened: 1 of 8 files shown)


potentialwook:
Superseded by the consolidated stack — see #1170.

PR #1161, opened 2026-09-07
feat(ds-app): style SideNavigation to the 24.04 spec

potentialwook:
## Done

Implements `SPEC.md` §2/§3 (AC2: "The Side Navigation must follow the styling in the spec") — PR 3 of the sequence tracked in `SPEC.md` §8:

- New provisional token file `packages/styles/main/src/navigation.css`: rail width (240px), collapsed rail width and logo width (both derived, not spec-stated — see `SPEC.md` §9.2/§9.3), shared row insets (1rem/1.5rem) and icon↔label gap (0.5rem), and a fixed leading-icon column width (1rem, matching the `Icon` component's own default render size) so labels align down the rail whether or not a row has an icon. Same provisional-then-promote convention as `spacing.css`/`overflow.css`/`motion.css`.
- `SideNavigation` now owns its own rail width and spans the full parent height (previously `max-height`, which let a short item list stop short of the bottom instead of pinning the footer there); overrides the shared `--overflow-gradient-height` to the spec's 1.75rem locally.
- `Header` lines up the logo, collapse toggle, and (via the shared inset tokens) every item's icon on the same 1rem/1.5rem/0.5rem rhythm.
- Item rows get real colour states (background base/hover/active/disabled from `--color-foreground-navigation-primary` and its suffixes; text/icon dim on disabled) — previously deferred/structural-only. Drops the previous baseline-grid-derived fixed row height, whose underlying 8px baseline unit changed to 4px in a prior pass without updating the comment describing it (see `SPEC.md` §9.7).
- `NavTree` groups get the spec's 0.5rem/1rem block padding and a top divider on every group but the first; the group header adopts the heading-5 typography token pairing.
- `Content`'s overflow fade gains a matching top instance (previously only the bottom seam faded into the footer); both fall back to the plain background token when not nested in a `.surface` ancestor, matching the fallback idiom already used elsewhere (`Popover`, `ContextualMenu`).
- `CollapseToggle` gets a real hover/active background and a fixed `"button"` type that a wider prop surface can no longer override.
- Updates the Storybook nav-layout decorator's column width to match the real 240px rail (previously a stale 300px) and refreshes a few stale code comments left over from the removed row-grid technique.

Every inference (row height, icon column width, collapsed rail width, top fade, colour-token naming match) is logged in `SPEC.md` §9 for design consolidation, per the existing token-inference-log convention.

## QA

- Visual: check the Storybook stories for `SideNavigation` and its subcomponents against the Figma 24.04 exploration for rail width, row insets, and divider placement.

### PR readiness check

- [ ] Chromatic: skip — this PR is visual styling; leaving default Chromatic review enabled.


potentialwook:
Superseded by the consolidated stack — see #1171.

PR #1162, opened 2026-09-07
feat(ds-app)!: group, header, separator and expandable navigation items

potentialwook:
## Done

**BREAKING CHANGE:** `SideNavigation`'s content model splits from one flat, recursive `NavItem` into a proper tiered shape — `NavRoot` (`root`/`footerRoot`) > `NavGroup` (a group's own items) > `NavItem` (`LeafNavItem | ExpandableNavItem`). An `ExpandableNavItem`'s own `items` are typed as `LeafNavItem[]` only, so a child of an expandable item cannot itself expand, and it cannot have a `url` — both enforced at the type level per `SPEC.md` §4.3 ("expandable items cannot link", "a child of an expandable item cannot itself have children"). Existing `NavItem`-typed data at root/group position needs to retype as `NavRoot`; existing 3+-level trees (if any) will no longer compile — this is the depth-1 constraint doing its job, not a bug. `Item.tsx`'s `items` prop (and its static disclosure caret) is removed — an entry with children is now `SideNavigation.ItemExpandable` instead.

New public subcomponents, all attached to `SideNavigation` by dot notation (`cs:react.component.subcomponent_export_api`) alongside a new `SideNavigationComponent` type (matching the ds-global `Accordion`/`Card` precedent) — PR 4 of the sequence tracked in `SPEC.md` §8:

- `SideNavigation.Group` / `.GroupHeader` — a labelled collection of entries. Owns the spec's 0.5rem/1rem group padding and the top divider on every group but the first, moved here from `NavTree`'s internal styles now that groups are real, independently-styled components.
- `SideNavigation.Separator` — a plain rule between content sections, a sibling of `Group` among the content's direct entries.
- `SideNavigation.ItemExpandable` — a collapsible row over native `<details>`/`<summary>` (no `role`/`aria-expanded` authored — the element supplies the disclosure semantics itself, per `cs:ui_blocks.nojs.disclosure`). Uncontrolled, seeded by `defaultExpanded`, mirroring `SideNavigation`'s own rail-collapse pattern rather than `Accordion.Item`'s externally-controlled one, since each instance is independent. `NavTree` seeds the initial value from `useNavigationTree`'s `inSelectedBranch` so the ancestor chain of the active page starts expanded (`SPEC.md` §5). The disabled guard prevents the *click* on the summary (the native toggle event isn't cancelable, and skipping the React state update alone doesn't undo a toggle the browser already applied to the DOM).

Also fixes a real, pre-existing gap while wiring this up: no `SideNavigation` subcomponent (`Header`/`Content`/`Footer`/`Item`/`CollapseToggle`) was reachable through the package's public export surface at all — `common/index.ts` existed but was never re-exported from `SideNavigation/index.ts`, and none were attached to `SideNavigation` by dot notation. Fixing this for the whole family in one pass (rather than attaching only the four new subcomponents) avoids shipping a mix of reachable and unreachable subcomponents in the same barrel.

`NavTree` drops its own `.group`/`.header`/`.list` styling (moved into `Group`/`GroupHeader`) and gains a third loop for an `ItemExpandable`'s children, stripping the tree-annotation fields (`_Item<T>`: `parentUrl`, `depth`) before spreading an entry onto a presentational subcomponent — previously spread through unfiltered, a latent leak onto the DOM this change also closes.

Fixes `fixtures.tsx` and the `SideNavigation`/`Content`/`Footer` test roots to retype as `NavRoot` (their actual data shape — `root.items` as groups, `group.items` as entries — was already correct; only the type annotation was wrong). Adds missing test/story coverage for `Group`, `GroupHeader`, `Separator`, and `ItemExpandable` (none existed before this PR), and closes a smaller pre-existing gap: `Item` itself had no unit test file at all.

**Companion fix, folded into this PR rather than split out** (it only becomes necessary the moment `NavGroup`/`NavSeparator`/`NavRoot` exist, so it has no independent reason to exist on its own): `@canonical/ds-types`'s shared `Item` requires a discriminated key-or-url identity (the shared navigation tree indexes every node by one of the two), which `useNavigationTree`'s own `T extends Item` bound enforces structurally. `NavGroup`, `NavSeparator` and `NavRoot` left `key` fully optional even though every real usage already supplies one (`annotateTree`/`getItemId` already require it at runtime) — made required to close that gap. `_AnyNavNode` gains the same identity union, with a single documented cast at the `useNavigationTree` call site bridging it to `LeafNavItem`'s own, deliberately looser shape (key/url both optional, by design — see that type's own doc comment, added in PR1). Also retypes `Content`/`Footer`'s `root` prop from the generic WD405 `Item` to this package's own `NavRoot` — it was already the right shape at the call sites; the looser `Item` type just let it through by accident.

## QA

- New Storybook stories for `Group`, `GroupHeader`, `Separator`, `ItemExpandable` — visually check disclosure open/close and the group-divider placement.

### PR readiness check

- [x] PR title follows Conventional Commits, marked `!` for the content-model breaking change.
- [ ] Chromatic: skip — new subcomponents; leaving default Chromatic review enabled.


potentialwook:
Superseded by the consolidated stack — see #1171.

PR #1163, opened 2026-09-07
feat(ds-app): navigation control variants — ItemButton, ItemSwitch, ContextSwitcher

potentialwook:
## Done

Includes a small companion fix (`fix(react-ds-global-form)`) needed for the feature below, kept in this PR rather than split out since nothing else in the stack depends on it independently.

**Companion fix — `react-ds-global-form` package resolution:** the package declared only the bundler-only `module` field — no `main` and no `exports` map — so nothing outside its own package (and not Node's own ESM resolver) could actually import it; every existing consumer either lived inside the package or imported its CSS by a literal dist path, so this had never been exercised. Adds an `exports` map for the JS entry point and the CSS subpath already imported elsewhere (`apps/react/boilerplate-vite`), following the same pattern already used by `@canonical/utils`. Also curates `SwitchInput`/`SwitchInputProps` onto the public barrel, extending the existing `RatingInput` precedent — needed by this PR's `SideNavigation.ItemSwitch`.

**Feature — navigation control variants.** Implements `SPEC.md` §4.4/§4.5 (AC3: "a polymorphic link component variant including a button, a switch, and a dropdown selection"; AC4: same three listed as the required set) — PR 5 of the sequence tracked in `SPEC.md` §8:

- `SideNavigation.ItemButton` — a real `<button>`, for actions instead of navigation ("Log out", "Create context"). Native `:disabled`/`:hover` styling — no `data-disabled` attribute needed.
- `SideNavigation.ItemSwitch` — a `<label>` wrapping `SwitchInput` (from `@canonical/react-ds-global-form`), so activating anywhere on the row toggles it via native label-wraps-input association.
- `SideNavigation.ContextSwitcher` — a dropdown built on the global `Popover` pattern (native `<details>`/`<summary>`), showing the current context, a user-generated context list with optional descriptions/badges, and an optional "create context" action. A standalone subcomponent for now — not yet a `root.items` entry kind (`SPEC.md` §4.5, §10.12): its `currentContext`/`contexts` state shape doesn't fit the existing `NavGroup`/`NavItem` tiers without adding a fourth heterogeneous shape to `_AnyNavNode` for a widget that isn't tree-structured.

`LeafNavItem` gains a `control?: "link" | "button" | "switch"` field (default `"link"`) plus the fields each variant needs (`onClick`; `checked`/`defaultChecked`/`onCheckedChange`) — one flat shape across all three, not a discriminated union (`SPEC.md` §10.11, a deliberate authoring-simplicity trade-off against the extra type safety a union would buy). `NavTree`'s `renderEntry` helper dispatches on `control`, stripping whichever fields the chosen variant doesn't use — several would otherwise either type-error against a same-named native HTML attribute of a different shape (a bare `onClick` expects a `MouseEvent` handler; `ComponentProps<"li">`'s global `slot` attribute is a string, not this module's `ReactNode`) or leak onto the DOM.

All three are attached to `SideNavigation` by dot notation alongside the existing family (`SideNavigationComponent` type updated to match).

Fixes two latent bugs this PR is the first thing in `ds-app` to exercise: `vitest.setup.ts`'s `ResizeObserver` mock could not be constructed as a constructor (`Popover`'s positioning hook needs a real one).

**Note (a later PR in this stack, `ContextSwitcher`'s rebuild on `ContextualMenu`):** by the time this branch was rebased onto current `main`, `ContextualMenu` itself had moved on to a flat `items`/select-like API. `ContextSwitcher` here still targets the `Popover`-based implementation described above; a subsequent PR in this stack rebuilds it on `ContextualMenu` for real select-like semantics (see that PR's description).

## QA

- New Storybook stories exercise `ItemButton`, `ItemSwitch`, `ContextSwitcher` — verify keyboard toggling and dropdown open/close.

### PR readiness check

- [ ] Chromatic: skip — new components; leaving default Chromatic review enabled.


potentialwook:
Superseded by the consolidated stack — see #1172.

PR #1164, opened 2026-09-07
feat(ds-app): collapsed rail behaviour, tooltips, inert shortcut scaffold

potentialwook:
## Done

Implements `SPEC.md` §1.2/§7 (collapsed state) and §5/§9.4 (tooltips, keyboard shortcut) — PR 6 of the sequence tracked in `SPEC.md` §8:

- The rail narrows to `--sidenav-rail-inline-size-collapsed` (derived in `navigation.css`, PR 3) and `Content` is hidden entirely — not a styling choice, the spec's own deliberate scope cut (five listed reasons: icons aren't guaranteed, context switchers can't render collapsed, expandable state is undefined, the toggle needs to move, icon-only buttons are poor UX).
- `Header` switches to a centred column when collapsed (`data-expanded` on its root, newly wired from the `expanded` prop it already received): the logo centres and the collapse toggle relocates below it purely by changing `flex-direction` — DOM order (brand, then the toggle) already puts the toggle last, so no reordering is needed. The title is dropped entirely; there's no room for it in a rail this narrow.
- The footer stays visible but goes icon-only when collapsed (labels and end slots hidden, icon column centred) — labels return via tooltip in a follow-up (`SPEC.md` §10.14: threading a "collapsed" flag through `Footer` → `NavTree` → `renderEntry` → `Item`/`ItemButton` is cross-cutting, unlike the collapse button's own tooltip below).
- `CollapseToggle` gains a 1-second hover tooltip reading "Collapse" or "Expand" (distinct from its longer `aria-label`). `withTooltip`'s `Message` is fixed at wrap time, not reactive to the wrapped component's props, so two stable wrapped components are built once at module scope (not per render, which would force a remount every toggle) and the exported `CollapseToggle` just picks between them.
- New `common/hooks/useCollapseShortcut`: binds Ctrl+E to the rail toggle, but only when `enabled` — `SideNavigation`'s new `keyboardShortcut` prop defaults to `false`, so no listener ever attaches. Scaffolded per `SPEC.md` §5/§10.1 because the spec's own prose is self-contradictory on the key (states "Ctrl + E", then argues for a single letter instead, in the same paragraph) — ships inert pending a design ruling. Tested for both states, including that the default (disabled) case truly never fires, on the exact key combination.

Drive-by: the Storybook `withSideNavShell` decorator wrapped subcomponent stories in a `<div>`; updated to `<nav>` to match the real component's root since PR 2's `<nav>` landmark change.

## QA

- Manually toggle collapse/expand and confirm the tooltip text, the header's centred layout, and the footer's icon-only state.
- Confirm Ctrl+E does nothing by default, and (in a follow-up consumer opt-in) fires only when `keyboardShortcut` is `true`.

### PR readiness check

- [ ] Chromatic: skip — visual collapsed-state changes; leaving default Chromatic review enabled.

### DIFF
--- a/packages/react/ds-app/src/lib/SideNavigation/SPEC.md
+++ b/packages/react/ds-app/src/lib/SideNavigation/SPEC.md
@@ -372,14 +372,17 @@ entirely and repositions `CollapseToggle` below the logo (§1.2, §10.9).
 
   ```ts
   // common/hooks/useCollapseShortcut/useCollapseShortcut.ts
-  export const COLLAPSE_SHORTCUT = { ctrlOrMeta: true, key: "e" } as const;
+  export const COLLAPSE_SHORTCUT = { key: "e", ctrlKey: true } as const;
   ```
 
   `SideNavigation` calls `useCollapseShortcut({ enabled: keyboardShortcut,
   onTrigger: handleToggle })`; `keyboardShortcut` defaults to `false`, so no
   listener attaches and no keydown is ever handled. The single named constant
   is the one place the key changes if/when design ratifies "Ctrl+E" vs. a
-  single letter (§10.1). Tested for both: enabling attaches and fires the
+  single letter (§10.1) — matching the spec's literal "Ctrl + E" only
+  (`event.ctrlKey`, not also `event.metaKey`/Cmd); extending it to Cmd on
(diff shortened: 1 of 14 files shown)


potentialwook:
Superseded by the consolidated stack — see #1172.

PR #1165, opened 2026-09-07
feat(ds-app): secondary navigation, Help item, footerItems, certificate user

potentialwook:
## Done

Implements `SPEC.md` §1.3/§4.2 (secondary navigation), §1.1/§4.3 (the mandatory Help item), and §4.1 (closed `footerItems` vocabulary, certificate user) — PR 7 of the sequence tracked in `SPEC.md` §8:

- `SideNavigation.Secondary` — a second `<nav>` landmark, a **sibling** of `SideNavigation` rather than a child of it: the spec's "hides on primary click" behaviour is a consumer-level mount/unmount decision, not something `Secondary` controls internally (no `onClose` prop — there is no internal control that would call one). Reuses `SideNavigation.Content` directly for its own content region (same overflow-fade treatment) rather than duplicating it. Its entries are typed `SecondaryNavItem` (`LeafNavItem` without `icon`, no expandable variant at all), so a well-typed consumer cannot author the two shapes the spec says secondary nav must not have.
- New `root.items` architectural fix: `Item`/`ItemButton`/`ItemSwitch`/`ItemExpandable`/`CollapseToggle`/`ContextSwitcher` now consume a `--sidenav-row-background(-hover/-active/-disabled)` channel instead of the navigation-primary token directly, so `Secondary`'s own root can retint every descendant row to the secondary token in one place (an inherited custom-property override) instead of threading an "is this secondary?" flag through every row component.
- `createHelpItem(legalUrl, options)` — a data-construction helper (not a component) that builds the content's one mandatory item: a collapsible "Help" `ExpandableNavItem` containing an external link to legal information. Removes the chance of authoring it slightly wrong every time; wired into the MAAS/LXD story fixtures as a trailing group.
- `footerItems: FooterItem[]` — the spec's closed footer-items vocabulary (account/notifications/logout), each carrying its own `url`/`onClick`/`label`/`slot`. Navigable items (`url` set) render as `Item`; action items (no `url` — nearly always logout) render as `ItemButton`. `certificateUser` swaps the account item's icon to `certificate` and drops any logout item, per the spec's own named example ("If the application allows for a certificate user (like LXD)"). `footerRoot` (the pre-24.04 free-form escape hatch) stays available and takes precedence only when `footerItems` is absent.

Known gaps recorded rather than rushed: `brandHref`/logo-target isn't implemented (`SPEC.md` §10.15 — would require `SideNavigation` to own the brand's anchor, a bigger change than any current acceptance criterion needs); collapsed-footer tooltips remain the PR 6 follow-up (already shipped in PR 6, tracked as a known gap there).

## QA

- New Storybook stories for `Secondary`, the Help item, `footerItems`, and `certificateUser` — visually check secondary nav retinting and the closed footer vocabulary.

### PR readiness check

- [ ] Chromatic: skip — new components; leaving default Chromatic review enabled.


potentialwook:
Superseded by the consolidated stack — see #1173.

PR #1166, opened 2026-09-07
feat(ds-app): responsive layout, text-overflow tooltips, focus-order and reduced-motion tests

potentialwook:
## Done

Implements `SPEC.md` §7 (responsive, text overflow) and closes out §6/§8 (focus order, reduced motion) with dedicated coverage — PR 8 of the sequence tracked in `SPEC.md` §8:

- Below Vanilla's small breakpoint (<768px): `CollapseToggle` swaps its icon for a "Menu"/"Close menu" text label — plain CSS-toggled siblings (an icon and a text span, one hidden per breakpoint via media query), not a JS media-query read, so the choice needs no client-only branch and stays SSR-identical. `SideNavigation`'s own collapsed state means "menu closed" (header only, both `Content` AND `Footer` hidden — unlike the desktop collapsed rail, which keeps the footer visible icon-only) and expanded means a fullscreen fixed overlay rather than a 240px rail. `Header`'s centred-column collapsed layout (built for the desktop icon rail) is suppressed under the same breakpoint, since mobile's header stays in its normal row layout.
- Every label (`Item`/`ItemButton`/`ItemSwitch`/`ItemExpandable`/`GroupHeader`/`Secondary`'s title) now truncates with an ellipsis on overflow and sets a native `title` attribute as a working, if not pixel-accurate, fallback for the spec's custom 800ms-delay tooltip (`SPEC.md` §10.17 — a precise version needs to detect *actual* truncation via a `ResizeObserver`-backed hook that doesn't exist yet in this repo; unconditionally wrapping every label in `withTooltip` would show a tooltip even when nothing is truncated).
- New tests: focus order across the whole component (logo → collapse toggle → content → footer, matching DOM order, `SPEC.md` §6) and a reduced-motion regression guard reading `ItemExpandable`'s own stylesheet source to confirm the caret's rotation transition stays behind `@media (prefers-reduced-motion: no-preference)` — jsdom doesn't evaluate this package's CSS at all (confirmed in PR 3/PR 6: nested `&` rules and `@media` blocks aren't applied to computed styles), so a source-level assertion is the reliable option available without a real browser.

The mobile drill-down for expandable/secondary-opening items (chevron-right navigating to a full-screen sub-page, with a back button) is out of scope (`SPEC.md` §10.18) — it needs its own navigation-stack state and a mobile-only `ItemExpandable` rendering path, a separate interaction mode rather than a CSS variation of what ships here.

## QA

- `bun run test` and `bun run check` pass, including the new focus-order and reduced-motion regression tests.
- Resize the viewport below 768px and confirm the menu/close-menu label swap and fullscreen overlay behaviour.

### PR readiness check

- [ ] Chromatic: skip — responsive behaviour change; leaving default Chromatic review enabled.

### DIFF
--- a/packages/react/ds-app/src/lib/SideNavigation/SPEC.md
+++ b/packages/react/ds-app/src/lib/SideNavigation/SPEC.md
@@ -1,11 +1,13 @@
 # SideNavigation — Component Specification
 
-> **Status:** Draft — rewritten against the 24.04 Side Navigation spec
-> (Figma: [App Mafia exploration](https://www.figma.com/design/4BxGmZUlhgWChPWu4lGzZv/24.04-Side-Navigation-exploration---App-Mafia?node-id=656-31955),
-> prose spec: *Side Navigation (App Layout)*). Anatomy, token pairings,
-> properties, state/keyboard, accessibility, and responsive behaviour are all
-> specified below; implementation lands across the PR sequence in
-> [§8 Implementation status](#8-implementation-status).
+> **Status:** Implemented, against the 24.04 Side Navigation spec (Figma:
+> [App Mafia exploration](https://www.figma.com/design/4BxGmZUlhgWChPWu4lGzZv/24.04-Side-Navigation-exploration---App-Mafia?node-id=656-31955),
+> prose spec: *Side Navigation (App Layout)*). All 8 PRs in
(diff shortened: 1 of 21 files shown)


potentialwook:
Superseded by the consolidated stack — see #1173.

PR #1167, opened 2026-09-07
refactor(ds-app)!: compose Item, ItemButton, ItemSwitch and ItemExpandable via children

potentialwook:
## Done

`Item`, `ItemButton`, and `ItemSwitch` each took a plain-string `label` prop for their row content, out of step with the rest of the design system (e.g. `Button`), where a component's own visible content is composed via `children`. `ItemExpandable`'s `label` conflated two different things — its own trigger text and the disclosed content — under `children` alone, with no way to pass both.

- `Item`, `ItemButton`, `ItemSwitch`: `label?: string` → `children?: ReactNode`.
- `ItemExpandable`: adds `heading?: ReactNode` for the summary's own label (matching `Accordion.Item`'s heading/children split); `children` is now exclusively the disclosed content.
- `NavTree` maps each data-driven entry's `label` field to `children` (or `heading`, for expandable entries) when rendering these subcomponents from the WD405 root/footerRoot data shape — the data vocabulary itself is unchanged, only how `NavTree` wires it up.
- `Footer`'s built-in footer-items renderer and `Group`'s stories/tests updated to compose the new children-based API.

Also, `SideNavigation.Content` already fell back to its own `children` when no `root` is given, but `SideNavigation` itself had no `children` prop and never forwarded one — a consumer wanting to compose the main navigation region directly out of JSX (e.g. to place a `ContextSwitcher` at its "content-defined position", `SPEC.md` §1.1/§4.5, without authoring the WD405 root data shape) had no way to reach that fallback short of using `SideNavigation.Content` directly and losing the Header/Footer/rail-state wiring `SideNavigation` itself owns. `SideNavigation` now accepts `children` and forwards it to `Content`, ignored when `root` is also given (`Content`'s existing precedence). Adds a `ComposedContent` story exercising `ContextSwitcher`, `Group`, `Item`, `ItemExpandable`, `ItemSwitch`, `ItemButton` and `Separator` together through this path, plus tests for the fallback and the root/children precedence.

**BREAKING CHANGE:** `SideNavigation.Item`, `SideNavigation.ItemButton`, and `SideNavigation.ItemSwitch` no longer accept a `label` prop — pass content as `children` instead. `SideNavigation.ItemExpandable` no longer accepts `label`; pass the trigger's own label as `heading` and the disclosed content as `children`. Consumers using `SideNavigation`'s own `root`/`footerRoot` data props (`LeafNavItem.label`, unchanged) are unaffected.

## QA

- New `ComposedContent` story — visually check `ContextSwitcher`/`Group`/rows composed directly as JSX children.

### PR readiness check

- [x] PR title follows Conventional Commits, marked `!` for the `label` → `children` breaking change.
- [ ] Chromatic: skip — no visual diff (prop rename plus a new fallback path).

### DIFF
--- a/packages/react/ds-app/src/lib/SideNavigation/SideNavigation.stories.tsx
+++ b/packages/react/ds-app/src/lib/SideNavigation/SideNavigation.stories.tsx
@@ -1,4 +1,5 @@
 import type { Meta, StoryObj } from "@storybook/react-vite";
+import { fn } from "storybook/test";
 import {
   lxdContentRoot,
   lxdFooterRoot,
@@ -11,6 +12,7 @@ import {
   withNavigationRouterProps,
   withNavLayout,
 } from "../../storybook/navigation/story-utils.js";
+import type { ContextSwitcherItem } from "./common/ContextSwitcher/types.js";
 import SideNavigation from "./SideNavigation.js";
 
 const meta: Meta<typeof SideNavigation> = {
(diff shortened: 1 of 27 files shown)


potentialwook:
Superseded by the consolidated stack — see #1174.

PR #1168, opened 2026-09-07
fix(ds-app): re-verify SideNavigation dividers, spacing and icon sizing; reinstate the overflow fade

potentialwook:
## Done

A three-commit Figma re-verification and fix pass (`SPEC.md` §9 token-inference-log convention), following on from PR 3's initial styling implementation:

**Group header case (§9.9):** Re-verified directly against the Figma spec (24.04 Side Navigation exploration — App Mafia, node 656:31955). Its group-heading component sets the text style's case to `UPPER` — a literal transform of the authored string ("Header" → "HEADER"), not a font feature. `GroupHeader` instead relied solely on `typography.heading.5` (ADR-T05), which uses `font-variant: small-caps` — for a mixed-case label this only affects the non-leading-capital letters, so group headers rendered inconsistently capitalised rather than uniform block capitals. Layers `text-transform: uppercase` on top of the token pick (the DOM text stays mixed-case, so this doesn't affect the accessible name, search, or copy/paste). Also corrects the group-header typography token itself, from `typography.heading.5` to `typography.text.tertiary.bold` (ADR-T05), and its colour, from `color.text` to `color.text.muted` (ADR-T03) — both re-verified against the same Figma node.

**Dividers, spacing and icon sizing (§9.10 and others):** Direct inspection of the Figma source file's fill data (not a rendered screenshot) found a divider at the Header/Content seam and another at the Content/Footer seam, both identical to the ones between item groups — the file draws **one** divider type throughout, not the previously-assumed divider-between-groups/gradient-at-the-frame-edges split (reversing ADR-T02/ADR-T07's original reading). `Header` renders its own seam separator directly; a footer's Content/Footer seam divider needs no special-casing — it's just its `root`'s trailing group setting the same per-group `separator` flag every other group divider uses. Also corrects a related icon-sizing/spacing pass across `Item`, `ItemButton`, `ItemExpandable`, `ItemSwitch`, and the collapsed header's own divider (rendered in place of, not in addition to, its usual trailing one).

**Overflow fade reinstated, row sizing unified:** The divider correction above initially removed `Content`'s overflow-fade entirely (not present anywhere in the Figma file, and its old always-on implementation reserved padding that broke the spec's stated item insets) — this last commit reinstates it under conditional rendering instead (CSS `scroll-state` container queries: the fade only shows when there's actually more content to scroll to in that direction), avoiding both problems. Also unifies row sizing (insets, icon-box dimensions) consistently across `Item`, `ItemButton`, `ItemExpandable`, and `ItemSwitch`, which had drifted slightly apart across PRs 3–6.

Every correction is logged in `SPEC.md` §9 with its own before/after and rationale, per the existing token-inference-log convention.

## QA

- Visual: compare group header casing/colour, seam dividers, and the conditional overflow fade against the Figma 24.04 exploration.

### PR readiness check

- [ ] Chromatic: skip — visual fixes; leaving default Chromatic review enabled.


potentialwook:
Superseded by the consolidated stack — see #1174.

PR #1169, opened 2026-09-07
feat(ds-app)!: give footer items control variants and rebuild ContextSwitcher on ContextualMenu

potentialwook:
## Done

**Feature — footer control variants + ContextSwitcher rebuild.** Extends PR 5's `control?: "link" | "button" | "switch"` field, previously content-only, to footer items (`FooterItem` gains the same field, defaulting the same way), so a footer row can be a switch or plain button, not just a link/logout-button split.

Rebuilds `SideNavigation.ContextSwitcher` on `ContextualMenu` (from `@canonical/react-ds-global`) instead of the `Popover`-based implementation from PR 5, for real select-like keyboard semantics: a genuine `<button>` trigger (`aria-haspopup="menu"`/`aria-expanded`) and a `role="menu"` popup with full roving-focus keyboard navigation (arrow keys, Home/End, type-ahead) — `Popover`'s native `<details>`/`<summary>` gave a plain list of buttons, not this. `title`, when given, still renders via `SideNavigation.GroupHeader` immediately above the dropdown field, as a real sibling rather than something `ContextualMenu` itself understands. `ContextualMenu` itself gains a `surfaceClassName` prop (its popup surface is portaled to `document.body`, so a plain `className` on the component's own root can never reach it — this is the one hook a consumer needing to theme the surface itself has).

**Fix — adapt to `ContextualMenu`'s current flat items API.** `ContextualMenu` moved to a flat `items: MenuEntry[]` API (with first-class separators) upstream of this branch's fork point — adapts `ContextSwitcher`'s construction to flatten its two groups (contexts, then an optional "create context" action) into one array with a separator between them, instead of the `groups: MenuItem[]` shape the previous commit was written against. Also fixes `ContextMenuItem` to compose `MenuItem` with `&` instead of `interface … extends` — `MenuItem` is itself a union (via the shared WD405 `Item`), which `extends` cannot target (TS2312). Fixes a leftover duplicate `:root` token block in `ContextualMenu`'s own `Item` stylesheet, introduced by the same adaptation and caught by `styles.layer.test.ts`.

**Fix — restore symmetric row inset on `Item`.** `Item`'s own row had drifted to a start-only inset (`padding-inline-start`) while `ItemButton`/`ItemSwitch`/`ItemExpandable` all use the symmetric `padding-inline` — a one-line drive-by fix, unrelated to the two items above, bundled here rather than as its own PR given its size.

**Fix — give the LXD footer's Log out fixture item its icon.** Its siblings (Documentation, Report a bug) each carry an icon; this one didn't, leaving it visually inconsistent in the LXD footer showcase — another small, unrelated drive-by.

**BREAKING CHANGE:** `ContextSwitcher`'s internal `ContextualMenu` integration changes its rendered DOM (native `<details>`/`<summary>` → a real `<button>` + portaled `role="menu"` popup) — any consumer test querying the old `Popover`-based markup needs to update to query the button trigger and `role="menu"` popup instead. No other package in this repo consumes `ContextSwitcher` yet.

## QA

- `bun run test` and `bun run check` pass, including updated `ContextSwitcher` tests exercising keyboard navigation on the new `ContextualMenu`-based popup.
- Manually exercise the context switcher dropdown: arrow keys, Home/End, type-ahead, and the "create context" action separated from the context list.

### PR readiness check

- [x] PR title follows Conventional Commits, marked `!` for the `ContextSwitcher` DOM/interaction change.
- [ ] Chromatic: skip — `ContextSwitcher` markup change; leaving default Chromatic review enabled.


potentialwook:
Superseded by the consolidated stack — see #1175.

PR #1170, opened 2026-09-07
feat(ds-app)!: specify SideNavigation against the 24.04 spec, render as a self-contained nav landmark

potentialwook:
## Done

Combines the spec authoring PR with the nav-landmark refactor (PRs 1–2 of the sequence tracked in `SPEC.md` §8).

### 1. Specify SideNavigation against the 24.04 spec (docs only)

- Rewrite `SideNavigation/SPEC.md` against the 24.04 Side Navigation spec (Figma: App Mafia exploration; prose spec: Side Navigation (App Layout)):
  - Full anatomy for primary (expanded/collapsed) and secondary navigation, including the mandatory collapsible Help item with an external legal link, group headers, separators, and the context switcher.
  - Exact layout/spacing values from the spec, mapped to existing dimension tokens where a match exists.
  - Token pairings for navigation background/text/icon/border states, with a token inference log recording every value that has no existing semantic token or is inferred by name rather than confirmed against the spec's colour swatches (which only exist as images in the source document).
  - A full properties table, including a closed-vocabulary `footerItems` prop (superseding the current free-form `footerRoot` for spec conformance) and a discriminated `NavItem` union (`LeafNavItem | ExpandableNavItem`) that encodes the spec's depth-1 nesting and no-link-on-expandable constraints at the type level.
  - Keyboard shortcut scaffolding recorded as shipping disabled by default: the spec's own prose contradicts itself on the key (Ctrl+E vs. a single letter), so the binding is designed but not enabled pending a design ruling.
  - Collapsed rail width and logo aspect ratio recorded as inferred (logo width + one spacing unit each side, logo centred; logo aspect ratio 1:2), since the spec never states either directly.
  - A known-issues section enumerating every spec ambiguity and inference for design/engineering follow-up.
- Amend the existing Header/Content/Footer/Item anatomy YAML files to match.

### 2. Render SideNavigation as a self-contained nav landmark

**BREAKING CHANGE:** `SideNavigation`'s root element changes from `<div>` to `<nav>`, with `aria-label` defaulting to `"Main navigation"` (overridable). Consumers querying by role or asserting on the root tag (e.g. `getByRole("generic")`) need to update to `getByRole("navigation")`. No other package in this repo consumes `SideNavigation` yet, so the blast radius is contained to this PR.

Per `SPEC.md` §1 and §6 (AC1: "The Side Navigation must have a self-contained `<nav>` tag"):

- `SideNavigation.tsx` now renders `<nav>`, not `<div>`.
- Every `SideNavigation` prop-types file converts from `interface extends HTMLAttributes<HTMLDivElement>`/`ButtonHTMLAttributes<HTMLButtonElement>` to the house style — a `type` alias intersecting an `OwnProps` object with `Omit<ComponentProps<"tag">, keyof OwnProps>`, per this repo's React component props convention. This was more than a mechanical rename: none of these types previously extended their root element's native props at all, so native attributes like `id`, `data-*`, or `aria-*` were previously accepted at runtime (via an untyped `...props` spread) but rejected by the type checker. They are now both accepted and typed.
- `CollapseToggleProps` additionally excludes `type` from its native surface (documented inline) since this is always a disclosure trigger, never a submit/reset control.
- `NavTree` (private, not part of the public barrel) gains the same `className`/native-prop passthrough as its siblings.
- Adds two tests asserting the nav landmark and the `aria-label` override.

## QA

- New tests assert `getByRole("navigation", { name: "Main navigation" })` and the `aria-label` override.

### PR readiness check

- [x] PR title follows Conventional Commits, marked `!` for the breaking root-element change.
- [ ] Chromatic: skip — root tag change may affect snapshots; leaving default Chromatic run enabled.

### DIFF
--- a/packages/react/ds-app/src/lib/SideNavigation/common/Content/SideNavigationContent.anatomy.yaml
+++ b/packages/react/ds-app/src/lib/SideNavigation/common/Content/SideNavigationContent.anatomy.yaml
@@ -1,7 +1,10 @@
 ---
 # SideNavigation.Content — required main navigation region. Grows to fill the
 # space between Header and Footer. Renders the direct children of its `root` Item
-# (the root itself is a non-rendered container) as a nav item list.
+# (the root itself is a non-rendered container) as a free composition of an
+# optional context switcher followed by content entries — each entry is either
+# a group (optional header + its navigation items) or a standalone separator.
+# See SPEC.md §1.1 and §4.3 for the full content vocabulary.
 #
 # The shell grows past the viewport (min-height), so the page scrolls rather than
 # this region: overflow is vertical-only as a safety valve, never horizontal.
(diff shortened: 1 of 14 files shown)


packages/react/ds-app/src/lib/SideNavigation/common/Content/SideNavigationContent.anatomy.yaml
```diff
@@ -15,23 +18,45 @@ node:
...
-          # via the WD405 Item.Component substitution point.
-          - switch:
-              "on": override
-              cases:
-                - uri: apps.subcomponent.side-navigation-item
-                - uri: $custom
-            relation:
-              cardinality: "0..*"
-              slotName: default
+        uri: apps.subcomponent.side-navigation-context-switcher
```
advl: Leave out the context switcher for now. Also it doesn't seem to exist anywhere else in the pr. Let's track as follow-up

packages/react/ds-app/src/lib/SideNavigation/common/Content/SideNavigationContent.anatomy.yaml
```diff
@@ -15,23 +18,45 @@ node:
...
+                          - uri: apps.subcomponent.side-navigation-item
+                          - uri: apps.subcomponent.side-navigation-item-expandable
+                          - uri: apps.subcomponent.side-navigation-item-button
+                          - uri: apps.subcomponent.side-navigation-item-switch
+                          - uri: $custom
+                    relation:
+                      cardinality: "0..*"
+                      slotName: default
+            - node:
+                uri: apps.subcomponent.side-navigation-separator
```
advl: I'd like to hear more about this choice of anatomy and what it means in relation to the navigation type contract

potentialwook:

PR #1171, opened 2026-09-07
feat(ds-app)!: style SideNavigation to the 24.04 spec, add group/header/separator/expandable items

potentialwook:
## Done

Combines the full-spec styling pass with the group/header/separator/expandable content-model refactor (PRs 3–4 of the sequence tracked in `SPEC.md` §8).

### 1. Style SideNavigation to the 24.04 spec

Implements `SPEC.md` §2/§3 (AC2: "The Side Navigation must follow the styling in the spec"):

- New provisional token file `packages/styles/main/src/navigation.css`: rail width (240px), collapsed rail width and logo width (both derived, not spec-stated — see `SPEC.md` §9.2/§9.3), shared row insets (1rem/1.5rem) and icon↔label gap (0.5rem), and a fixed leading-icon column width (1rem, matching the `Icon` component's own default render size).
- `SideNavigation` now owns its own rail width and spans the full parent height (previously `max-height`); overrides the shared `--overflow-gradient-height` to the spec's 1.75rem locally.
- `Header` lines up the logo, collapse toggle, and every item's icon on the same 1rem/1.5rem/0.5rem rhythm.
- Item rows get real colour states (background base/hover/active/disabled; text/icon dim on disabled) — previously deferred/structural-only.
- `NavTree` groups get the spec's 0.5rem/1rem block padding and a top divider on every group but the first; the group header adopts the heading-5 typography token pairing.
- `Content`'s overflow fade gains a matching top instance.
- `CollapseToggle` gets a real hover/active background and a fixed `"button"` type.

Every inference is logged in `SPEC.md` §9 for design consolidation.

### 2. Group, header, separator and expandable navigation items

**BREAKING CHANGE:** `SideNavigation`'s content model splits from one flat, recursive `NavItem` into a proper tiered shape — `NavRoot` (`root`/`footerRoot`) > `NavGroup` (a group's own items) > `NavItem` (`LeafNavItem | ExpandableNavItem`). An `ExpandableNavItem`'s own `items` are typed as `LeafNavItem[]` only, so a child of an expandable item cannot itself expand, and it cannot have a `url` — both enforced at the type level per `SPEC.md` §4.3. Existing `NavItem`-typed data at root/group position needs to retype as `NavRoot`; existing 3+-level trees (if any) will no longer compile — this is the depth-1 constraint doing its job, not a bug. `Item.tsx`'s `items` prop (and its static disclosure caret) is removed — an entry with children is now `SideNavigation.ItemExpandable` instead.

New public subcomponents, all attached to `SideNavigation` by dot notation alongside a new `SideNavigationComponent` type:

- `SideNavigation.Group` / `.GroupHeader` — a labelled collection of entries. Owns the spec's 0.5rem/1rem group padding and the top divider on every group but the first.
- `SideNavigation.Separator` — a plain rule between content sections, a sibling of `Group` among the content's direct entries.
- `SideNavigation.ItemExpandable` — a collapsible row over native `<details>`/`<summary>`. Uncontrolled, seeded by `defaultExpanded`; `NavTree` seeds the initial value from `useNavigationTree`'s `inSelectedBranch` so the ancestor chain of the active page starts expanded (`SPEC.md` §5).

Also fixes a real, pre-existing gap: no `SideNavigation` subcomponent was reachable through the package's public export surface at all — `common/index.ts` existed but was never re-exported from `SideNavigation/index.ts`, and none were attached to `SideNavigation` by dot notation.

`NavTree` drops its own `.group`/`.header`/`.list` styling (moved into `Group`/`GroupHeader`) and gains a third loop for an `ItemExpandable`'s children, stripping the tree-annotation fields before spreading an entry onto a presentational subcomponent — previously spread through unfiltered, a latent leak onto the DOM this change also closes.

Fixes `fixtures.tsx` and the test roots to retype as `NavRoot` (their actual data shape was already correct; only the type annotation was wrong). Adds missing test/story coverage for `Group`, `GroupHeader`, `Separator`, and `ItemExpandable`.

**Companion fix, folded in rather than split out** (it only becomes necessary the moment `NavGroup`/`NavSeparator`/`NavRoot` exist): `@canonical/ds-types`'s shared `Item` requires a discriminated key-or-url identity, which `useNavigationTree`'s own `T extends Item` bound enforces structurally. `NavGroup`, `NavSeparator` and `NavRoot` left `key` fully optional even though every real usage already supplies one (`annotateTree`/`getItemId` already require it at runtime) — made required to close that gap. `_AnyNavNode` gains the same identity union, with a single documented cast bridging it to `LeafNavItem`'s own, deliberately looser shape (key/url both optional, by design). Also retypes `Content`/`Footer`'s `root` prop from the generic WD405 `Item` to this package's own `NavRoot`.

## QA

- Visual: rail width, row insets, and divider placement against the Figma 24.04 exploration.
- New Storybook stories for `Group`, `GroupHeader`, `Separator`, `ItemExpandable` — visually check disclosure open/close and the group-divider placement.

### PR readiness check

- [x] PR title follows Conventional Commits, marked `!` for the content-model breaking change.
- [ ] Chromatic: skip — visual styling and new subcomponents; leaving default Chromatic review enabled.


potentialwook:

PR #1172, opened 2026-09-07
feat(ds-app): navigation control variants and collapsed rail behaviour

potentialwook:
## Done

Combines the navigation control-variants feature with collapsed-rail behaviour (PRs 5–6 of the sequence tracked in `SPEC.md` §8).

### 1. Navigation control variants — ItemButton, ItemSwitch, ContextSwitcher

Includes a small companion fix (`fix(react-ds-global-form)`) needed for this feature: the package declared only the bundler-only `module` field — no `main` and no `exports` map — so nothing outside its own package could actually import it. Adds an `exports` map for the JS entry point and the CSS subpath already imported elsewhere (`apps/react/boilerplate-vite`), following the same pattern already used by `@canonical/utils`. Also curates `SwitchInput`/`SwitchInputProps` onto the public barrel, needed by this PR's `SideNavigation.ItemSwitch`. (A second, previously-unexercised consumer — `apps/react/storybook-hub`'s own stylesheet — imported the package's un-built source directly; repointed at the same `dist/esm/index.css` subpath.)

Implements `SPEC.md` §4.4/§4.5 (AC3: "a polymorphic link component variant including a button, a switch, and a dropdown selection"; AC4: same three listed as the required set):

- `SideNavigation.ItemButton` — a real `<button>`, for actions instead of navigation ("Log out", "Create context"). Native `:disabled`/`:hover` styling.
- `SideNavigation.ItemSwitch` — a `<label>` wrapping `SwitchInput`, so activating anywhere on the row toggles it via native label-wraps-input association.
- `SideNavigation.ContextSwitcher` — a dropdown built on the global `Popover` pattern (native `<details>`/`<summary>`), showing the current context, a user-generated context list with optional descriptions/badges, and an optional "create context" action. (Later rebuilt on `ContextualMenu` for real select-like semantics — see the last PR in this stack.)

`LeafNavItem` gains a `control?: "link" | "button" | "switch"` field (default `"link"`) plus the fields each variant needs — one flat shape across all three, not a discriminated union (`SPEC.md` §10.11, a deliberate authoring-simplicity trade-off). `NavTree`'s `renderEntry` helper dispatches on `control`, stripping whichever fields the chosen variant doesn't use.

All three are attached to `SideNavigation` by dot notation. Fixes a latent bug: `vitest.setup.ts`'s `ResizeObserver` mock could not be constructed as a constructor (`Popover`'s positioning hook needs a real one).

### 2. Collapsed rail behaviour, tooltips, inert shortcut scaffold

Implements `SPEC.md` §1.2/§7 (collapsed state) and §5/§9.4 (tooltips, keyboard shortcut):

- The rail narrows and `Content` is hidden entirely while collapsed — the spec's own deliberate scope cut (icons aren't guaranteed, context switchers can't render collapsed, expandable state is undefined, the toggle needs to move, icon-only buttons are poor UX).
- `Header` switches to a centred column when collapsed: the logo centres and the collapse toggle relocates below it purely by changing `flex-direction`.
- The footer stays visible but goes icon-only when collapsed (labels return via tooltip in a follow-up, `SPEC.md` §10.14).
- `CollapseToggle` gains a 1-second hover tooltip reading "Collapse" or "Expand".
- New `common/hooks/useCollapseShortcut`: binds Ctrl+E to the rail toggle, but only when `enabled` — `SideNavigation`'s new `keyboardShortcut` prop defaults to `false`, so no listener ever attaches. Scaffolded per `SPEC.md` §5/§10.1 because the spec's own prose is self-contradictory on the key — ships inert pending a design ruling.

Drive-by: the Storybook `withSideNavShell` decorator wrapped subcomponent stories in a `<div>`; updated to `<nav>` to match the real component's root.

## QA

- New Storybook stories exercise `ItemButton`, `ItemSwitch`, `ContextSwitcher` — verify keyboard toggling and dropdown open/close.
- Manually toggle collapse/expand and confirm the tooltip text, the header's centred layout, and the footer's icon-only state.
- Confirm Ctrl+E does nothing by default.

### PR readiness check

- [ ] Chromatic: skip — new components and collapsed-state changes; leaving default Chromatic review enabled.


potentialwook:

PR #1191, opened 2026-09-08
fix(styles)!: close three leaks in the Vanilla boundary

advl:
## Done

An adversarial correctness pass over the merged adapter found that three of its stated guarantees do not hold on an ordinary page. This closes all three, corrects two claims that were simply wrong, and rewrites the limitations section so that what is left unfixed is explained rather than noted.

Every fix is measured against a page running pragma alone with the same markup, because a check that only reads the mixed page passes on a value that is wrong on both.

### The order statement collapses the page, silently

A cascade layer takes its place the first time the browser sees its name, and no later statement can move it. So a stylesheet that mentions a `ds` layer before `layers.css` is read puts `ds` at the bottom of the order, below `boundary` — and the boundary's `all: revert` then runs over pragma's own rules instead of Vanilla's.

Measured, the same island on the same page:

| sheets, in order | `<p class="ds">` | card border |
|---|---|---|
| `layers.css` first | 14px, no margin | 1px |
| a component sheet first | 16px, 16px margin | 0px |

Every pragma component renders as bare markup. Nothing appears in the console, and there is no partial mode.

`adapter.css` now imports `layers.css` as its own first rule, so the single-import path cannot go wrong. The requirement is stated in installation step 3 rather than implied, step 7 gains the check, and `tests/boundary.test.ts` measures both sides.

### Five Vanilla `!important` rules reach inside an island

`all: revert` cannot touch an important declaration. For important rules the cascade reverses the layer order, so the declaration in the **first** layer wins and one written into `boundary` loses to one in `vanilla`. Measured: a list under `.u-text-max-width` kept `max-width: 40em` against an important revert in `boundary`.

Vanilla ships exactly six rules whose important declaration has a class-free subject compound, enumerated from both compiled builds and byte-identical in 4.56 and 4.58. Five of them match an element inside an island while the Vanilla class sits on an ancestor **outside** it, and four counters answer those, because one covers both table-layout utilities — which is the arrangement the rules recommend, `<div class="col-6"><div class="ds card">`.

| Vanilla rule | mixed page | pragma alone |
|---|---|---|
| `.u-text-max-width ul, ol` | `max-width: 640px` | `none` |
| `.u-table-layout--fixed table` | `table-layout: fixed` | `auto` |
| `.p-content-card__author-and-date > :first-child` | `margin-bottom: 0px` | `16px` |
| `.u-vertically-center > img` | `align-self: center` | `auto` |

The counter goes in `vanilla.escapes`, a sublayer of `vanilla`. A rule written directly into a layer lands in that layer's implicit final sublayer, which sorts after every named sublayer, so the counter beats a consumer's own `@layer vanilla { … }` block whichever order the two sheets arrive in. Measured both ways.

Each rule mirrors Vanilla's own selector and adds the island requirement, and names a single property rather than `all`, so it fires only where the leak is. The cost is stated in the README rather than hidden: inside those four utilities the important revert also outranks pragma's own value for that one property. Narrowing further is not possible, because excluding pragma's own elements hands them straight back to Vanilla.

The sixth, `* { animation: none !important; transition: none !important }` under reduced motion, is left alone: pragma honours the same preference by zeroing its duration tokens and the two agree.

### An element that is itself an island root got almost nothing

A relative selector inside `@scope` never matches its own scoping root, and only three selectors had a `:scope` twin. Measured, mixed against pragma alone:

| markup | mixed | pragma alone |
|---|---|---|
| `<pre class="ds">` | the baseline sans font | monospace |
| `<h2 class="ds">` | weight 400, line-height normal, 19.92px margins | weight 300, 32px, no margin |
| `<p class="ds">` | 16px, 16px margins | 14px, no margin |
| `<sub class="ds">` | 16px, `line-height: normal` | 12px, `0` |

Every element rule now carries a root-reaching twin, and the binding test generates them rather than comparing text, so it proves the property instead of restating the mapping.

Three properties then had to be split out of the root baseline, each with its own exclusion list, because `normalize` sits below `ds.reset` and the baseline would otherwise beat an element's own rule on a root that is that element. The lists are inert on a document element, which is never a `<pre>`, so they live in pragma's own files and the copy inherits them mechanically. A form control keeps the baseline family and size on purpose, since `normalize` gives it `inherit` and `100%`, which on an island root would resolve against the host page.

### Inherited properties cross, and `revert` cannot undo inheritance

Rolling an inherited property back to the user agent leaves it at the parent's value. Measured with Vanilla present against absent, a Vanilla ancestor pushed uppercase headings, italic captions, small caps, old-style figures, letter spacing, alignment, list markers, `user-select` and table collapse into an island.

The island root now declares the typographic inherited properties Vanilla sets anywhere in its build, enumerated from the compiled stylesheet. Four are deliberately left to cross, because a host setting them is telling the island something rather than styling it: `direction`, `visibility`, `pointer-events` and `cursor`.

One deliberate difference remains, and the fixture asserts it as one rather than diffing it away. An island root does not inherit a table cell's text alignment, because the same property is how Vanilla's `u-align--*` family would otherwise reach in and nothing can tell a browser default from a framework rule.

### The monospace fix did not survive minification

Blink gives the bare `monospace` keyword a smaller default size, and `normalize.css` defeated that by naming the family twice. Lightning CSS 1.32 collapses a repeated name, so `monospace, monospace` minifies to `monospace` and a `<pre>` goes from 16px back to 13px. `ui-monospace, monospace` is two different names, measures the same 16px, and no minifier collapses it; `ui-monospace` falls through wherever it is not supported. It is also what the rest of this repository already writes: ten stylesheets under `packages/react/tokens` and `packages/storybook/helpers` use that exact list, so this makes `normalize.css` consistent rather than introducing a spelling of its own.

### What the README now says

Every guarantee names the file that actually holds its assertion. Two named fixtures that did not contain the checks beside them.

"Known limitations" becomes "What this package does not fix", and each entry explains the difference instead of noting it: the order statement, the seven unprefixed pragma class names that collide with host markup, the Vanilla surfaces the bridge cannot see, the table-cell alignment, presentational attributes, the properties left to cross, and the minifier.

Two claims were wrong and are corrected. The entries do reference a file — pragma's criticality modifiers declare an icon URL, which nothing reads today but which would 404 if something did. And Vanilla's important declarations no longer "still apply".

## QA

`bun run test` in `packages/styles/vanilla-adapter`: 85 cases pass against Vanilla 4.56 and 4.58, up from 62. `bun run check` passes there and biome passes in `styles/main` and `styles/typography`.

The four important declarations carry a biome suppression naming the reason, which is that an important declaration can only be answered by another one from a lower layer.

## Screenshots

None here, but expect a Chromatic diff on `pre`, `code`, `kbd` and `samp`: `ui-monospace` resolves to a different face from `monospace` on platforms that have one. That is the intended change.

### PR readiness check

- [x] PR title follows the Conventional Commits format. It is marked breaking: an island root now computes different values for several inherited properties, and the monospace family changes on a pragma-only page.
- [x] The code follows the appropriate code standards. The four `!important` declarations carry a `biome-ignore` naming why nothing else can answer an important declaration.
- [x] All packages define the required scripts. No script changed here.
- [ ] No new package.
- [ ] No `Chromatic: skip`: the monospace family changes, and although it resolves to the same face on a Linux runner it will differ for a reviewer on a Mac, so the run is worth having.

## What the review round changed

Four review lanes went over the branch. Three found real defects, all now fixed and covered.

**The root baseline flattened what the browser gets right.** An `<em class="ds">` computed upright where pragma gives italic, and the same for `i`, `cite`, `address`, `var` and `dfn`. A `<th class="ds">` computed weight 400 against 700 and `start` against `center`. `font-weight`, `font-style` and `text-align` now sit in rules of their own with the exclusion list each property needs, the same shape the earlier three already had.

**The content-card counter made things worse for the elements pragma already zeroes.** A `<p class="ds">` as that first child took 14px where pragma gives 0px, because Vanilla's own `margin-bottom: 0 !important` agreed with pragma there and the revert replaced the agreement with the browser's `1em`. It now excludes them, and it gains the `:first-child` its mirror was missing, without which a second `.ds` child of the same container lost its own margin.

**`body:not(…)` raised a rule's specificity.** A `:not()` takes the specificity of its most specific argument, so the bare list moved the typography base font from (0,0,1) to (0,0,2) and it began beating a later rule in the same layer that used to win on order. `body:where(:not(…))` measures identical to the bare `body` it replaced.

**`list-style-type` was briefly declared and is removed again.** No value is right both inside a host `<ol>`, where a marker is a number, and inside a Vanilla `.p-list`, which asked for none. An island root that is a list item belongs to the host's list, so the property joins the ones left to cross.

**The counts were wrong.** Vanilla ships six rules whose important declaration has a class-free subject. Five reach inside an island, and four counters answer them because one covers both table-layout utilities. The files said six and seven, on three different bases.

The suite is at 85 cases from 62. Each new case was mutation-tested: reverting the fix it covers turns it red.

## Chromatic

The only change that can move a pixel is the monospace family, and on Linux it moves none. Measured: `monospace`, `monospace, monospace` and `ui-monospace, monospace` all resolve to the same platform face with the same metrics, because Chromium maps `ui-monospace` to the platform fixed font off Apple platforms. A reviewer on a Mac will see SF Mono, which is a local artifact rather than a diff.

The whole Storybook estate re-snapshots all the same, because every workflow lists `packages/styles/**` in its `externals` array and TurboSnap disables itself when an externals file changes. Expect a large diff count that is almost entirely noise.

## Two things found in passing, neither this PR's to fix

`--font-family-mono` is read at three places under `react-ds-app-launchpad` and defined nowhere. The higher-layer declaration wins and then becomes invalid at computed-value time, so the element inherits its parent's family rather than falling back. Measured: a `<pre>` under such a rule computes the ancestor's sans-serif. The diff viewer renders code in a proportional font today.

Every Chromatic workflow triggers on `packages/tokens/**`, a directory that does not exist since a rename. `packages/react/tokens` is consequently ungated, as are the two Storybook addons.

## One limitation the new exclusions do not retire

The `th` exclusion applies when the island root **is** a table header. An island **inside** one still does not inherit its centring, which is the trade named under "What this package does not fix" and asserted as a deliberate difference in the fixtures.

## Second review round

A second pass over the fixed branch found that the exclusion lists were right in principle and wrong in detail, in both directions.

**Where the browser's value is relative, excluding an element hands it to the host.** That is the opposite of what the exclusion was for. Measured with an island root inside a Vanilla heading whose text is 24px:

| island root | mixed | pragma alone |
|---|---|---|
| `<pre class="ds">` | 24px | 16px |
| `<small class="ds">` | 19.2px | 12.8px |
| `<b class="ds">` | weight 400 | 700 |

The size pin now covers every island root, with `small`, `sub` and `sup` given the same proportion of the root instead. `pre` and its family need no exclusion at all: their smaller size on a bare page is Blink shrinking the `monospace` keyword, which the two-name family list already defeats. `b`, `strong` and `optgroup` keep their exclusion and gain the absolute weight the browser would have computed.

**Three lists were wrong the other way.** `code`, `kbd` and `samp` have no user-agent white-space, so excluding them bought nothing and let a host's `nowrap` through. The button-typed inputs and the file input, which do have one, were missing from that list, and the button-typed inputs were missing from the alignment and selection list. And `color` had no list at all, so the root declaration was overriding the grey the browser gives an `<hr>` and the label colour it gives a button.

The suite is at 85 cases. Three new island-root hosts were added: a Vanilla heading, a muted heading and a truncating container. Each fix was mutation-tested.

## Third review round

A second pass over the documentation and the fixtures found that several of the fixes were guarded only by a string comparison, and the prose had drifted behind the code again.

**The headline fix of round two had no behavioural case.** Removing the exclusion list from the content-card counter left the whole suite green, because only the order test's selector-string comparison noticed. A paragraph and a heading as the first child of a Vanilla content card now measure it, at the numbers the code comment cites.

**Nine members of the exclusion lists could be deleted with the suite still green**, among them two the comments call load-bearing and one the README promises. Each now has an island root of its own, and three more properties join the comparison.

**One inherited property had no falsifying host.** The only fixture with a `text-transform` ancestor is the island inside a table header, which the comparison skips because its alignment differs by design. It has a case of its own now, and the deliberate difference grew a second half: an island inside a header does not inherit its boldness either.

**And the minification claim had nothing behind it**, because nothing in the fixtures minifies. A source assertion now requires both monospace lists to name two different families.

The documentation follows. The second boundary's cost is stated in the README, which two other files already claimed it was. The order contract gains the second way to defeat it, measured: a named sublayer of Vanilla arriving before the statement places itself ahead of this package's own, and an important declaration there beats the counter. Eleven stale or mangled cross-references are repaired, and four comments that counted rules or properties are corrected.

The suite is at 85 cases from 62.

## Fourth review round

The last lane audited the tests rather than the code, and found the binding test could be fooled.

**The mapping table passed an unrecognised selector shape through unchanged.** So a rule pasted into the copy verbatim passed, even though a relative selector inside `@scope` never matches its own scoping root and the rule could not reach an island root at all. Writing the *correct* twin failed instead. A selector list mixing a pseudo-element with a bare element was mandated wrong for the same reason. An unrecognised shape is now reported by name rather than guessed at.

**The second boundary's inventory was a transcript of the adapter**, so it saw drift in the adapter and nothing in Vanilla. Three mutations to the counters failed only that string comparison, and a Vanilla release adding a class-free important rule would have gone unnoticed. It is now derived from each compiled build.

**One counter turns out not to work, and is removed rather than kept.** Vanilla's `.u-table-layout--auto` forces the browser's own default, so reverting lands on the value Vanilla already forced and a table that asked for `fixed` cannot get it back. Measured, `auto` either way. A counter that changes nothing reads as cover, so the README names the leak instead.

Coverage follows: an island root that *is* the list or table a counter reaches, a search input for two rules nothing read, and the four properties deliberately left to cross, which nothing pinned. Two vacuous assertions are made observable and an excepted rule's declarations are compared rather than skipped.

The suite is at 91 cases from 62, and runs in 26 seconds.

## A note on the CI browser

Every measurement here was taken in Chromium 151, the system browser. CI installs Playwright's pin, Chromium 149, which cannot be launched on this machine. The branch adds absolute assertions, so a 149-specific difference would show up in CI rather than locally. Nothing suggests one, but it has not been ruled out here.


anthonydillon:
This was a rather large change without a review. It appears you didn't wait for copilot to complete it's review and therefore ignored it's recommendations. This brings into concern the confidence of the project. In projects of this maturity we should not have unchecked merges. @advl

PR #1210, opened 2026-09-10
feat(ds-app)!: rebuild SideNavigation into a single data-driven region shell

potentialwook:
## Done

A full rebuild of `SideNavigation` against the 24.04 design spec, delivered as a three-commit series on top of main (spec, then the component rebuild, plus a supporting `ContextualMenu` fix):

- **One navigation landmark.** The root is a plain `<div>` layout container; the component's single `<nav>` landmark is Content's, with `aria-label` forwarded and defaulting to `"Main navigation"`. Branding (Header), the context switcher and the footer's actions sit outside the landmark by design, so screen readers announce exactly one navigation region.
- **Skip navigation.** A visually hidden "Skip to main content" link is the first focusable element in the component's DOM order (target via `skipTo`, default `#main-content`), letting keyboard users bypass the navigation block.
- **Four regions, all data-driven:** Header (branding, collapse toggle), an optional ContextSwitcher region (typed `contextSwitcher` props — a `role="menu"` select-like widget is not navigation, so it renders in its own plain-`<div>` region between Header and Content, outside the landmark, and hides when the rail collapses), Content (the landmark; links, plain labels and expandable disclosures only) and Footer (free-form rows via `footerRoot`).
- **Strict content vocabulary.** Interactive action rows belong to the Footer, so the landmark's contents always read as destinations. There is no JSX escape hatch — the API carries no `children` props; every subcomponent is private (`cs:react.component.subcomponent_export_api`) and renders exclusively from data.
- **Footer vocabulary.** `footerRoot` is the footer's only data surface: free-form leaves (`LeafFooterItem` — `label` required, optional `icon`/`url`/`control`/`slot`; buttons via `control: "button"` + `onClick` flow through data) and depth-1 expandables (`ExpandableFooterItem`, mirroring the content tree's `ExpandableNavItem` split). Leaf rows render as links (active when the row matches the current location, with `LinkComponent` integration); action rows render as buttons.
- **Collapsed rail.** All state styling keys off `data-expanded` on the root — the single source of truth. Collapsed: Content and the switcher region hide, the footer degrades to icon-only, and any rendered `ItemExpandable` discloses its sub-items into a floating popover on the inline-end with visible labels. Where CSS anchor positioning is supported (Baseline 2026), `position-try-fallbacks` flips the popover above its trigger on viewport overflow — browser-measured, plus the mirrored 0.5rem gap — otherwise it stays top-anchored. Native `<details>` throughout: no Esc-close/outside-click dismissal by design, Tab navigation stays viable.
- **Auto-expand on active.** A nested sub-item whose url becomes active re-opens its collapsed parent (one-way: manual collapse is respected, never auto-closed) and carries `data-active` + `aria-current` under the expanded parent.
- **Ctrl+B rail-collapse shortcut** — opt-in via `keyboardShortcut`, off by default (pending approval).
- **Supporting changes:** `ContextualMenu` gains a `surfaceClassName` hook (the one way to theme the portaled surface); the provisional sidenav token sheet in `@canonical/styles` is retired — the architecture tokens now live in the component's own stylesheet with motion-token durations, and the row inset is unified to one 0.5rem inline value.
- **Docs.** SPEC.md specifies region semantics (including the switcher region and the single data-driven consumption pattern), properties, rail state, keyboard behaviour, accessibility and the deliberate exclusions. All root fixtures live in `storybook/navigation/fixtures.tsx`; the stories are data-driven, with a practical `with ContextSwitcher` story (url'd contexts routing through the hash router, both `ItemExpandable` flavours side by side).

## QA

- `bun run check` and `bun run test` pass from the repo root (62 projects).
- Storybook: MAAS/LXD data-driven; `with ContextSwitcher` — switch "staging" routes via the hash router, "Create context" fires the action; `Collapsed` — content + switcher hidden, footer icon-only, footer expandables open as inline-end popovers; route to a nested sub-item — parent auto-opens with the sub-item highlighted.
- Keyboard: Tab → first stop is "Skip to main content" → Enter jumps past the navigation.

### PR readiness check

- [x] PR title follows the Conventional Commits format; breaking change marked with `!`.
- [x] All packages define the required scripts in `package.json`.
- [ ] Not a new package.
- [ ] `Chromatic: skip` **not** set — visual-heavy diff, Chromatic review wanted.


potentialwook:
Split into a six-PR stack for review, replacing this PR. Same changes, dependency-ordered — merge in order:

- #1225 fix(react-ds-global): TooltipEngine export + ContextualMenu surfaceClassName
- #1226 docs(ds-app): SideNavigation SPEC.md + anatomy
- #1227 feat(ds-app)!: the SideNavigation rebuild
- #1228 fix(ds-app): rail-collapse behaviour (shortcut default-on among it)
- #1229 fix(ds-app): footer label rows + stable keys
- #1230 refactor(ds-app): cleanup (CSS/token truth pass, cascade layer, citations)

The working-tree leftovers from this branch (shortcut default flip, story fixture swap) are folded into #1228 and #1230 respectively.

PR #1213, opened 2026-09-10
chore(cli): re-derive the perf ceilings, and keep the pass manual

advl:
## Done

`bun run test:perf` was failing on `main` before this change, and for two reasons that hid each other.

- **Two ceilings were genuinely exceeded.** Re-derived by the rule the file already states — 2× the measured median for the fast paths, a projected p95 for the store — with the arithmetic written out at each constant and the environment recorded:

  | Constant | Designed | Median | Old | New |
  |---|---|---|---|---|
  | `BUDGET_HELP_MS` | 50, unmet | 103.3 ms | 130 | **210** |
  | `BUDGET_COMPLETE_MS` | 50, unmet | 115.7 ms | 150 | **240** |
  | `BUDGET_WARM_STORE_MS` | 300, unmet | 536.0 ms | 500 | **850** |
  | `BUDGET_PROJECT_CONFIG_MS` | 10 | 0.011 ms | 10 | left — 900× headroom |
  | `BUDGET_MCP_P95_WARM_MS` | 100 | 0.506 ms | 100 | left — 137× headroom |

  The designed 50 ms and 300 ms targets stay recorded as designed-and-unmet rather than quietly moved. Two constants are left alone deliberately: their medians are sub-millisecond, so the 2× rule would assert only jitter.
- **A per-test timeout was underneath.** Raising the ceilings did not make the pass green — it exposed vitest's 5 s default, which the failing assertions had been winning the race against. The store case's own duration on `main` was ~6.9 s per attempt. `testTimeout` is now derived rather than picked: 2 × (12 runs × 850 ms) ≈ 20000 ms.
- **The measurements refuse a contended box.** This machine drifts 1.5–2×, and netting against a control does not rescue a uniformly shifted run, so the harness gates on the in-run control median and **discards** rather than averages: 5 of 9 fast-path repetitions accepted, 4 of 7 store repetitions.
- **The pass stays manual, and that is now durable.** It is referenced by no workflow, no Nx target, no git hook and no script chain — verified mechanically. Its config header, `docs/CI.md`'s "what CI deliberately does not run", and an `AGENTS.md` bullet all now say so, so re-adding it is a deliberate act rather than an accident. No test asserts a workflow file's contents, because this repo does that nowhere.
- A dated section in `BUDGETS.md` records the new measurements beside the old; no past measurement was edited.
- Cause noted at the constants, for whoever reads them next: the eagerly-imported capability barrel grew 37.6 → 68.0 ms, and 68.0 of `--help`'s 69.1 ms of work is that barrel. `lazy.test.ts` is green, so this is surface growth rather than a lazy-boundary leak. Meanwhile the embedded pack grew 8479 → 49630 triples (5.85×) for a 1.75× store cost.

## QA

```bash
bun install && bun run --cwd packages/cli/pragma build
bun run --cwd packages/cli/pragma test:perf     # 22/22, ~20 s; run it twice
bun run check                                   # green, 66 projects
```

To see the state this replaces, run `test:perf` on `main`: 21 of 22, with all three warm-store retries red.

Two pre-existing conditions on this machine, neither caused here and both verified against `main` in the same worktree: the vanilla-adapter package's browser tests cannot launch Chromium, and the default test pass flakes 1–5 files under heavy load with 5 s timeouts in store-boot and spawn tests unrelated to this change.

### PR readiness check

- [x] No new package
- [x] No visual testing required — `Chromatic: skip` applied

## Screenshots

Not applicable — measurement constants and documentation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -203,6 +203,12 @@ precedent.
   does not, however important that package is.
 - **If you believe you need a workflow change, say why in the PR body** and
   expect it to be challenged.
+- **One target is deliberately off CI and must stay off it:** the `pragma`
+  CLI's perf-budget pass (`packages/cli/pragma`, `bun run test:perf`). Owner
+  ruling 2026-08-30, restated 2026-09-10 — a wall-clock spawn measurement
+  cannot mean anything on a shared runner. Nothing reaches it today: not
+  `test`, not an Nx target, not a hook. Putting it back on any CI target
+  reverses a ruling; see `docs/CI.md` and `packages/cli/pragma/BUDGETS.md`.
 ### React component props — extend the root element's native props
 
 A React component's props type **must extend the native props of the element it
--- a/docs/CI.md
+++ b/docs/CI.md
@@ -127,6 +127,23 @@ Chromatic workflows use a shared template (`.github/workflows/chromatic._templat
 
 On pull requests, Chromatic requires manual approval for visual changes. On pushes to main, changes are automatically accepted as new baselines. This allows reviewing visual changes during PR review while keeping baselines current after merge.
 
+## What CI deliberately does not run
+
+### The `pragma` perf budgets
+
+`packages/cli/pragma` has a serial performance-budget pass (`bun run test:perf`, `vitest.perf.config.ts`) that spawns the shipped CLI entry and times it against the ceilings in `src/testing/perf/budgets.ts`. **It is not part of CI, by owner ruling of 2026-08-30, restated 2026-09-10, and it must not be added back.**
+
+The reason is not that the budgets do not matter. It is that a wall-clock spawn measurement cannot be made to mean anything on a shared runner: the same ceilings measured green on the reference machine and red on GitHub-hosted runners, and raising one of them moved the failure from the budget to Vitest's per-test timeout rather than clearing it. A gate that fails for the runner's reasons teaches contributors to re-run it, which is worse than no gate.
+
+So the pass is manual. Nothing invokes it automatically — not `pr.yml`, not `push.yml`, not `tag.yml`, no Nx target, no git hook, and not the package's own `test` script (which is why `nx affected -t test` no longer reaches it). Anyone can run it, and it stays enforced when they do.
+
+Two consequences are worth naming rather than discovering:
+
+- **A performance regression can reach `main` unobserved.** That is the accepted cost of the ruling. Where the pass should run instead — a scheduled job, a release gate, or a maintainer's run before a release — is an open question, recorded in `packages/cli/pragma/BUDGETS.md`.
+- **It needs a quiet machine.** Spawn latency inflates 2–3× under other load, so a red run on a busy box is usually measuring the box. `BUDGETS.md` documents the `--version` control that tells the two apart.
+
+If you believe this pass belongs in a workflow, you are reversing an owner ruling: say so explicitly in the PR body, per `AGENTS.md` ("CI workflows are global — do not add checks for one package").
+
 ## How to Release
 
 Releases require write access to the repository and must run from the main branch.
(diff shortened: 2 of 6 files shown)


advl:
## ADR J needs a correction — reported here, not edited

Per the brief, this is reported and **not** fixed: nothing in `canonical/pragma-adrs` was touched. Lines are from `git show origin/docs/session-j-anatomy-tokens:session/J.ANATOMY_TOKENS.md` (PR #13).

### 1. Superseded — the `tag.yml` wiring (two places)

The 2026-09-09 ruling was that `test:perf` does not *gate* the release tag; ADR J read that as licence to add it as a **non-gating job**. The 2026-09-10 instruction ("it should not be part of ci") supersedes that: gating or not, it does not go in a workflow.

**Line 681** (§8.1 prose, after the branch table):

> The owner ruled on 2026-09-09 that `test:perf` does not gate the release tag — "no" — so J-5 wires it into `tag.yml` as a non-gating job that runs `./.github/actions/setup-env` as the `test` job does (the composite builds, `bunx lerna run build:all` by default, which is what lets `test` spawn `dist/src/bin.js` already) and then `bun run --cwd packages/cli/pragma test:perf`

**Line 725** (the J-5 row in §10's work table), final clause:

> The two `bun run --cwd packages/cli/pragma test:perf` measurements of §8.1, recorded in `BUDGETS.md` as a new dated section rather than an edit to the PR7 record, and `test:perf` on `tag.yml` non-gating as ruled.

### 2. Compatible as written — gate GJ-4

**Line 735** asks only for a manual run with the results recorded, which is exactly what the instruction leaves in place, so it needs no change:

> `bun run --cwd packages/cli/pragma test:perf` run and both measurements of §8.1 recorded in `BUDGETS.md` — recorded, not gating, by the owner's "no"

Worth confirming explicitly, since the ADR names both `tag.yml` and GJ-4 in the same breath and only the former is affected.

### 3. Materially affected by this PR — the 130/150 conditional

**Line 756** (§10 decision row) and the branch table at **lines 678–679** hinge on the fast paths holding at 130/150 ms:

> **Ruled** (item 6 of 2026-09-10, keep as written): a raised `BUDGET_WARM_STORE_MS` is a recorded `BUDGETS.md` decision, with its measurement beside it, only while the fast paths hold — `pragma --help` under 130 ms, `pragma __complete` under 150 ms — and otherwise the records leave the snapshot

That conditional has now fired, but **not for the reason it was written to detect.** `--help` exceeded 130 ms on p95 because the eagerly-imported capability barrel grew 37.6 → 68.0 ms with the surface (22 capability modules, 42 MCP tools). ADR J's own reasoning at line 681 already rules the records out as a cause — they live in `pack.generated.ts`, which no fast path imports, and the only fast-path module the programme touches is `pragma.conf.ts`, measured at zero. So taken literally the rule would drop the records from the snapshot for something they did not do.

The ceilings are now 210 and 240, re-derived on the stated rule. If the intent of item 6 was "records stay only while the fast paths are not degraded *by the records*", the conditional should be rewritten against a measured delta rather than against absolute ceilings that move for unrelated reasons. That is the owner's call, not this PR's.

### 4. Partly satisfied already — §8.1's "first act is two measurements"

**Line 673** states:

> `BUDGET_WARM_STORE_MS = 500` was projected from an 8,479-triple graph (`BUDGETS.md`: 332.7 ms projected median), the shipped pack is 49,630 triples and has never been measured against it, and `test:perf` runs in no workflow

The 49 630-triple arm is now measured and recorded in `BUDGETS.md`: median 536.0 ms, p95 608.9 ms, netted store work +499.9 ms, ceiling re-derived to 850. J-5's **baseline** arm (~54 300 triples, token-ontology 0.11.0, anatomy-dsl 0.4.0, design-system pre-J-4) is still its own work, and the ledger it compares against now exists — so J-5 should net against this record rather than re-measuring the shipped arm from scratch.

PR #1218, opened 2026-09-10
feat(cli): answer the token graph — symbols, variables, values, consumers and chains

advl:
## Done

- **Repointed the `token` noun at the token symbols.** Its story queried
  `ds:Token` with a required `ds:tokenId`, and no graph this CLI ships asserts
  either: the shipped pack carries **zero** instances of that class and **745**
  `dt:TokenSymbol`. So `token list` answered an empty table on every install
  while the population sat one namespace over, and `token lookup` and
  `token sample` were unreachable for the same reason — silently, because a
  declaration that maps onto no property is exactly what OPTIONAL parity
  swallows. The noun now addresses one row per logical dotted name, with the
  modifier-channel relation, and with the definitions, coverage and resolved
  values reachable from the lookup.
- **Type and description are read from the symbol's own definition.** They are
  definition-level facts, 393 of the 745 symbols have more than one definition
  and `color.text` has 20 — but the graph already designates one of them: the
  head of the `dt:resolutionChain` on the symbol's base resolved value (the one
  with no `dt:coordinate`, the resolver's default position). The list reads it
  with `FILTER NOT EXISTS { ?resolved dt:coordinate ?c }`; the lookup reads it
  as the path `^dt:forSymbol/dt:resolutionChain/rdf:first/…`. 718 of 745 carry
  both facts; the 27 without are the 25 channels (no definition anywhere) and
  two typography symbols with no resolved value. No aggregation, and no
  upstream change: this answers on the published token-ontology 0.10.1.
- **Added a `variable` noun over the platform stratum.** A symbol is the logical
  name; a variable is what a stylesheet declares, and **236 of the 1,156**
  variables stand for no symbol at all — the duplicated legacy twins the
  ontology names as such, plus the computed states and resets. No surface keyed
  on symbols can reach them. Its published name is the CSS name with the leading
  dashes stripped, which is what makes it usable: a `--`-prefixed positional
  cannot be typed, and the parser answers `unknown option`. Measured safe — 1,156
  variables produce 1,156 distinct names, none colliding with a symbol name.
- **Three list-shaped verbs.** `token values` (the value each symbol resolves to
  at each position, with its resolution chain **or** its derivation — the shape
  admits exactly one, and reading only the chain would blank all 165 channel
  routings); `token consumers` (which block consumes which symbol, at which
  style key, state, rank and node — nameable by the dotted symbol **or** by a CSS
  variable standing for it, because a web implementer holds the latter);
  `variable chain` (the transitive walk from a variable to every symbol it
  reaches, one property path).
- **Edited the frozen surface covenant by hand**: one new noun, six new tools,
  `flags` arrays on the new list bodies, and the tool-count literal moved in all
  four places it is written across the two protected suites. Derived from the
  projector rather than assumed: **43 → 49 tools, 22 → 23 nouns, 50 → 56 CLI
  commands.** Tool hints and the generated reference pages follow.
- **Bumped the two packs that make this answer real data.** `token-ontology`
  0.10.0 → 0.10.1 publishes the `rdfs:label` both nouns key on, on all 745
  symbols and all 1,156 variables; before it, both nouns answered empty by
  design. `anatomy-dsl` 0.3.0 → 0.5.0 moves the `anatomy:` namespace off its
  `example.org` placeholder and ships the style-key registry — 111 individuals
  and the shape that closes the interaction states — which is what gives
  `token consumers` a roster to admit `--key` and `--state` values against. The
  embedded snapshot is rebuilt: 49,630 → 53,467 triples.
- **Re-derived the default page, 500 rows → 300** (breaking, see below). The 500
  was justified only from below — it sat above every population, so nothing was
  truncated. Three of the new populations pass it, and at 500 rows `token values`
  serialised 121,776 bytes: inside the 125,000-byte payload ceiling but past the
  0.8 gate the budget suite enforces, with no headroom. 300 is the largest number
  that stays above the 252 rows of the largest pre-existing population *and*
  leaves the biggest default answer at 61 per cent of the ceiling.
- **Widened one grammar rule**: a declared filter's param may now be camelCase.
  The projector already writes a flag as `--<kebab-case>`, so `channelOf` was
  always going to become `--channel-of`; only the validator refused the one
  spelling that projects well, while admitting `channelof`.
- **Recorded a dated tool-catalogue measurement**, and it is worth a look: the
  figure on record implied ample headroom (2,767 tokens at 38 tools), while the
  real number was 5,434 at 43 tools *before* this work and is 7,320 at 49 after
  it — **91 per cent** of the 8,000-token ceiling. Input schemas outweigh
  descriptions across the catalogue. Every description added here was cut back
  against that measurement; the ceiling itself is left alone, because what to do
  about a limit the next few tools will breach is a decision rather than a
  constant edit.

**Deliberately not included, and marked in the source where it would go:** a
`tokens` expand on `block lookup` — the mirror of `token consumers`, from the
block's end. There is no path from a block to a consumed symbol that the shipped
ontology defines. `ds:hasTokenBinding`, `ds:consumesSymbol`, `ds:rank`,
`ds:viaBlock` and `ds:node` are declared as properties nowhere; the anatomy
route (`Specification → rootNode → Node → hasStyle → Style → consumes`) is
complete in itself but **no property anywhere ranges over
`anatomy:Specification`**, so nothing joins a block to its anatomy graph — a
block carries its anatomy as a string, not as nodes. A declared expand over
undefined terms can never render on any block on any install, which is the exact
defect class `block.shipped.exec.test.ts` exists to refuse, so it waits for the
design system to mint the record. `token consumers` already answers the same
question from the symbol's end and needs no code change when the records arrive.

Also repointed rather than deleted, where this work retired a premise: the
first-install journey asserted `token list` returns no rows "because the graph
carries no `ds:Token`" (it returns 745 now, so the case became one that asserts the
noun resolves offline, that a named symbol is reachable through a filter
whatever page it sorts onto, and that `token consumers` is the honestly-empty
verb it inherited the role from); the empty-corpus allowlist lost its only entry; the budget suite's
whole-population claim became the paging contract that survives it; and the
shared fixtures gained the terms and populations the new nouns read.

**BREAKING CHANGE:** a list-shaped verb called with no `--limit` returns at most
300 rows instead of 500. No answer that fitted one page before needs more pages
now — the largest pre-existing population is 252 — and the number is in `--help`
and the generated reference. `token list` no longer publishes a `category`
column, and `token lookup` no longer publishes `category`, `valueLight` or
`valueDark`: those read the class no shipped graph asserts, so no install can
have been receiving them. Read `type` and the `values` expand instead.

## QA

Run from `packages/cli/pragma`.

- `bun run check` — biome, types, architecture rules, and the embedded-pack
  provenance gate.
- `bun run test` — the whole suite, including the two protected covenant suites
  at their new counts, the two new parity suites, the new shipped-corpus suite,
  the new CLI round trip, and the payload-budget sweep.
- `bun run eval` — 23 cases, including four new ones over the shipped pack.
- Then against the built CLI (`bun run build`, then `node dist/src/bin.js …`):
  - `token lookup color.text` — resolves the dotted name; shows 20 definitions
    apart, seven covering families, and two resolved positions.
  - `token lookup modifier.color.text` — a channel: names the symbol it
    provisions, and every value row carries a derivation and **no** value cell.
  - `variable lookup color-text` — resolves the dash-stripped name; shows the
    selector `:root`, the at-rule `@layer ds.modifiers`, the emitted
    `light-dark(…)` value and the source location.
  - `variable lookup --color-text` — must be **refused** with `unknown option`.
    That refusal is why the published name has no dashes.
  - `token values --position mode.dark --limit 3` — positions read as
    `mode.dark`, not as an IRI.
  - `token consumers` — answers 0 rows, exit code **0**, with a recovery hint
    naming both causes. `token consumers --key nope` is refused and names the
    111 admissible keys; `--key appearance.background` is admitted and answers
    the same calm empty list.
  - `variable chain --variable modifier-color-text` — a transitive walk that
    passes through variables standing for no symbol.
  - `--limit 2` then `--limit 2 --after <cursor>` on each of `token list`,
    `token values`, `variable list`, `variable chain` — the two pages together
    are byte-identical to `--limit 4`.
- And over MCP (`node dist/src/bin.js mcp serve`, JSON-RPC on stdio):
  `tools/list` reports 49 tools including the six new ones; all nine
  `token_*`/`variable_*` tools answer; `token_list {type:"shadow"}` and
  `variable_list {limit:9007199254740991}` are both typed `INVALID_INPUT`.

### PR readiness check

- [ ] If this PR introduces a **new package**: not applicable — no new package.

## Screenshots

Not applicable — this is a CLI and MCP surface with no visual component.
`Chromatic: skip` applies.

---

### Two things found while testing, worth a reviewer's eye

**The resolution-chain cell is a set, not a sequence.** `token values`
aggregates the chain into one cell, and `GROUP_CONCAT` has no defined order in
SPARQL and takes no `ORDER BY` — so the links of a multi-link chain may
serialise either way round. A parity assertion that pinned one spelling flaked
before it was corrected. The cell answers *which* definitions a value passed
through; the ordered anchor a caller usually wants is the chain's head, the
definition the value was authored in, which the lookup's `values` expand
projects on its own. Both the story and the test now say so.

**`token list` used to be the most expensive declared story, and is now an
ordinary one.** An earlier revision computed type and description by an
agreement rule that aggregated all 1,311 definitions before a page could be cut,
so a 50-row page cost the same as all 745 (about 1.1 s a query, and the protected
paged walk timed out at 190 s). Reading the base chain head instead takes the
unpaginated 745-row read to 71 ms and the paged walk to 1.8 s. The lookup path
cannot carry the base restriction, so `tokenGraph.shipped.exec.test.ts` pins the
corpus invariant it relies on: no symbol's chain heads name two types or two
descriptions.


advl:
## The protected suite: three causes found, two fixed, one for a ruling

The whole-corpus suite could not finish. In isolation with nothing else on the machine it took **55 minutes** and timed out on five of 43 cases; CI's build matrix ran **1h45m** before failing. It is now **4m38s with one failure**, reproduced twice.

**Fixed — a real assertion, not a timeout.** The narrowable-body pin still named three stories. Five more are narrowable now: `token consumers`, `token list`, `token values`, `variable chain`, `variable list`. That case failed in 4ms.

**Fixed — the sweeps were infeasible, and this narrows a promise, so it wants your eye.** The parity and calm-empty sweeps ask one query per admissible filter value. Every filter predating this work admitted at most 36 values, so exhausting one cost seconds. Two of the new filters admit **745** symbols and **1,156** variables, and exhausting those two took 22 minutes between them:

| case | before |
| --- | --- |
| calm-empty across all bodies | 26 min |
| `token list` parity sweep | 15 min |
| `variable chain` parity sweep | 7 min |

A choice-sized vocabulary is still swept exhaustively. A catalogue-sized one is swept at a deterministic evenly-spaced sample of the same size, so it stays reproducible across runs and still exercises every shape of value the filter admits. The boundary is the constant that already decides when an error message stops naming values and starts counting them, because it is the same distinction. The suite's header now states the narrower promise instead of overstating it.

**Fixed.** The first-install case boots the embedded pack, which grew to 53,467 triples here, so it takes an explicit budget like every other store-booting case in that file.

### The one that remains, deliberately

`token list: a paged walk is the unpaginated answer, in order` — 129s against a 120s budget. This is not a test problem:

```
pragma token list --limit 7     ->   7 rows in 1112ms
pragma token list --limit 745   -> 745 rows in 1190ms
```

The cost is **per call and independent of page size**, because the agreement rule must aggregate all 1,311 definitions before any page can be cut. The walk pages by 7, so 745 rows is 107 pages at ~1.1s each, which is the 129s exactly. **Every reader paging that list pays the same 1.1s per page.**

I did not raise the budget, because that would hide a cost real callers pay, and I did not enlarge the test's page size, because that would hide it too. The query cannot be fixed portably: a SPARQL sub-select is not correlated, so the aggregation runs in full however the outer query is paged.

The fix is for the graph to publish the agreed pair on the symbol itself, so this story reads a value instead of computing one. That is a modelling decision about the token ontology rather than anything this PR can settle, and it is the same item already raised for `token lookup`, where a property path cannot express the agreement rule at all. Both disappear together.

PR #1225, opened 2026-09-11
fix(react-ds-global): export TooltipEngine and add ContextualMenu surfaceClassName

potentialwook:
## Done

Two supporting fixes for the SideNavigation rebuild stack (the following PRs in this series consume both):

- **Export `TooltipEngine`** from `@canonical/react-ds-global`. `withTooltip` fixes its `Message` at wrap time, so a component whose tooltip must follow its own prop state (e.g. a toggle naming the action for each state) had to pick between two wrapped component types — remounting the target on every state flip and dropping focus and hover state. Such consumers now render `TooltipEngine` directly with `Message` as a live prop; the trade-off is documented on `withTooltip`.
- **Add `surfaceClassName` to `ContextualMenu`** (drive-by in the same package): the menu's popup surface portals to `document.body`, so the root's `className` cannot reach it — this is the one hook for theming the surface itself. Covered by ContextualMenu tests.

## QA

- `bun run check` and `bun run test` pass from the repo root (66 projects).
- `TooltipEngine` tests: message reactive to rerender; target element identity stable across prop changes.

### PR readiness check

- [ ] If this PR introduces a **new package**: first-time publish has been done manually from inside the package directory with `npm publish --access public` (first-time publishing is not automated). Run `bun run publish:status` from the repo root to verify. Then configure an OIDC [trusted publisher](https://docs.npmjs.com/trusted-publishers) for the package on npmjs.com (repo `canonical/pragma`, workflow `tag.yml`) and set publishing access to disallow tokens, so the automated release workflow can publish future versions.

### DIFF
--- a/packages/react/ds-global/src/lib/component/ContextualMenu/ContextualMenu.tsx
+++ b/packages/react/ds-global/src/lib/component/ContextualMenu/ContextualMenu.tsx
@@ -29,6 +29,7 @@ const ContextualMenu = ({
   items,
   label,
   className,
+  surfaceClassName,
   preferredDirections,
   distance,
   gutter,
@@ -144,6 +145,7 @@ const ContextualMenu = ({
         "contextual-menu__surface",
         "modal",
         bestPosition?.positionName,
+        surfaceClassName,
       ]
         .filter(Boolean)
         .join(" ")}
--- a/packages/react/ds-global/src/lib/component/Tooltip/withTooltip.tsx
+++ b/packages/react/ds-global/src/lib/component/Tooltip/withTooltip.tsx
@@ -21,7 +21,7 @@ const componentCssClassName = "ds tooltip-area";
  * is the shared implementation that {@link withTooltip} builds on — it renders
  * the wrapped component as its `children` target.
  */
-const TooltipEngine = ({
+export const TooltipEngine = ({
   children,
   style,
   className,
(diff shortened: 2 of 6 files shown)


packages/react/ds-global/src/lib/component/Tooltip/withTooltip.tsx
```diff
@@ -21,7 +21,7 @@ const componentCssClassName = "ds tooltip-area";
  * is the shared implementation that {@link withTooltip} builds on — it renders
  * the wrapped component as its `children` target.
  */
-const TooltipEngine = ({
+export const TooltipEngine = ({
```
advl, line 24: style : This export is forbidden by the code standards (only default exports unless exceptions)

Either split into a new file or remove export

advl (changes requested):
Review the Modal component in particular the way the HOC and the custom consumption pattern share the public api

packages/react/ds-global/src/lib/component/ContextualMenu/ContextualMenu.tsx
```diff
@@ -144,6 +145,7 @@ const ContextualMenu = ({
         "contextual-menu__surface",
         "modal",
         bestPosition?.positionName,
+        surfaceClassName,
```
advl, line 148: Cleanup the approach to the surface
If you believe we dont want a modal surface, change it L146

potentialwook: Modal surface and Modal component share the same class name.
We will remove the modal class from here.
The surface class name also appears to be redundant so we can remove that too.

packages/react/ds-global/src/lib/component/Tooltip/withTooltip.tsx
potentialwook: We will move this to a different file and we can discuss together discouraging HOCs in the future

potentialwook: Resolved up-stack in #1285 (merge last): TooltipEngine now lives in its own file (`TooltipEngine.tsx`) with a single default export; `withTooltip.tsx` keeps only the HOC and consumes the engine like any other caller. The barrel re-exports both uniformly, and the engine's tests moved with it into `TooltipEngine.tests.tsx`.

packages/react/ds-global/src/lib/component/ContextualMenu/ContextualMenu.tsx
potentialwook: Resolved up-stack in #1285: the `modal` class is gone from both the root surface and the submenus — it was reading Modal's surface channel rather than the menu's own mapped token. `surfaceClassName` is removed with its test, and the surface paints `--color-background` directly.

advl (approved):
Comment for the future : implement fixes in the same PR the issues are raised

PR #1226, opened 2026-09-11
docs(ds-app): specify SideNavigation against the 24.04 spec

potentialwook:
## Done

Specify SideNavigation against the 24.04 design spec, ahead of the component rebuild that follows in this stack:

- `SPEC.md` — region semantics (Header, optional ContextSwitcher region, Content landmark, Footer), the data-driven consumption pattern, properties, rail state, keyboard behaviour, accessibility, and the deliberate exclusions.
- 11 `*.anatomy.yaml` files — the component and each subcomponent's anatomy in the DSL.

No runtime change. The rebuild PR in this stack implements what this specifies.

## QA

- `bun run check` and `bun run test` pass from the repo root (66 projects); the anatomy files parse under the package's anatomy tooling.

### PR readiness check

- [x] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format, using one of the allowed types: `feat`, `fix`, `docs`, `refactor`, `chore`, `test`, `ci`, `revert`.
  - The matching type label is applied automatically from the title — there is no type label to add by hand.
  - Breaking changes are marked with `!` in the title (e.g. `feat(router)!: …`), which adds the `breaking` label.
- [x] The code follows the appropriate [code standards](https://github.com/canonical/web-code-standards)
- [x] All packages define the required scripts in `package.json`:
  - [x] All packages: `check`, `check:fix`, and `test`.
  - [x] Packages with build steps: `build` to build the package for development or distribution, `build:all` to build **all** artifacts. See [CONTRIBUTING.md](../old/CONTRIBUTING.md#24-full-artifact-builds-buildall) for details.
- [ ] If this PR introduces a **new package**: first-time publish has been done manually from inside the package directory with `npm publish --access public` (first-time publishing is not automated). Run `bun run publish:status` from the repo root to verify. Then configure an OIDC [trusted publisher](https://docs.npmjs.com/trusted-publishers) for the package on npmjs.com (repo `canonical/pragma`, workflow `tag.yml`) and set publishing access to disallow tokens, so the automated release workflow can publish future versions.
- [x] If this PR does not require visual testing, add the `Chromatic: skip` label to skip Chromatic.

PR #1227, opened 2026-09-11
feat(ds-app)!: rebuild SideNavigation into a single data-driven region shell

potentialwook:
## Done

Full rebuild of `SideNavigation` as a single data-driven region shell, per the spec landed earlier in this stack. **Breaking** — the public API changes shape.

- **One navigation landmark.** The root is a plain `<div>`; the single `<nav>` landmark is Content's, `aria-label` forwarded, defaulting to `"Main navigation"`. Branding, the context switcher and footer actions sit outside the landmark by design.
- **Skip navigation.** A visually hidden "Skip to main content" link is the first focusable element (target via `skipTo`, default `#main-content`).
- **Four regions, all data-driven:** Header (branding, collapse toggle), optional ContextSwitcher region (typed props slot, plain `<div>` outside the landmark, hidden when collapsed), Content (links, plain labels, expandable disclosures only), Footer (free-form rows via `footerRoot` — leaves and depth-1 expandables; buttons via `control: "button"`).
- **No JSX escape hatch.** No `children` props; every subcomponent is private and renders exclusively from data.
- **Collapsed rail** keyed off `data-expanded` on the root: Content and the switcher region hide, the footer degrades to icon-only, and expandables disclose into a floating inline-end popover (CSS anchor positioning + `position-try-fallbacks` where supported, top-anchored otherwise; native `<details>` throughout).
- **Auto-expand on active:** a nested sub-item whose url becomes active re-opens its collapsed parent (one-way — manual collapse is respected).
- `Ctrl+B` rail-collapse shortcut ships off by default in this commit (`keyboardShortcut`, reserved); the follow-up PR in this stack turns it on.
- Fixtures live in `storybook/navigation/fixtures.tsx`; stories are data-driven (MAAS, LXD, `with ContextSwitcher`, `Collapsed`, `Mobile`, `CertificateUser`).

Based on #1225 (`TooltipEngine` export — the rebuild imports it), #1226 (the spec it implements) and #1234 (the subcomponent primitives it consumes); merge after all three.

Split for reviewability: the five private subcomponents (`ItemButton`, `ItemExpandable`, `Group`, `GroupHeader`, `ContextSwitcher`) and the `useCollapseShortcut` hook landed separately in #1234; this PR is the wire-up — the breaking swap to the data-driven region shell.

## QA

- `bun run check` and `bun run test` pass from the repo root; 162 ds-app tests.
- Storybook: MAAS/LXD data-driven; `with ContextSwitcher` routes url'd contexts through the hash router and fires the create-context action; `Collapsed` — content + switcher hidden, footer icon-only, footer expandables open as inline-end popovers; routing to a nested sub-item auto-opens its parent with the sub-item highlighted.
- Keyboard: Tab → first stop is "Skip to main content" → Enter jumps past the navigation.

### PR readiness check

- [ ] If this PR introduces a **new package**: first-time publish has been done manually from inside the package directory with `npm publish --access public` (first-time publishing is not automated). Run `bun run publish:status` from the repo root to verify. Then configure an OIDC [trusted publisher](https://docs.npmjs.com/trusted-publishers) for the package on npmjs.com (repo `canonical/pragma`, workflow `tag.yml`) and set publishing access to disallow tokens, so the automated release workflow can publish future versions.


packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/styles.css
```diff
@@ -1,8 +1,53 @@
-/* Generated by @canonical/summon-component:react v0.1.0
- * Layer: ds.components.apps (@canonical/styles README, "Cascade layers"). */
+/* Generated by @canonical/summon-component:react v0.1.0 */
 
-@layer ds.components.apps {
-  .ds.collapse-toggle {
-    /* Component styles */
+.ds.collapse-toggle {
```
advl, line 3: needs the layer

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/styles.css
```diff
@@ -1,8 +1,53 @@
...
+  }
+
+  /* Icon (desktop) vs "Menu"/"Close menu" text (mobile) — plain CSS-toggled
+     siblings, not a JS media-query read, so the choice needs no client-only
+     branch and stays SSR-identical. */
+  .mobile-only {
+    display: none;
+  }
+
+  /* Vanilla's "small" breakpoint (SPEC.md §7: "on all breakpoints above the
```
advl, line 36: Please fix comments - they should not quote spec. Use agents.md

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/styles.css
```diff
@@ -1,8 +1,53 @@
...
+     small breakpoint of Vanilla" / "at breakpoints below small"). */
+  @media (max-width: 767px) {
+    .desktop-only {
```
advl, line 39: Avoid utility classes - I would reocmmend to select directly the items that have this property

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/styles.css
potentialwook: Already resolved up-stack: #1228 and #1229 restored the `@layer ds.components.apps` wrapper and its header comment to every SideNavigation stylesheet — verified all 12 at the tip of the stack. Nothing further needed; tracked in #1285.

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/styles.css
potentialwook: Resolved in #1285 (stacked on #1229): every spec-quoting comment across SideNavigation is rewritten to carry the decision and its rationale in the codebase's own words — no quoted spec text, no §N citations, no ADR-Txx references; test titles drop their spec parentheticals too. Deliberate-deviation records keep naming the upstream 24.04 design's stated value so the shipped numbers stay guarded.

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/styles.css
potentialwook: Resolved in #1285 (stacked on #1229): the collapse toggle drops `.desktop-only`/`.mobile-only` — the button's only two children are the icon svg and the label span, so the stylesheet selects them directly. Rendered output is unchanged on either side of the small breakpoint.

packages/react/ds-app/src/storybook/navigation/story-utils.tsx
```diff
@@ -97,32 +98,44 @@ export const withSideNavShell: Decorator = (Story) => (
...
- * Imposes a page-like grid so the nav is shown in a realistic context: the nav
- * sits in the start column (responsive — full width under ~300px, fixed 300px
- * above) with a placeholder main-content column filling the rest.
+ * Imposes a page-like grid so the nav sits in a realistic context: a start
+ * column matching the rail's own 240px width (SideNavigation sets its own
+ * `inline-size` regardless; the matching column just avoids a dead gap)
+ * with a placeholder main-content column filling the rest.
  */
 export const withNavLayout: Decorator = (Story) => (
-  <div
```
advl, line 105: Avoid shorthand names -> `withNavigationLayout`  or similar, let's reduce the cognitive load of users wondering how things are named

packages/react/ds-app/src/storybook/navigation/CanonicalLogo/CanonicalLogo.stories.tsx
advl, line 1: This file / story looks misplaced. Is is a component part of the library or a helper ? Atm it lives in the story helpers folder and shouldn't be part of any public api.

packages/react/ds-app/src/lib/SideNavigation/SideNavigation.tsx
```diff
@@ -40,6 +64,10 @@ const SideNavigation = ({
     setExpanded((current) => !current);
   }, []);
 
+  // Reserved (§10.1) — inert until `keyboardShortcut` opts in.
```
advl, line 67: cleanup comments

packages/react/ds-app/src/lib/SideNavigation/SideNavigation.tsx
```diff
@@ -12,24 +13,47 @@ const componentCssClassName = "ds side-navigation";
...
+ * and render exclusively from data.
+ *
  * @implements ds:apps.pattern.side-navigation
  */
 const SideNavigation = ({
   className,
   brand,
   applicationName,
   root,
+  contextSwitcher,
```
advl, line 46: This is an unidiomatic pattern. The name of the prop doesn't indicate clearly what a consumer should expect. Also, as this happens to be a prop object passed to the contextswitcher might also itself feel like a fragile prop drilling pattern. While I am not fully opposed to this - I'd like you to reflect more on whether this is the right intended consumption api.

packages/react/ds-app/src/lib/SideNavigation/SideNavigation.tsx
```diff
@@ -55,32 +83,47 @@ const SideNavigation = ({
...
+        // Own plain <div> region — a select-like widget is not navigation,
+        // so it sits outside the <nav> landmark. Hidden when collapsed:
+        // its rows are label-driven, unviable icon-only.
+        <div
+          className="ds side-navigation-context-switcher-region"
+          data-expanded={expanded}
+        >
+          <ContextSwitcher {...contextSwitcher} />
+        </div>
+      ) : null}
```
advl, line 112: It feels a little bit strange that while all other subcomponents are directly called, this one has a hard-coded container. I'd recommend making it a full section of its own right for consistency.

packages/react/ds-app/src/lib/SideNavigation/SideNavigation.tsx
```diff
@@ -55,32 +83,47 @@ const SideNavigation = ({
...
-        .join(" ")}
+      className={[componentCssClassName, className].filter(Boolean).join(" ")}
       data-expanded={expanded}
       {...props}
     >
+      {/* Skip navigation, first in DOM order — visually hidden until focused,
+          then jumps keyboard users past the navigation block. */}
+      <a className="skip-link" href={skipTo}>
+        Skip to main content
+      </a>
```
advl, line 94: I believe we have or have planned to have a skiplink component in the global tier. 

It can be added as a follow-up. I wonder whether the mcp can catch this ?

packages/react/ds-app/src/lib/SideNavigation/styles.css
```diff
@@ -1,28 +1,266 @@
...
+
+  /* The nav spans the full viewport height ("height of the navigation is
+     100% of the viewport"). Content (flex:1, min-height:0) absorbs the
+     remainder and scrolls internally so Footer stays pinned to the bottom. */
+  block-size: 100%;
+
+  /* Per-component override of the shared overflow-fade token (default 2rem
+     in @canonical/styles/overflow.css): the spec calls for 1.75rem
+     specifically at the Header/Content and Content/Footer seams. */
+  --overflow-gradient-height: 1.75rem;
```
advl, line 77: you can upstream this change directly onto the styles :)

packages/react/ds-app/src/lib/SideNavigation/styles.css
```diff
@@ -1,28 +1,266 @@
...
+      the transition above. */
+
+  transition:
+    width var(--motion-duration-medium) ease-in-out,
+    inline-size var(--motion-duration-medium) ease-in-out;
+
+  /* Skip navigation — visually hidden until keyboard focus, then it
+     overlays the rail so keyboard users can bypass the block. */
+
+  & .skip-link {
```
advl, line 92: Would likely become its own component in the global tier. Look at the svelte example

packages/react/ds-app/src/lib/SideNavigation/styles.css
```diff
@@ -1,28 +1,266 @@
...
+       icon-only degradation. */
+    & > .ds.side-navigation-context-switcher-region {
+      display: none;
+    }
+
+    /* Footer stays visible, icon-only — spec: "the collapsed state footer
+       items show their labels in tooltips when hovered after 800ms delay."
+       The tooltip itself is a documented follow-up (SPEC.md §10.14); this is
+       the icon-only layout half. */
+    & > .ds.footer {
```
advl, line 139: This is correct but one character away from css conflicts - without the ">" it could potentially select footers in user provided components. I would recommend improving the specificity of the classnames for navigation subcomponents. I wonder if we have a standard for that already ?

packages/react/ds-app/src/lib/SideNavigation/styles.css
```diff
@@ -1,28 +1,266 @@
...
+      .label,
+      .end {
+        display: none;
+      }
+
+      .row {
+        grid-template-columns: var(--sidenav-icon-column-inline-size);
+      }
```
advl, line 147: use direct children selectors (`>`) for subelements, preferably

packages/react/ds-app/src/lib/SideNavigation/styles.css
```diff
@@ -1,28 +1,266 @@
...
+    }
```
advl, line 148: In general -  this applies to all the subcomponents - I am unsure I understand what is the separation of concerns between styles in the subcomponent folders and styles here. The fact that there are two files to describe sucomponent stying makes it confusing and hard to debug. I would recommend to pick one approach and stick with it. Either all in root styles or in subcomponents styles.

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/styles.css
```diff
@@ -1,8 +1,53 @@
-/* Generated by @canonical/summon-component:react v0.1.0
- * Layer: ds.components.apps (@canonical/styles README, "Cascade layers"). */
+/* Generated by @canonical/summon-component:react v0.1.0 */
 
-@layer ds.components.apps {
-  .ds.collapse-toggle {
-    /* Component styles */
```
advl, line 6: I don't understand why this correct code had to be removed. The `@layer` is the new vetted pattern.

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/styles.css
```diff
@@ -1,8 +1,53 @@
...
+.ds.collapse-toggle {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  /* Icon-only control: square hit target roughly matching the leading icon
+     column, no explicit gap/label. */
+  inline-size: var(--dimension-250); /* 1.25rem */
+  block-size: var(--dimension-250);
```
advl, line 10: Following the styling code standards, variables should be declared in `:root` rather than hard coded

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/CollapseToggle.tsx
```diff
@@ -1,37 +1,71 @@
...
         ariaLabel ?? (expanded ? "Collapse navigation" : "Expand navigation")
       }
       {...props}
+      type="button"
     >
-      <Icon icon={expanded ? "collapse-side-nav" : "expand-side-nav"} />
+      {/* Desktop: an icon (SPEC.md §5, §9). Below the small breakpoint
+          (<768px — SPEC.md §7), a text label reading "Menu"/"Close menu"
+          replaces it — CSS-toggled siblings (see styles.css), no
+          media-query read in JS, so it stays SSR-identical. */}
```
advl, line 27: Review comments

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/CollapseToggle.tsx
```diff
@@ -1,37 +1,71 @@
...
 
+// Spec: hovering the collapse button for 1s shows a tooltip reading
+// "Collapse"/"Expand" by state. withTooltip's Message is fixed at wrap
+// time, so two stable wrapped components are built once at module level
+// (inside the render, recreating the component type would remount it,
+// losing focus/hover state) and CollapseToggle picks between them.
+const CollapseToggleWithCollapseTooltip = withTooltip(
+  CollapseToggleButton,
+  "Collapse",
+  { activateDelay: 1000 },
```
advl, line 45: I'd recommend keeping this in a constant at the top of the file and reviewing with designers what the right value is.

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/CollapseToggle.tsx
```diff
@@ -1,37 +1,71 @@
...
+ * SideNavigation.CollapseToggle — icon-only button that expands or collapses
+ * the navigation rail. Carries the disclosure ARIA contract: `aria-expanded`
+ * reflects the current state and `aria-controls` should point at the id of the
+ * navigation region it toggles. Hovering for 1s shows a tooltip naming the
+ * action ("Collapse"/"Expand" — SPEC.md §5, §9.4); the `<button>`'s own
+ * `aria-label` carries the fuller "Collapse/Expand navigation" text.
+ *
+ * @implements ds:apps.subcomponent.side-navigation-collapse-toggle
+ */
+const CollapseToggle = (props: CollapseToggleProps): React.ReactElement => {
```
advl, line 63: this should use the same pattern as the other components, in particular the prop extraction. I also wonder if there is a need for the helpers above since they seem to only e be used once.

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/styles.css
```diff
@@ -1,8 +1,53 @@
...
+  &:active {
+    background: var(--sidenav-row-background-active);
+  }
+
+  /* Icon (desktop) vs "Menu"/"Close menu" text (mobile) — plain CSS-toggled
+     siblings, not a JS media-query read, so the choice needs no client-only
+     branch and stays SSR-identical. */
+  .mobile-only {
+    display: none;
+  }
```
advl, line 34: Sounds good - we'd likely like to follow-up with a code standard to handle responsiveness. This is one of the first components that requires responsive behavior to be directly baked in. Please get in touch when you get there to discuss together recommended paths to move forward.

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/CollapseToggle.tsx
```diff
@@ -1,37 +1,71 @@
...
+const CollapseToggleWithCollapseTooltip = withTooltip(
+  CollapseToggleButton,
+  "Collapse",
+  { activateDelay: 1000 },
+);
+const CollapseToggleWithExpandTooltip = withTooltip(
+  CollapseToggleButton,
+  "Expand",
+  { activateDelay: 1000 },
+);
```
advl, line 51: Might be worth checking with @stavropouloskonstantinoss whether this is the optimial consumption pattern since the contents of the tooltip are variable

packages/react/ds-app/src/lib/SideNavigation/common/Content/styles.css
```diff
@@ -1,36 +1,60 @@
...
-    /* min-height:0 lets this flex item shrink below its content height so it can
-       scroll internally (flex items default to min-height:auto, which refuses to
-       shrink and would push the whole nav taller instead). Vertical scroll only;
-       never horizontal (long labels must not cause sideways scroll). */
+    display: flex;
+    flex-direction: column;
+    gap: var(--dimension-100);
     min-height: 0;
     overflow: hidden auto;
+    container-type: scroll-state;
```
advl, line 12: (Side comment, might be worth doing a small show and tell about the patterns you've used for instance in the next State of Web conversation)

packages/react/ds-app/src/lib/SideNavigation/common/Content/types.ts
```diff
@@ -1,17 +1,14 @@
-import type { Item as NavItem } from "@canonical/ds-types";
-import type { ComponentProps, ComponentType, ReactNode } from "react";
-import type { LinkComponentProps } from "../../types.js";
+import type { ComponentProps, ComponentType } from "react";
+import type { LinkComponentProps, NavRoot } from "../../types.js";
 
 type OwnProps = {
-  /** Root item whose direct children are rendered as the nav item list. */
-  root?: NavItem;
+  /** Root whose direct children render as groups — see NavRoot. Data-only: rows render exclusively from this tree (there is no children escape hatch into the landmark). */
```
advl, line 5: Confusing and unnecessary comment

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.stories.tsx
```diff
@@ -0,0 +1,41 @@
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import { fn } from "storybook/test";
+import { lxdProjectContexts } from "../../../../storybook/navigation/fixtures.js";
```
advl, line 3: `import * as fixtures from ...`

This enables us to keep the fixtures into a declared namespace which makes reading simpler. Iy should be described in a code standard if not already.

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+    title: "Context",
+  },
+};
+
+/** With a "create context" action at the bottom of the list. */
+export const WithCreateContext: Story = {
+  args: {
+    onCreateContext: fn(),
+  },
+};
```
advl, line 41: While those two stories are relevant, it would be great if you had the chance to think more of them from a user perspective. Are we doing as much as we can to ensure stories will illustrate valid use cases with plausible data?

packages/react/ds-app/src/lib/SideNavigation/common/Footer/Footer.stories.tsx
```diff
@@ -1,5 +1,8 @@
 import type { Meta, StoryObj } from "@storybook/react-vite";
-import { maasFooterRoot } from "../../../../storybook/navigation/fixtures.js";
+import {
+  contextSwitcherFooterRoot,
+  maasFooterRoot,
+} from "../../../../storybook/navigation/fixtures.js";
```
advl, line 5: `import * as fixtures from`

packages/react/ds-app/src/lib/SideNavigation/common/Footer/Footer.stories.tsx
```diff
@@ -22,9 +25,16 @@ const meta: Meta<typeof Footer> = {
...
   args: {
     root: maasFooterRoot,
   },
 };
+
+/** Free-form footer rows via the footer root: user link, expandable theme switcher, action button. */
+export const Items: Story = {
+  args: {
+    root: contextSwitcherFooterRoot,
+  },
```
advl, line 39: I am unsure what this story aims to show me - what makes is necessary. Please make it more explicit

packages/react/ds-app/src/lib/SideNavigation/common/Footer/Footer.tsx
```diff
@@ -1,14 +1,64 @@
...
- * `root` (independent of Content). Falls back to `children` when no `root`.
+ * Renders one footer leaf row. The Footer is the one place a navigation row
+ * may be a button. A navigable item (`url` set, `control` not `"button"`)
+ * dispatches to `Item` (a link via `LinkComponent`, active when current);
+ * anything else renders as `ItemButton`. Explicit `control` always wins
+ * over the `url`-presence inference, so an action item may carry a `url`
+ * yet render as a button. Toggle-style rows compose `ItemButton` plus a
+ * `slot` — there is no dedicated switch.
+ */
+const renderFooterItem = (
```
advl, line 20: components should be capitalized

packages/react/ds-app/src/lib/SideNavigation/common/Footer/Footer.tsx
```diff
@@ -1,14 +1,64 @@
...
+  item: LeafFooterItem,
+  currentUrl: string | undefined,
+  LinkComponent: FooterProps["LinkComponent"],
```
advl, line 23: we want to use the regular props object, otherwise this creates a confusing precedent that components can be instantiated with args instead of a dict of props.

packages/react/ds-app/src/lib/SideNavigation/common/Footer/Footer.tsx
```diff
@@ -1,14 +1,64 @@
...
+): React.ReactElement => {
+  const row = item.url !== undefined && item.url === currentUrl;
+
+  if (item.control !== "button" && item.url) {
```
advl, line 27: Have you thought of using a switch ? I'd recommend constructing a props object that you'd then complement for switch cases.

packages/react/ds-app/src/lib/SideNavigation/common/Footer/Footer.tsx
```diff
@@ -17,24 +67,41 @@ const Footer = ({
...
+                  (child) => child.url === currentUrl,
+                )}
+              >
+                {entry.items.map((child) =>
+                  renderFooterItem(child, currentUrl, LinkComponent),
+                )}
+              </ItemExpandable>
+            ) : (
+              renderFooterItem(entry, currentUrl, LinkComponent)
+            ),
```
advl, line 100: This needs to be retaken. We should use stronger, more explicit logic to dispatch to a particular renderer. In addition, alternative renderers should be ideally the same type.

packages/react/ds-app/src/lib/SideNavigation/common/Footer/styles.css
```diff
@@ -1,9 +1,18 @@
-/* Generated by @canonical/summon-component:react v0.1.0
- * Layer: ds.components.apps (@canonical/styles README, "Cascade layers"). */
+/* Generated by @canonical/summon-component:react v0.1.0 */
 
-@layer ds.components.apps {
-  .ds.footer {
-    /* Tree structure (.group/.header/.list) is owned by NavTree's own styles;
-       the footer renders a NavTree and adds no tree-specific layout of its own. */
```
advl, line 7: Please be more careful about what you remove. While this looks unintentional it might give the impression the generated code was not intentionally checked for regressions.

packages/react/ds-app/src/lib/SideNavigation/common/Item/Item.stories.tsx
```diff
@@ -35,15 +35,15 @@ type Story = StoryObj<typeof Item>;
 export const Link: Story = {
   args: {
     url: "/machines",
-    label: "Machines",
+    children: "Machines",
```
advl, line 38: What is the reason behind the change ? Do we recommend to stop using the label altogether ? In which case should we use one or the other ?

packages/react/ds-app/src/storybook/navigation/CanonicalLogo/CanonicalLogo.stories.tsx
advl: Please check with design the favoured approach

packages/react/ds-app/src/storybook/navigation/fixtures.tsx
```diff
@@ -79,32 +80,33 @@ export const maasContentRoot: NavItem = {
         },
       ],
     },
+    // The mandatory collapsible "Help" item, with an external link to legal
+    // information (the 24.04 spec §1.1) — an unlabelled trailing group, the common
+    // placement for it.
+    { key: "help-group", items: [createHelpItem("/legal")] },
   ],
```
advl, line 87: I dont understand the role of this. It seems like a hard-coded flake

packages/react/ds-app/src/storybook/navigation/fixtures.tsx
```diff
@@ -178,37 +180,159 @@ export const lxdContentRoot: NavItem = {
         { url: "/ui/settings", label: "Settings", icon: "settings" },
       ],
     },
+    // The mandatory collapsible "Help" item, with an external link to legal
+    // information (the 24.04 spec §1.1) — an unlabelled trailing group.
+    {
+      key: "help-group",
+      items: [createHelpItem("/legal")],
```
advl, line 187: same

packages/react/ds-app/src/storybook/navigation/fixtures.tsx
```diff
@@ -178,37 +180,159 @@ export const lxdContentRoot: NavItem = {
...
+    name: "default",
+    url: "/ui/project/default/instances",
+    description: "The default project",
+  },
+  {
+    key: "staging",
+    name: "staging",
+    url: "/ui/project/staging/instances",
+    description: "Pre-production environment",
+    badge: badge(12),
```
advl, line 320: Components should be capitalized. We should refrain from using this sort of exotic helpers that require context to understand what happens. Use simply `() => <Badge>12</Badge>` or similar

packages/react/ds-app/src/storybook/navigation/story-utils.tsx
```diff
@@ -97,32 +98,44 @@ export const withSideNavShell: Decorator = (Story) => (
...
 /**
- * Imposes a page-like grid so the nav is shown in a realistic context: the nav
- * sits in the start column (responsive — full width under ~300px, fixed 300px
- * above) with a placeholder main-content column filling the rest.
+ * Imposes a page-like grid so the nav sits in a realistic context: a start
+ * column matching the rail's own 240px width (SideNavigation sets its own
+ * `inline-size` regardless; the matching column just avoids a dead gap)
+ * with a placeholder main-content column filling the rest.
  */
 export const withNavLayout: Decorator = (Story) => (
```
advl, line 106: rename to withApplicationShellLayout - the domain here is the app, and the navigation is part of it. I wonder if you could use the ApplicationLayout block directly as well. This can be a follow-up

packages/react/ds-app/src/storybook/navigation/story-utils.tsx
```diff
@@ -141,7 +154,7 @@ export const MockBadge = ({ children }: { children: ReactNode }): ReactNode => (
       borderRadius: "0.625rem",
       fontSize: "0.75rem",
       lineHeight: 1.4,
-      background: "rgb(0 0 0 / 0.25)",
+      background: "var(--color-icon-warning-disabled)",
```
advl, line 157: ok for now - as a follow-up, when we have the right badge, we should replace.

packages/react/ds-app/src/lib/SideNavigation/common/CollapseToggle/CollapseToggle.tsx
```diff
@@ -1,37 +1,71 @@
...
-/**
- * SideNavigation.CollapseToggle — icon-only button that expands or collapses
- * the navigation rail. Carries the disclosure ARIA contract: `aria-expanded`
- * reflects the current state and `aria-controls` should point at the id of the
- * navigation region it toggles.
- *
- * @implements ds:apps.subcomponent.side-navigation-collapse-toggle
- */
-const CollapseToggle = ({
+const CollapseToggleButton = ({
```
advl, line 8: Ideally just CollapseToggle - if you use "Button" as a suffix it might indicate that we are inheriting the `Button` component which is rightly not the case here. (This is an ontology rule we might need to amend for this kind of cases). Look at the component differentiation model in the concepts table.

advl (approved):
Please split this PR.

packages/react/ds-app/src/storybook/navigation/CanonicalLogo/CanonicalLogo.stories.tsx
potentialwook: resolving this to merge will add to follow up

PR #1234, opened 2026-09-11
feat(ds-app): add SideNavigation's data-driven subcomponents

potentialwook:
## Done

The five private render primitives the rebuilt `SideNavigation` consumes, plus the hook that drives its rail-collapse shortcut — landed ahead of the rebuild (#1227) so that PR stays reviewable:

- **`ItemButton`** — a navigation row that performs an action instead of navigating ("Log out", "Create context").
- **`ItemExpandable`** — an expandable disclosure row: native `<details>`, and in the collapsed rail its sub-items disclose into a floating inline-end popover.
- **`Group`** / **`GroupHeader`** — grouped content with a semantic divider header.
- **`ContextSwitcher`** — the optional region's dropdown: `ContextualMenu`-backed trigger with menu-button semantics, title rendered via `GroupHeader`, typed props slot.
- **`useCollapseShortcut`** — the Ctrl+B rail-collapse shortcut hook (off by default; the follow-up PR in this stack ratifies it on).

All are self-contained — types import only react, `@canonical/ds-assets` and `@canonical/react-ds-global` — and none is wired into the component or publicly exported yet: #1227 consumes them. Every component ships with unit and SSR tests, and the styles already carry the stack's final cascade-layer and token truth.

Based on #1226, the spec these implement. No runtime or visual change at this boundary — the current `SideNavigation` is untouched.

## QA

- `bun run check` and `bun run test` pass at this stack boundary. Playwright browser suites (svelte packages, vanilla-adapter) were not run locally — Playwright is non-functional on this machine; they don't consume `react-ds-app`, and CI runs them.
- No visual diff: nothing renders these components yet — `Chromatic: skip` applies.

### PR readiness check

- [ ] If this PR introduces a **new package**: first-time publish has been done manually from inside the package directory with `npm publish --access public` (first-time publishing is not automated). Run `bun run publish:status` from the root to verify. Then configure an OIDC [trusted publisher](https://docs.npmjs.com/trusted-publishers) for the package on npmjs.com (repo `canonical/pragma`, workflow `tag.yml`) and set publishing access to disallow tokens, so the automated release workflow can publish future versions.


packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
```diff
@@ -0,0 +1,153 @@
+import type { MenuEntry, MenuItem } from "@canonical/react-ds-global";
+import { ContextualMenu, Icon } from "@canonical/react-ds-global";
+import type React from "react";
+import { GroupHeader } from "../GroupHeader/index.js";
+import type { ContextSwitcherItem, ContextSwitcherProps } from "./types.js";
+import "./styles.css";
+
+const componentCssClassName = "ds side-navigation-context-switcher";
+const surfaceCssClassName = "ds side-navigation-context-switcher-surface";
```
advl, line 9: surfaces exist as part of a set, custom surfaces should not exist

packages/react/ds-app/vitest.setup.ts
```diff
@@ -1,15 +1,23 @@
...
-  observe: vitest.fn(),
-  unobserve: vitest.fn(),
-  disconnect: vitest.fn(),
-}));
+// A class, not `vitest.fn().mockImplementation(arrowFn)`: an arrow function
+// cannot be invoked with `new`, and components that construct a real
+// `new ResizeObserver(callback)` (e.g. anything using
+// @canonical/react-ds-global's window-fitment positioning, like Popover)
+// throw "is not a constructor" against the previous mock. Matches the
+// working pattern already used in @canonical/react-ds-global's own setup.
```
advl, line 16: nit : doesn't add much

packages/react/ds-app/src/lib/SideNavigation/common/hooks/useCollapseShortcut/index.ts
advl, line 1: nit : hooks can be one folder up, SideNavigation/hooks/useCollapse... or directly in the subcomponent folder (DDD)

packages/react/ds-app/src/lib/SideNavigation/common/ItemExpandable/ItemExpandable.tsx
```diff
@@ -0,0 +1,136 @@
...
+ * `cs:ui_blocks.nojs.disclosure`. `heading` and `children` are composed,
+ * not string props — matching `Accordion.Item`'s own split.
+ *
+ * Uncontrolled — seeded by `defaultExpanded` (NavTree supplies it from the
+ * tree's own state: whether this node's branch is the selected one,
+ * the 24.04 spec §5), mirroring SideNavigation's own rail-collapse state
+ * rather than Accordion.Item's externally-controlled circuit; each
+ * instance is independent. That seed is live: when it turns `true` on
+ * navigation the disclosure re-opens — one-way, so a manual collapse stays
+ * collapsed and the disclosure never auto-closes.
```
advl, line 22: reduce verbosity, remove references to spec

packages/react/ds-app/src/lib/SideNavigation/common/ItemExpandable/ItemExpandable.tsx
```diff
@@ -0,0 +1,136 @@
...
+    [collapseOnChildClick],
+  );
+
+  return (
+    <li
+      className={[componentCssClassName, className].filter(Boolean).join(" ")}
+      data-disabled={disabled || undefined}
+      {...props}
+    >
+      <details className="details" open={expanded} onToggle={handleToggle}>
```
advl, line 100: Check if there is an actual need to control the state, esp if we're using native elements

packages/react/ds-app/src/lib/SideNavigation/common/ItemExpandable/types.ts
```diff
@@ -0,0 +1,33 @@
...
+   */
+  heading?: ReactNode;
+  /** Leading icon (start slot), by ds-assets icon name. */
+  icon?: IconName;
+  /** Whether the item (and its disclosure) is interactive. */
+  disabled?: boolean;
+  /**
+   * Initial open state when uncontrolled. Defaults to `false`. Seed from
+   * whether the active item is among this item's children (the 24.04
+   * spec §5) so the current page's ancestor chain starts expanded.
```
advl, line 17: remove references to specs

packages/react/ds-app/src/lib/SideNavigation/common/hooks/useCollapseShortcut/useCollapseShortcut.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export const COLLAPSE_SHORTCUT = { key: "b", ctrlKey: true } as const;
+
+/**
+ * Reserved: binds `COLLAPSE_SHORTCUT` to `onTrigger` while `enabled`.
+ * `SideNavigation` calls this with `enabled: keyboardShortcut`, which
+ * defaults to `false` — the shortcut is **pending approval**, so no
+ * listener is ever attached and it stays inert until a consumer opts in.
+ */
+export const useCollapseShortcut = ({
```
advl, line 20: default export per the code standards

packages/react/ds-app/src/lib/SideNavigation/common/hooks/useCollapseShortcut/useCollapseShortcut.ts
advl, line 1: Needs to check what's in focus

packages/react/ds-app/src/lib/SideNavigation/common/hooks/useCollapseShortcut/types.ts
```diff
@@ -0,0 +1,14 @@
+export interface UseCollapseShortcutProps {
+  /**
+   * Whether the shortcut listener is attached at all. Defaults to `false` —
+   * the spec's own prose is self-contradictory on the key (states "Ctrl + E",
+   * then argues for a single letter instead, in the same paragraph), so this
+   * ships disabled pending a design ruling (SPEC.md §5, §10.1).
+   */
+  enabled?: boolean;
```
advl, line 8: condition is more semantic than enabled

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
```diff
@@ -0,0 +1,153 @@
...
+    if (context) onContextChange?.(context);
+  };
+
+  return (
+    <>
+      {title != null && <GroupHeader>{title}</GroupHeader>}
+      <ContextualMenu
+        className={[componentCssClassName, className].filter(Boolean).join(" ")}
+        surfaceClassName={surfaceCssClassName}
+        trigger={
```
advl, line 139: Doesn't seem like the best pattern to pass the parent to its own child in this way. This might need to be fixed on the contextual menu. 

Potential follow-up

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
```diff
@@ -0,0 +1,153 @@
...
+    if (item.key === "create-context") {
+      onCreateContext?.();
+      return;
+    }
+    const context = contexts.find((candidate) => candidate.key === item.key);
+    if (context) onContextChange?.(context);
+  };
+
+  return (
+    <>
```
advl, line 134: We want to avoid Fragments as the wrapper of components since they make component-based customization impossible, for instance passing props to the element - or an id, or custom styles

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
```diff
@@ -0,0 +1,153 @@
...
+  item: MenuItem;
+}): React.ReactElement => {
+  const { renderLabel } = item as ContextMenuItem;
+  return (
+    <>
+      <Icon icon="plus" />
+      <span className="label">{renderLabel}</span>
+    </>
+  );
+};
```
advl, line 65: Nit: really needed as a subcomponent for just one usage ? Left to your choice

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
advl, line 9: Surfaces should come from the modifier families compiled from the design tokens, not hard-coded and made specific to a component

advl (changes requested):
Minor comments 

It would be great to double check the accessibility contract for all subcomponents

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
potentialwook: agreed will remove

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
potentialwook: trigger to become child instead of prop

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
potentialwook: Resolved up-stack in #1285: the custom surface is gone entirely — no `surfaceClassName` hook, no surface-specific class. The dropdown is sized through ContextualMenu's public `maxWidth` prop (rail width minus row insets) and its items are styled structurally via their own `.ds.contextual-menu-item` class, which the apps cascade layer already outranks at equal specificity.

packages/react/ds-app/vitest.setup.ts
potentialwook: Resolved up-stack in #1285: trimmed to the one line it earns — why a class instead of a `vitest.fn()` mock.

packages/react/ds-app/src/lib/SideNavigation/common/hooks/useCollapseShortcut/index.ts
potentialwook: Resolved up-stack in #1285: moved to `SideNavigation/hooks/useCollapseShortcut/` — the hooks folder of the domain the shortcut serves, per the custom-hooks placement standard.

packages/react/ds-app/src/lib/SideNavigation/common/ItemExpandable/ItemExpandable.tsx
potentialwook: Resolved up-stack in #1285: the doc comment is trimmed to the contract (native disclosure, the one-way re-open seed, the footer opt-in) — no spec references, no citation style.

packages/react/ds-app/src/lib/SideNavigation/common/ItemExpandable/ItemExpandable.tsx
potentialwook: Resolved up-stack in #1285: the `<details>` now owns its open state. The `open` attribute is seeded from a mount-only snapshot of `defaultExpanded` (SSR renders it; a seed flipping back to false never force-closes an open branch), and the two behaviours that genuinely need JS — the one-way re-open on navigation and the footer's collapse-on-child-activation — reach the DOM through a ref. No state mirror, no toggle handler; every behaviour test passes unchanged.

packages/react/ds-app/src/lib/SideNavigation/common/ItemExpandable/types.ts
potentialwook: Resolved up-stack: the spec citations here were already gone in #1285's base with the spec-free-comments pass (`docs(ds-app): stop quoting the 24.04 spec in comments and test titles`); nothing further needed.

packages/react/ds-app/src/lib/SideNavigation/common/hooks/useCollapseShortcut/useCollapseShortcut.ts
potentialwook: Resolved up-stack in #1285: the hook is now the module's single default export; `COLLAPSE_SHORTCUT` is single-use and stays as an unexported module constant, so the barrel no longer re-exports it.

packages/react/ds-app/src/lib/SideNavigation/common/hooks/useCollapseShortcut/useCollapseShortcut.ts
potentialwook: Resolved up-stack in #1285: the match only fires when the event's target is not an editable element (input, textarea, select, or a contenteditable host) — Ctrl+B is bold in editors and the shortcut must never steal it. Selector-based so jsdom resolves the same answer; covered by the editable-target tests in `useCollapseShortcut.tests.tsx`.

packages/react/ds-app/src/lib/SideNavigation/common/hooks/useCollapseShortcut/types.ts
potentialwook: Resolved up-stack in #1285: renamed to `condition` — the prop now names the state it represents rather than a bare on/off switch; `SideNavigation` feeds it its `keyboardShortcut` flag.

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
potentialwook: Resolved up-stack in #1285: the trigger content is now ContextualMenu's children — `<ContextualMenu items={…}>…</ContextualMenu>` — rendered inside the same real `<button>`. The `trigger` prop is gone (marked breaking for the released surface).

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
potentialwook: Resolved up-stack in #1285: the root is now the component's own wrapper `<div>` — it carries the DS class and the spread rest props, so a consumer can address it (className, id, data attributes); the caption and the dropdown field are its children.

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
potentialwook: Resolved up-stack in #1285: the two single-usage content renderers merged into one that branches on `renderLabel` — the create-context action no longer needs its own component.

packages/react/ds-app/src/lib/SideNavigation/common/ContextSwitcher/ContextSwitcher.tsx
potentialwook: Agreed — resolved up-stack in #1285: the switcher no longer defines or passes any surface of its own. Sizing goes through the public `maxWidth` prop and item styling through `.ds.contextual-menu-item` (see the reply on the sibling thread).

advl (approved):

PR #1242, opened 2026-09-11
fix(react-ds-global-form): refine choice options, toggle fields and label markers

potentialwook:
## Done

- `Choices Option` now composes the `RadioInput`/`CheckboxInput` and `Field.Label` subcomponents instead of raw `<input>`/`<label>` markup, dropping the `biome-ignore` a11y suppression; its stylesheet gains a `margin-inline` reset and centres the option row on the control.
- Toggle fields (`CheckboxField`, `SwitchField`) recentre the control row (spacing belongs to the field, not the label/control) and, when the control is disabled, the label recolours to `--color-text-disabled` — state-driven via `:has(input:disabled)`, replacing the old whole-row opacity dim.
- Choice options recolour their label the same way via the existing `.disabled` class.
- The disabled required marker recolours with the label through the `--form-required-marker-color-disabled` / `--form-label-color-disabled` variable chain (falls back to `--color-text-disabled`).
- The field Label required marker moves from `::before` to `::after` — it now trails the label text ("Email*") — with stories/tests/comments updated.
- `CheckboxInput` checkmark mask scales to 100% (was 75%, leaving a gap in the glyph).
- Density body font tier extended to toggle rows and choice options, which lost `.p`.

Fixes the label staying full-colour while its checkbox/radio/switch is disabled.

## QA

- [x] `bun run check` and `bun run test` green from the repo root (66 / 55 projects).
- Storybook: verify a disabled `CheckboxField`/`SwitchField` and a disabled `Choices` group — label + required marker take the disabled text colour; required marker renders after the label.
- Chromatic: visual diff expected (disabled states, marker position, checkbox mask, option composition) — do not skip.

### PR readiness check

- [x] PR title follows Conventional Commits (`fix` type).
- [x] Code follows the repo code standards; root `check`/`test` gates green.
- [x] No new package; all required scripts already defined in `package.json`.
- [ ] `Chromatic: skip` intentionally NOT set — this PR has visual diffs.


advl:
Merged with thanks

PR #1260, opened 2026-09-13
fix(Accordion): seat the h6 heading on the density cell, and give the content its 16px bottom gap

stavropouloskonstantinoss:
## Done

Two accordion fixes from the design [review](https://warthogs.atlassian.net/browse/WD-38577)

1. **h6 heading** — an `<h6>` heading made its row taller than the others.
   Now it follows the shared row height like plain-text headings (32/24px
   apps, 36/28px sites) and keeps its heading look.

2. **Content bottom gap** — the space above the closing divider was too
   tight (8px). Now 16px, and 8px in apps-dense, following the density
   spacing.

3. **Cap text length** — cap the content text measure at 40em

## QA

Please first read the [review fixes](https://warthogs.atlassian.net/browse/WD-38577)

bun install
cd packages/react/ds-global
bun run storybook

 Storybook → **components/Accordion/Heading**: in all four Context/Density
   combos both rows are the same height.
2. **components/Accordion/Default**: gap above the divider is 16px (8px dense).

### PR readiness check

## Screenshots


packages/react/ds-global/src/lib/component/Accordion/common/Item/styles.css
```diff
@@ -328,5 +367,16 @@
...
+    /* The h6 keeps its heading-6 font, but its own line box is FIXED (the
+       heading-6 tier's line-height plus the typography engine's padding
+       nudges), so it can't follow the seat — declare the seat line on it and
+       zero the nudges. Without this the h6's 24px box overflows the dense
+       cells (seat line 16px apps / 20px sites); the comfy cells only fit it
+       by coincidence, which this makes true by construction. */
+    & h6 {
+      padding-block: 0;
+      line-height: inherit;
+    }
```
advl, line 380: We should ideally not hard-code any specific typography logic here, this is a "snowflake" and is considered to be an anti-pattern for maintenance. We need to find a better way to do that. 

Also we cannot zero the nudge since the text will not sit on the baseline anymore.

advl (changes requested):
See comment

PR #1285, opened 2026-09-15
refactor!: answer the SideNavigation stack's review threads

potentialwook:
## Done

Answers every open review thread across the SideNavigation stack — the two on #1225 (`@canonical/react-ds-global`) and the twelve on #1234 (`@canonical/react-ds-app`), on top of the three #1227 threads this PR already carried. Stacked on #1229; merge last, after the whole SideNavigation stack.

**react-ds-global (from #1225's review):**

- **TooltipEngine gets its own file.** The engine was a named export of `withTooltip.tsx` — the one module in the package exporting two differently-shaped things under one roof. It now lives in `TooltipEngine.tsx` with a single default export; `withTooltip.tsx` keeps only the HOC and consumes the engine like any other caller. The engine's tests moved with it, so each test file mirrors the module it covers.
- **ContextualMenu stops dressing its surface as a modal.** The surface wore the design tokens' `.modal` class — the class the Modal pattern owns — so it read the modal surface channel instead of its own mapped token and silently tracked Modal's background. The class is dropped from the root surface and the submenus; the surface paints `--color-background` (its own mapped value) directly. `surfaceClassName` — the hook this stack added so a consumer could restyle the portaled surface — goes with it: consumers size the menu through the public `maxWidth` prop and style the items structurally via their own `.ds.contextual-menu-item` class (the apps cascade layer outranks the item's own global layer at equal specificity, so the override holds without a per-consumer class).
- **BREAKING — the trigger content is ContextualMenu's children.** `<ContextualMenu items={…}>Actions</ContextualMenu>` instead of the `trigger` prop: the menu's caller no longer hands the parent its own child as configuration. The same real `<button>` renders underneath; every test, story, and consumer is updated.

**ds-app (from #1234's review):**

- **ContextSwitcher restructured around a real wrapper element.** The Fragment root is gone — the component's own `<div>` carries the DS class and the spread rest props, with the caption and the dropdown field as its children. The two single-usage content renderers merged into one that branches on `renderLabel`; the create-context row's layout rides a class only this component's renderer emits. The ContextualMenu passthroughs (`open`, `onOpenChange`, positioning) are forwarded explicitly now that the rest spread lands on the wrapper, and the dropdown defaults to the rail width minus its row insets via `maxWidth`, overridable per instance.
- **ItemExpandable's `<details>` owns its open state.** The controlled mirror is gone: the `open` attribute is seeded from a mount-only snapshot of `defaultExpanded` (so SSR renders it, and a seed flipping back to `false` never force-closes an open branch), and the two behaviours that genuinely need JS — the one-way re-open when navigation makes this branch the selected one, and the footer's collapse-on-child-activation — reach the DOM through a ref. Every existing behaviour test passes unchanged.
- **useCollapseShortcut moved into its domain** — `SideNavigation/hooks/useCollapseShortcut`, the hooks folder of the component it serves — and became the module's single default export; `COLLAPSE_SHORTCUT` was its only other consumer-facing name and is single-use, so it stays as an unexported module constant. The boolean gate renamed from `enabled` to `condition`: the prop now names the state it represents (the `keyboardShortcut` flag `SideNavigation` feeds it), not a bare on/off switch. The chord is also ignored when the event's target is an editable element — Ctrl+B is bold in editors, and the shortcut must never steal it.
- **The vitest setup's ResizeObserver comment** is trimmed to the one line it earns.

**Carried from the original description (#1227's threads):**

- **Direct element selection instead of utility classes.** The collapse toggle showed and hid its icon and its "Menu"/"Close menu" label through `.desktop-only`/`.mobile-only` utility classes. The button's only two children are the icon's `svg` and the label's `span`, so the stylesheet selects them directly — no utility classes, no extra class surface on the DOM. Rendered output is unchanged on either side of the small breakpoint.
- **Spec-quoting comments rewritten.** Comments across SideNavigation cited the 24.04 design spec two ways — quoted sentences and section numbers — which reads as spec transcription rather than reasoning and goes stale the moment the spec is revised. They now carry the decision and its rationale in the codebase's own words; deliberate deviations still record what they deviate from (the upstream 24.04 design's stated value) but nothing quotes spec text or cites a section number. Six test titles drop their "(the 24.04 spec §N)" parentheticals. Two navigation pointers to `SPEC.md` stay (the stories props-table pointer and the ContextSwitcher known-issues pointer) — they reference the document without quoting it. `SPEC.md`'s own citation-convention note goes with the convention it described.
- **Cascade layers.** The review's "needs the layer" thread is already resolved by #1228 and #1229: every SideNavigation stylesheet at the tip of the stack is wrapped in `@layer ds.components.apps` with its layer header. This PR keeps that intact; no change needed.

## QA

- `bun run check` passes from the repo root (66 projects).
- The two touched packages' suites pass standalone: react-ds-global 422 tests, ds-app 162 tests (behaviour tests unchanged through the ItemExpandable uncontrolled refactor and the ContextSwitcher restructure).
- The root-wide serial test run was cut short locally at the maintainer's call; CI runs the full matrix.
- Grep-verified: no `§` citations, quoted spec text, or ADR-Txx references remain in SideNavigation source outside `SPEC.md` itself.

### PR readiness check

- [x] PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format, using one of the allowed types: `feat`, `fix`, `docs`, `refactor`, `chore`, `test`, `ci`, `revert`. Marked `!` — the ContextualMenu trigger-prop removal is breaking for the released surface.
  - Breaking changes are marked with `!` in the title (e.g. `feat(router)!:
…`), which adds the `breaking` label.
- [ ] If this PR does not require visual testing, add the `Chromatic: skip` label to skip Chromatic. — intentionally NOT skipped: ContextualMenu now renders at its max-width instead of content width (a dropdown reads as one stable-width field), so visual diffs in the menu stories are expected and need a look in Chromatic review.

## Screenshots

Not relevant beyond the Chromatic note above — the ContextSwitcher trigger and ItemExpandable rows render the same DOM as before; the intended visual change is the menu surface's width.


packages/react/ds-app/src/lib/SideNavigation/styles.css
stavropouloskonstantinoss, line 101: try making the transition very slow 

transition:
      width 2s ease-in-out,
      inline-size 2s ease-in-out;

, you will see that when expanding the sidenav the header is immediately visible. 

possibly a fix is adding

    &:not([data-expanded="false"]) {
      overflow: hidden;
    }

advl (approved):

PR #1311, opened 2026-09-16
feat(InputPrimitive): migrate to design tokens

TudorGiuroiu:
## Done

- Migrated `InputPrimitive` styles from `--lp-*` Launchpad tokens to design tokens (`@canonical/styles`):

Fixes LP-4491

## QA

- Run `bun run check` and `bun run test` at the repository root.
- Run Storybook for the consuming components and visually check them. Some aspects might remain unchanged due to some DS tokens haveing the same values as the LP tokens. At least, check the differences noted in LP-4491.
- Confirm that no `--lp-*` Launchpad tokens are still used by `InputPrimitive` and that they have been replaced with the corresponding DS tokens.

### PR readiness check

## Screenshots

Differences are highlighted through comparing the components consuming `InputPrimitive`, because `InputPrimitive` alone does not have a Storybook entry.
Invalid Before: 
image
Invalid After:
image

Invalid Focus Before: 
image
Invalid Focus After:
image

Invalid Active Before:
image

Invalid Active After:
image

Disabled Before:
image

Disabled After:
image


packages/svelte/ds-app-launchpad/src/lib/components/common/InputPrimitive/styles.css
```diff
@@ -50,7 +49,9 @@
 
   &:disabled {
     cursor: not-allowed;
-    opacity: var(--opacity-input-disabled);
+    background-color: var(--color-foreground-input-disabled);
+    color: var(--disabled--color-text);
```
goulinkh: can you please map this to a component token similar to `--color-foreground-input-disabled`. I am not sure whether pragma has a standard for this anymore (it used to be), for now let's keep the tokens consistent.

same for the token bellow: `--disabled--color-border`

goulinkh (changes requested):
Changes look good, left a nit comment.

packages/svelte/ds-app-launchpad/src/lib/components/common/InputPrimitive/styles.css
TudorGiuroiu: Didn't you mean to give as example another variable? This one,  `--color-foreground-input-disabled`, doesn't seem to be mapped to any component token in this file.

goulinkh: good point! it seems that `--color-foreground-input-disabled` is not defined which is another issue :)

an existing "component token" example is `--color-foreground-input-disabled`

goulinkh (approved):

PR #1312, opened 2026-09-16
feat(ds-app): add the SidePanel component

stavropouloskonstantinoss:
## Done

- Adds the app-tier `SidePanel` pattern to `@canonical/react-ds-app`: a
  non-modal `<dialog>` docked to the inline-end edge of the viewport, opened
  with `show()` so the application behind stays interactive.
- Controlled imperatively through a required `ref` handle (`open()`,
  `close()`, `element`). The dialog's native open state is the single source
  of truth — there is no `open` prop — and every change is reported through
  `onOpenChange`.
- Focus moves into the panel on open and is handed back on close, dismissal,
  or unmount-while-open. Escape closes the panel unless `disableEscapeClose`.
- Composes from `SidePanel.Header` / `SidePanel.Content` / `SidePanel.Footer`;
  header and footer stay fixed while only the content scrolls. A panel without
  a header takes `aria-label` instead.
- Exported from `@canonical/react-ds-app`.

Stacked PR — 1 of 4. Its tests and stories land in the PR stacked on this one.

## QA

bun install
cd packages/react/ds-app

- The component is exercised by the tests and stories PR stacked on top of
  this one.
### PR readiness check

## Screenshots


packages/react/ds-app/src/lib/SidePanel/common/Header/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * Title content. Names the panel for assistive technology.
+   *
+   * The title is deliberately not a heading element: the panel is a layer on
+   * top of the page, not part of its document outline, so it names the dialog
+   * through `aria-labelledby` instead.
+   */
+  children: ReactNode;
+  /** Accessible name for the close button. Defaults to "Close panel". */
+  dismissLabel?: string;
```
advl, line 13: not optional then if a default is provided

packages/react/ds-app/src/lib/SidePanel/hooks/useSidePanelState.ts
```diff
@@ -0,0 +1,20 @@
+import type { RefObject } from "react";
+import { useCallback, useId } from "react";
+import type { SidePanelContextValue } from "../types.js";
+
+/**
+ * The provider state threaded to the composed subcomponents: the id that names
+ * the panel, and the dismissal that closes it. Centralising it here keeps the
+ * provider free of state logic.
+ */
+const useSidePanelState = (
```
advl, line 10: The title of this hook is misleading. One might think we are being returned a state object, but we are only being returned metadata or state helpers. This hooks needs to be retaken or renamed following a clearer identification of the concerns it serves.

packages/react/ds-app/src/lib/SidePanel/styles.css
```diff
@@ -0,0 +1,178 @@
...
+    background-color: var(--surface-color-background, var(--color-background));
+    color: var(--surface-color-text, var(--color-text));
+
+    /* Closed state. `translate` is physical, so flip it when the panel docks left. */
+    --side-panel-translate-closed: 100%;
+    transform: translateX(var(--side-panel-translate-closed));
+
+    &:dir(rtl) {
+      --side-panel-translate-closed: -100%;
+    }
```
advl, line 138: probably worth creating a story

packages/react/ds-app/src/lib/SidePanel/styles.css
```diff
@@ -0,0 +1,178 @@
...
+
+    /*
+      The panel is focused when it opens so that reading starts inside it, but it
+      is a container rather than a control: a focus ring drawn around the whole
+      panel reads as a border. Suppressed for pointer focus only — a keyboard
+      user must still be able to see where focus landed, so `:focus-visible`
+      keeps its ring. Its focusable children keep their own either way.
+    */
+    &:focus {
+      outline: none;
```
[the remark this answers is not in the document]
advl: Might be worth a code standard if relevant

packages/react/ds-app/src/lib/SidePanel/types.ts
```diff
@@ -0,0 +1,81 @@
...
+   * from any cause — the handle, the header's close button, Escape, or the
+   * platform itself. Informational: the panel reports what happened, it does
+   * not ask permission. Consumers who mirror the state (a toggle button's
+   * pressed state, say) follow this; the panel never reads it back.
+   */
+  onOpenChange?: (open: boolean) => void;
+  /**
+   * Stop Escape from closing the panel. Defaults to false, so Escape closes
+   * the panel while focus is inside it. Focus outside the panel belongs to the
+   * application, so Escape there is deliberately not intercepted.
```
advl, line 42: The documentation of properties should take into account the user/consumer's perspective. Here nothing indicates in which case should this prop be used (likely exceptionally) - which makes it impossible to review it from a relevance perspective

packages/react/ds-app/src/lib/SidePanel/types.ts
```diff
@@ -0,0 +1,81 @@
...
+
+type OwnProps = {
+  /**
+   * Called whenever the panel's open state changes, in either direction and
+   * from any cause — the handle, the header's close button, Escape, or the
+   * platform itself. Informational: the panel reports what happened, it does
+   * not ask permission. Consumers who mirror the state (a toggle button's
+   * pressed state, say) follow this; the panel never reads it back.
+   */
+  onOpenChange?: (open: boolean) => void;
```
advl, line 38: How certain are you that this will be used ?

packages/react/ds-app/src/lib/SidePanel/common/Header/styles.css
```diff
@@ -0,0 +1,48 @@
+/* Layer: ds.components.apps (@canonical/styles README, "Cascade layers"). */
+
+@layer ds.components.apps {
+  .ds.side-panel-header {
+    /* Never shrinks, and stays visible: the panel scrolls beneath it. */
+    flex: none;
+    position: sticky;
+    inset-block-start: 0;
+    /* Above the in-flow content, whatever it contains. */
+    z-index: 1;
```
advl, line 10: I think we should avoid sticky/z-index and instead use a flexbox for the sidepanel column. Iti s a neater and more established pattern. Sticky can be used for elements that are not naturally "at the top" on scrolling - which is not the case for flexbox layouts

packages/react/ds-app/src/lib/SidePanel/common/Header/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const context = useSidePanelContext();
+
+  return (
+    <div
+      className={[componentCssClassName, className].filter(Boolean).join(" ")}
+      {...props}
+    >
+      <span className="title" id={context?.titleId}>
+        {children}
+      </span>
```
advl, line 40: This makes the assumption the consumer should have no say in the display of the sidepanel title. While it might be true, it seems to be undocumented. 

Also the particular heuristic to map to h2 font sizes, while possibly accurate, seems like a snowflake - we need proper semantic rules to determine typography sizes that don't make it seem mapping are arbitrary

packages/react/ds-app/src/lib/SidePanel/common/Content/styles.css
```diff
@@ -0,0 +1,17 @@
+/* Layer: ds.components.apps (@canonical/styles README, "Cascade layers"). */
+
+@layer ds.components.apps {
+  .ds.side-panel-content {
+    /* Takes the space the header and footer leave, so a short panel still fills. */
+    flex: 1 1 auto;
```
advl, line 6: This seems it hasn't taken into account the comment by @danielmutis 's - it is preferred that the content doesn't flex grow by default (we can perhaps add a toggle for that) so that the buttons are not kept at a distance from the text they refer to. 

I would recommend keeping a switch/override prop to enable the flex behaviour.

packages/react/ds-app/src/lib/SidePanel/common/Content/styles.css
```diff
@@ -0,0 +1,17 @@
...
+
+    /*
+      No scroller here: overflow spills to the panel, which scrolls. That is what
+      keeps the scrollbar at the panel's edge rather than inside the text column.
+    */
+    padding-block: var(
+      --side-panel-content-padding-block,
+      var(--dimension-200)
```
advl, line 14: Are you certain that this default is necessary ?

packages/react/ds-app/src/lib/SidePanel/common/Footer/styles.css
```diff
@@ -0,0 +1,39 @@
+/* Layer: ds.components.apps (@canonical/styles README, "Cascade layers"). */
+
+@layer ds.components.apps {
+  .ds.side-panel-footer {
+    /* Never shrinks, and stays visible: the panel scrolls beneath it. */
+    flex: none;
+    position: sticky;
```
advl, line 7: same comment as the header, wouldn't we be better off having those elements on a flex layout ?

packages/react/ds-app/src/lib/SidePanel/common/Footer/styles.css
```diff
@@ -0,0 +1,39 @@
...
+      the two cannot state different defaults: `--side-panel-color-border` is
+      declared in the panel's `:root` block.
+    */
+    border-block-start: var(--dimension-stroke-thickness-medium) solid
+      var(--side-panel-color-border);
+
+    display: flex;
+    justify-content: flex-end;
+    align-items: center;
+    gap: var(--dimension-200);
```
advl, line 34: use variable

packages/react/ds-app/src/lib/SidePanel/common/Footer/styles.css
```diff
@@ -0,0 +1,39 @@
...
+    flex-wrap: wrap;
+
+    padding-block: var(--side-panel-footer-padding-block, var(--dimension-200));
```
advl, line 37: please no defaults if variable always exists

packages/react/ds-app/src/lib/SidePanel/Provider.tsx
advl, line 1: This needs a bit more conceptual clarity. Your file conflates both the `Provider` logic with the rendering logic. While there is not particular harm in doing that- it feels that the standards apply in an unidiomatic way  here. For instance 

```
const Provider = ({
}: SidePanelProps)
```

makes me wonder whether we are looking at a markup block or at a context provider. It seems that you haven't fully decided whether your consumption api is markup that ships with a context provider or a context provider that contains markup

packages/react/ds-app/src/lib/SidePanel/Provider.tsx
```diff
@@ -0,0 +1,203 @@
...
+  return (
+    <Context.Provider value={{ close, titleId }}>
+      <dialog
+        ref={dialogRef}
+        className={[componentCssClassName, className].filter(Boolean).join(" ")}
+        // The header's title names the panel. Without a header the consumer
+        // supplies `aria-label`, and pointing at an absent element is worse than
+        // not pointing at all — so the two are mutually exclusive.
+        aria-label={ariaLabel}
+        aria-labelledby={ariaLabel === undefined ? titleId : undefined}
```
advl, line 158: shouldn't there be always a title ?

packages/react/ds-app/src/lib/SidePanel/Provider.tsx
```diff
@@ -0,0 +1,203 @@
...
+        onKeyDown={(event) => {
+          onKeyDown?.(event);
+          if (
+            !disableEscapeClose &&
+            event.key === "Escape" &&
+            !event.defaultPrevented
+          ) {
+            close();
+          }
+        }}
```
advl, line 173: these should live in the body of the function inside `useCallback` memoization hooks

advl (changes requested):
Overall looks clean with minor idiom fixes

packages/react/ds-app/src/lib/SidePanel/common/Header/types.ts
stavropouloskonstantinoss: Found out that in ts the ? means that the caller doesnt have to provide it, not that the argument can be missing. If we remove the ?, then the caller must always provide it (even if it has a default).

[ts doc](https://www.typescriptlang.org/docs/handbook/2/objects.html#:~:text=Here%20we%20used%20a%20destructuring%20pattern,optional%20for%20any%20callers%20to%20paintShape)

PR #1313, opened 2026-09-16
test(ds-app): add SidePanel tests and stories

stavropouloskonstantinoss:
## Done

- Unit and SSR tests for `SidePanel`, `SidePanel.Header`, `SidePanel.Content`
  and `SidePanel.Footer`.
- Storybook stories for the same four.

Stacked PR — 2 of 4. Stacked on #1312; review after it. This PR's diff
contains only the test and story files for sidepanel.

## QA

bun install
cd packages/react/ds-app

- Stories visible in the Storybook build: Components/SidePanel.

### PR readiness check

## Screenshots

image


packages/react/ds-app/src/lib/SidePanel/SidePanel.stories.tsx
```diff
@@ -0,0 +1,290 @@
+import { Button, withTooltip } from "@canonical/react-ds-global";
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import Component from "./Provider.js";
+import type { SidePanelHandle } from "./types.js";
+
+/*
+  Real copy from ubuntu.com/about, used as filler wherever a story needs the
+  content pane to overflow.
+*/
+const ubuntuStory = [
```
advl, line 10: put this in storybook/fixtures instead to enhance reusability

packages/react/ds-app/src/lib/SidePanel/SidePanel.stories.tsx
```diff
@@ -0,0 +1,290 @@
...
+  the open comes from an event instead —
+  `onClick={() => panelRef.current?.open()}` on the trigger that owns the
+  panel.
+*/
+
+/**
+ * The open panel — the visual baseline, showing its primary use case: static
+ * details of a selected entity, with the application behind it still usable.
+ */
+export const Open: Story = {
```
advl, line 53: It is unclear what how you intend this story to be consumed. As a visual test, it provides a plausible story, but as a functional test, the buttons are not wired to the state of the dialog, which makes a potential consumer fail to achieve the desired behavior.

packages/react/ds-app/src/lib/SidePanel/SidePanel.stories.tsx
```diff
@@ -0,0 +1,290 @@
...
+ * In a real application the tooltip is hover- and focus-driven: it opens when
+ * the user hovers (or focuses) the trigger and closes the moment they move or
+ * scroll away. This story pins it open with `open: true` for one reason only:
+ * it exists for a static Chromatic snapshot, and a hover-driven tooltip cannot
+ * be hovered in a screenshot — the overflow/stacking behaviour it demonstrates
+ * would never be captured. The pinned-open state is a snapshoting device, not
+ * component behaviour; in particular, a real tooltip would close as the panel
+ * scrolls, while this one stays put.
+ */
+export const OverflowingTooltip: Story = {
```
advl, line 148: Definitely an interesting story - to keep - but makes me wonder if there are other cases we'd like to test before getting there ? It seems like a "precise" zoomed-in case - are there things we'd like to look at before ?

advl (changes requested):
The stories open the panel as the story is rendered which results in a short lag until the component is fully rendered (the transition for it to open). This should be reviewed with more attention as it has the potential to break out visual testing suite chromatic. It would be preferable is you could find a way to render the stories with the preferred state on first load but perhaps not for all stories ?

Furthermore, the set of stories in the PR do not seem to explain well the consumption patterns. See the related comment on the main stories file.

PR #1314, opened 2026-09-17
feat(ds-app): add withSidePanel consumption pattern

stavropouloskonstantinoss:
## Done

- Adds `withSidePanel` to `@canonical/react-ds-app`: wraps a trigger (anything
  that accepts `onClick` — `Button`, a link) with a panel it toggles. Click
  opens, click again closes; the header's close button and Escape dismiss too.
- The second argument is a render factory: the HOC calls it with
  `{ close, ref }` and it returns the complete `<SidePanel>` element. The
  factory must attach the `ref` — `SidePanel` requires it, so forgetting it is
  a compile error.
- Pure composition wrapper: renders trigger and panel as siblings; the panel is
  `position: fixed`, so its place in the tree does not move it.
- Extends `SidePanel`'s public types with the `WithSidePanel*` contracts and
  exports the HOC.

Stacked PR — 3 of 4. Stacked on #1313 ; review after it. This PR's diff
contains only the HOC and its types/export.

## QA

bun install
cd packages/react/ds-app

- The HOC is exercised by the tests and stories PR stacked on top of this one

### PR readiness check

## Screenshots


packages/react/ds-app/src/lib/SidePanel/withSidePanel.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const panelRef = useRef<SidePanelHandle>(null);
+
+    // The dialog's native open state is the only one: read it, then flip it.
+    const toggle = (): void => {
+      const handle = panelRef.current;
+      if (!handle) return;
+      if (handle.element?.open) handle.close();
+      else handle.open();
+    };
+
```
advl, line 93: this seems it should either live on the main state hook or the main state hook logic should be changed - otherwise both this file and the hook would handle the computation of derived state helpers.

packages/react/ds-app/src/lib/SidePanel/types.ts
```diff
@@ -79,3 +86,61 @@ type OwnProps = {
...
+export type WithSidePanelTriggerProps = {
+  onClick?(event: MouseEvent): void;
+};
+
+/**
+ * The props the panel element returned by a {@link WithSidePanelRender}
+ * carries: everything `SidePanel` accepts — including the required `ref`,
+ * which the factory sets on the `<SidePanel>` so the trigger can toggle it.
+ */
+export type WithSidePanelPanelProps = SidePanelProps;
```
advl, line 112: I am not sure I understand the use of this, could you explain what is the value you see it bringing ? 

If people are declaring their own `SidePanel` why should we on top offer an api to passover the props ?

packages/react/ds-app/src/lib/SidePanel/types.ts
advl, line 1: I would recommend reviewing this file - it seems very verbose for the scope of additions i provides.

advl (changes requested):
Overall clean, only minor comments

PR #1321, opened 2026-09-17
feat(pragma-cli): a pack built by an older CLI yields to the shipped one, and sources reset removes it

advl:
## Done

A pack built with `pragma sources update` is kept per working directory and wins over the snapshot shipped inside the CLI. After an upgrade that pack kept answering with old data, silently: a user on 0.38.0 was still reading a design system from a week earlier, with no token bindings, until they rebuilt by hand.

- A built pack now yields to the shipped snapshot when it was built by a strictly older CLI and the project uses the default sources. The pack manifest already records the CLI version that built it, so nothing new is stored. A project that declares its own packs is never overridden.
- Reads say so: `meta.ignoredPack` carries the ignored pack's hash, the version that built it and when. `sources status` prints one line naming the two ways out (`sources update` to rebuild, `sources reset` to remove).
- Reusing a cached pack refreshes the recorded builder version, so a rebuild that finds nothing to fetch still clears the notice.
- New verb `sources reset` (tool `sources_reset`): removes the pointer to the built pack for this directory, so the shipped snapshot answers. It leaves the cache itself alone. The tool count goes from 49 to 50 and the surface file records the addition.
- Docs: setup, architecture, design-system and the generated reference.

## QA

In an empty directory with an isolated cache (`export XDG_CACHE_HOME=$(mktemp -d)`):

1. `pragma sources update --yes`, then `pragma sources status` → `Store: built`.
2. Edit `version` in `$XDG_CACHE_HOME/pragma/packs/<hash>/manifest.json` to `0.37.0`.
3. `pragma sources status` → `Store: embedded` and an `Ignored:` line naming 0.37.0 and both commands.
4. `pragma token consumers --symbol color.text --format json` → rows, and `meta.ignoredPack`.
5. `pragma sources reset` → `Removed: <hash>`; again → `Removed: nothing was built`; status shows no `Ignored:` line.

Known and unrelated: `src/kernel/runtime/probe.test.ts` times out at 5 s under a loaded full run on `main` as well.

### PR readiness check

- [x] No new package
- [x] No visual change — `Chromatic: skip`

https://claude.ai/code/session_01M2RhPptdCDv62gGVgo7n9A


advl:
Two review lanes ran (correctness; tests and surface). Fixed in the two follow-up commits: the recorded builder version now only moves up (the pack cache is shared between projects); `sources reset` no longer claims an `--undo` that did nothing; the MCP path and the llm renderers are now tested; the reset summary is correct for a project that declares its own packs; one duplicated type removed. Left as is: three test fixtures use `0.1.0` as the older CLI, which only fails if the suite is ever run with a `0.0.0` development version.

PR #1334, opened 2026-09-18
feat(icons): support subpath-hosted assets

### LINKED ISSUE #1304: Global icon root: configure the icon base path once for subpath-hosted apps
## Problem

Icons are fetched at runtime by URL, and the URL is hard-coded to `/icons` in two places:

1. **JS** — `Icon` (and `Spinner`) build `` `${rootPath}/${icon}.svg#${icon}` `` with `rootPath` defaulting to `"/icons"` (`packages/react/ds-global/src/lib/component/Icon/Icon.tsx`).
2. **CSS** — the `Accordion` caret uses `mask-image: url("/icons/chevron-down.svg#chevron-down")` (`Accordion/common/Item/styles.css`), and the criticality tokens in `@canonical/styles` set `--modifier-icon: url("/icons/<name>.svg#<name>")` (`modifiers.criticality.shim.css`).

An app hosted under a subpath (Landscape runs under `/new_dashboard/`) therefore gets empty icons everywhere unless it serves `@canonical/ds-assets` at the site root.

The per-instance `rootPath` prop is not a fix:

- `Button` (`icon=`) and `ContextualMenu` (submenu chevron) render `<Icon>` internally without forwarding `rootPath`, so a wrapper around `Icon` does not reach them.
- No prop can reach the CSS-referenced icons.

The README already says "a single global source of truth for the icon root is planned"; this issue tracks it.

## Proposal

One icon root, readable from both CSS and JS, set once by the app.

### CSS token as the source of truth

Pragma declares `--ds-icon-root: /icons` in its base stylesheet. An app overrides it once:

```css
:root { --ds-icon-root: /new_dashboard/icons; }
```

A CSS variable rather than a React context because CSS can read it, JS can read it, it works for Svelte and any future implementation with no extra API, and it needs no provider wrapping the tree.

### CSS side

CSS cannot concatenate inside `url()`, so `url(var(--ds-icon-root) "/x.svg")` is invalid. Instead, each CSS-referenced icon becomes its own full-URL token, the way `--modifier-icon` already works:

```css
:root {
  --ds-icon-chevron-down: url("/icons/chevron-down.svg#chevron-down");
  /* + information, success, warning, error */
}
```

and `Accordion` uses `mask-image: var(--ds-icon-chevron-down)`. Only ~5 icons are referenced from CSS today, so the override surface is small. Ideally these tokens are generated from `@canonical/ds-assets` so new CSS-referenced icons get a token automatically.

### JS side

`Icon` and `Spinner` keep the `rootPath` prop for per-instance overrides, but the default resolves from a new `IconRootContext` (exported from `@canonical/react-ds-global`), then falls back to `/icons`. `Button` and `ContextualMenu` need no change: their inner `<Icon>` picks up the context.

The context exists because of SSR: `getComputedStyle` is unavailable on the server, and reading it in an effect would produce a hydration mismatch. Apps that server-render set the provider alongside the CSS token:

```tsx
<IconRootProvider root="/new_dashboard/icons">…</IconRootProvider>
```

### Scope

- `Icon.tsx`, `Spinner.tsx`: default `rootPath` from context, then `/icons`.
- New `IconRootContext` + `IconRootProvider`, exported from `@canonical/react-ds-global`.
- `Accordion/common/Item/styles.css`: `var(--ds-icon-chevron-down)` instead of the literal `url()`.
- `@canonical/styles`: default `--ds-icon-root` and `--ds-icon-*` tokens next to the existing `--modifier-icon` ones.
- README: replace the "no global default" paragraph with the token + provider instructions.
- Tests: SSR test for the provider; `Button` with `icon=` honours the provider without a prop.

Small enough for one PR.

## Workaround until then

Serve the `@canonical/ds-assets` icons at the site root (`/icons/*.svg`) via a proxy or static-file rule, even when the app lives under a subpath. This covers the JS and CSS cases with no Pragma change.

## Context

Raised by the Landscape team: their app is hosted under `/new_dashboard/` and they are considering a shared wrapper component as a stopgap.

---

rubinaga:
## Done

- Resolve CSS-referenced icons in `@canonical/react-ds-global` and `@canonical/react-ds-global-form` through `@canonical/ds-assets` package URLs, allowing asset-aware bundlers to inline or rebase them for subpath-hosted applications.
- Add a framework-neutral `setIconRoot()` API to `@canonical/ds-assets` and re-export it from `@canonical/react-ds-global` so applications can configure runtime `Icon` and `Spinner` assets once.
- Preserve `/icons` as the default and `rootPath` as the per-instance override.
- Add an explicit `@canonical/ds-assets` package export map while preserving root, icon, font, and package metadata entry points.
- Document the runtime and CSS asset contracts and cover direct and internally rendered icons with tests.
- Leave the legacy criticality shim unchanged because `@canonical/ds-assets` remains optional for `@canonical/styles`.

Fixes https://github.com/canonical/pragma/issues/1304

## QA

- `bun run check` from the repository root: passed across all 68 projects.
- `bun run build` from the repository root: passed, 56/56 targets.
- `bun run test` in `packages/ds-assets`: passed, 21 tests.
- `bun run test` in `packages/react/ds-global`: passed, 427 tests with 5 skipped.
- `bun run test` in `packages/react/ds-global-form`: passed, 293 tests.
- `bun run test` in `packages/storybook/helpers`: passed, 71 tests.
- Built the Vite React boilerplate with `base=/new_dashboard/`; all changed CSS icons were emitted under `/new_dashboard/assets` with no unresolved package URLs.
- The full root test fan-out could not complete on this laptop because Playwright is missing required host libraries. The isolated failing Svelte target collected 21 passing tests before browser startup aborted for missing `libevent`/GStreamer dependencies.

### PR readiness check

  - The matching type label is applied automatically from the title; there is no type label to add by hand.
  - Breaking changes are marked with `!` in the title, which adds the `breaking` label.
- [x] The code follows the appropriate [code standards](https://github.com/canonical/web-code-standards).
- [x] All affected packages retain the required `check`, `check:fix`, `test`, `build`, and `build:all` scripts.
- [x] This PR does not introduce a new package.

Chromatic: skip

## Screenshots

Not applicable; this changes icon asset resolution without changing component appearance.


rubinaga:
After validating Landscape's server routing, /new_dashboard/* is an SPA fallback and cannot serve nested icon files; icons are correctly mounted at root /icons, which already matches Pragma's default, so the proposed API would not solve a Landscape problem. The CSS packaging change can be reconsidered separately if Pragma wants it independently.

